PR #3, opened 2016-02-24
Introduce a clock abstraction, so we don't need to rely on DateTime.U…

jskeet:
…tcNow directly

The GetCurrentDateTimeUtc name is unwieldy, but:
- It should be a method not a property, as Eric Lippert describes in http://ericlippert.com/2014/05/19/when-should-i-write-a-property/
- The name needs to indicate that it's UTC
- The noun part of the name might as well use DateTime as that's what it returns

We could consider using DateTimeOffset instead, although the offset would always be zero.

(In Noda Time 2.0 this is GetCurrentInstant, which is cleaner... but we really need to work with the BCL here...)

### DIFF
--- a/src/Google.Api.Gax/Api/Gax/SystemClock.cs
+++ b/src/Google.Api.Gax/Api/Gax/SystemClock.cs
@@ -1,0 +1,33 @@
+﻿/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+ 
+ using System;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// A singleton implementation of <see cref="IClock"/> which delegates to the BCL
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax/Api/Gax/SystemClock.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        private SystemClock()
+        {
+        }
+
+        /// <summary>
+        /// Returns the current date and time in UTC, using <see cref="DateTime.UtcNow"/>.
+        /// </summary>
+        /// <returns>The current date and time in UTC.</returns>
+        public DateTime GetCurrentDateTimeUtc()
```
chrisdunelm: C# 6!
public DateTime GetCurrentDateTimUtc() => DateTime.UtcNow;

chrisdunelm:
LGTM

src/Google.Api.Gax/Api/Gax/SystemClock.cs
jskeet: Doh, of course :) Done.

PR #6, opened 2016-02-24
Update project.json and AssemblyInfo files

jskeet:
These changes were made largely on the basis of what the corefx and aspnet teams are doing.

- The test project doesn't need a version at all, or AssemblyInfo
- The main project doesn't need most of the attributes in AssemblyInfo; they can be provided by project.json where appropriate
- Filled in a few more fields in project.json

We should now be okay to start building packages in CI on MyGet


chrisdunelm:
LGTM

PR #7, opened 2016-02-24
Initial docfx support

jskeet:
Enough to at least let us build and view documentation; whether we add articles here (vs in a language-neutral repo) is an open question.


chrisdunelm:
LGTM

PR #8, opened 2016-02-25
Sign both test and prod assemblies

jskeet:
The private key is checked in, as recommended in
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md

The public key can be derived from the private one, so there's no need to check it in here.
The test assembly needs to be signed as the prod assembly exposes its internals to the test assembly.


src/Google.Api.Gax/project.json
```diff
@@ -7,6 +7,9 @@
     "projectUrl": "htttps://github.com/googleapis/gax-dotnet",
     "licenseUrl": "https://developers.google.com/open-source/licenses/bsd",
     "tags": [ "Google" ],
+    "compilationOptions": {
+        "keyFile": "../../Gax.snk"
```
chrisdunelm: Unless I'm missing something, this ../../Gax.snk file doesn't seem to exist in this repo?

jskeet: It will when this PR is committed - it's the first file shown above, but with "binary file not shown" :)

chrisdunelm: Oh yes, my mistake :)

chrisdunelm:
LGTM

PR #9, opened 2016-02-29
First pass at myget support

jskeet:
It's a shame we need to use batch files, but the myget environment
appears to be a bit odd in terms of what's installed.

The batch files themselves are pretty ugly... this should definitely be a temporary hack.
(I've validated that the build fails if the tests fail though, which is pretty crucial :)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "projects": [ "src", "test" ],
   "sdk": {
-    "version": "1.0.0-rc1-final"
+    "version": "1.0.0-rc1-update1"
   }
 }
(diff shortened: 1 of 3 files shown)


chrisdunelm:
There doesn't appear to be anything myget-specific in this PR - this batch file builds & tests the code, then prepares a nuget package. This nuget package can then uploaded to whatever nuget feed we like. Is this correct?

You're right, this is a bit messy, and should be tidied later ;)

LGTM

jskeet:
It's myget-specific in that we've only added these batch files in order to support myget (the travis build is saner). I might look at using Cake at some point... probably prototyping it on Noda Time first :)

PR #10, opened 2016-03-02
First pass at PathTemplate and ResourceName.

jskeet:


jskeet:
Sorry for the additional commit - it became obvious it was required when using it...

jskeet:
Reassigned to Ivan while Chris recovers from illness.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+using System.Text;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Represents a path template used for resource names which may be composed of multiple IDs.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Templates use a subset of the syntax of the API platform. See the protobuf of <see cref="HttpRule"/> for
```
ivannaranjo: Where is `HttpRule`, could we point to the repo where it lives?

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// Just an array containing a single slash, to avoid constructing a new array every time we need
+        /// to split.
+        /// </summary>
+        private static readonly char[] SlashSplit = { '/' };
+
+        /// <summary>
+        /// List of segments in this template. Never modified after construction.
+        /// </summary>
+        private readonly IList<Segment> _segments;
+        /// <summary>
```
ivannaranjo: _Nit:_ maybe an empty line before the comment.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// List of the segments in this template which are wildcards. Never modified after construction.
+        /// </summary>
+        private readonly IList<Segment> _parameterSegments;
+
```
ivannaranjo: Could these then be `IReadOnlyList` instead of `IList` then?

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        private readonly string _originalTemplate;
+        private readonly bool _hasPathWildcard;
+
+        /// <summary>
+        /// The names of the parameters within the template. This collection has one element per parameter,
+        /// but unnamed parameters have a name of <c>null</c>.
+        /// </summary>
+        public ReadOnlyCollection<string> ParameterNames { get; }
```
ivannaranjo: Maybe use the interface `IReadOnlyCollection` instead so we don't tie ourselves to the particular implementation.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+
+        /// <summary>
+        /// Constructs a template from its textual representation, such as <c>shelves/*/books/**</c>.
+        /// </summary>
+        /// <param name="template">The textual representation of the template.</param>
+        public PathTemplate(string template)
```
ivannaranjo: Document that it cannot be `null`, use `Preconditions` to check.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+            if (serviceName == null)
+            {
+                return;
+            }
+            if (serviceName == "")
+            {
+                throw new ArgumentException("Service name cannot be empty", parameterName);
+            }
+            if (serviceName.Contains("/"))
+            {
```
ivannaranjo: Are there other restrictions on the name? Does it have to be a valid dns name?

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        /// <summary>
+        /// Validates a whole array of resource IDs, including that the count matches.
+        /// </summary>
+        internal void ValidateResourceIds(string[] resourceIds)
+        {
+            if (resourceIds.Length != ParameterCount)
+            {
+                throw new ArgumentException($"Expected {ParameterCount} ids, got {resourceIds.Length}");
```
ivannaranjo: Use the `resourceIds` name when throwing `ArgumentException`.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// <summary>
+        /// Validates that the given resource IDs are valid for this template, and returns a string representation
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This is equivalent to calling <c>new ResourceName(template, resourceIds).ToString()</c>, but simpler in
+        /// calling code and more efficient in terms of memory allocation.
+        /// </para>
+        /// <para>
+        /// This method assumes no service name is required. Call <see cref="ExpandWithService"/> to specify a service name.
```
ivannaranjo: Copy/pasted comment.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// </para>
+        /// </remarks>
+        /// <param name="resourceIds">The resource IDs to use to populate the parameters in this template.</param>
+        /// <returns>The string representation of the resource name.</returns>
```
ivannaranjo: Missing `serviceName` param.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        public string ExpandWithService(string serviceName, string[] resourceIds)
+        {
+            ValidateResourceIds(resourceIds);
```
ivannaranjo: Valdiate that resourceIds is not `null`.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// <remarks>
+        /// <para>
+        /// This is equivalent to calling <c>new ResourceName(template, resourceIds).ToString()</c>, but simpler in
+        /// calling code and more efficient in terms of memory allocation.
+        /// </para>
+        /// <para>
+        /// This method assumes no service name is required. Call <see cref="ExpandWithService"/> to specify a service name.
+        /// </para>
+        /// </remarks>
+        /// <param name="resourceIds">The resource IDs to use to populate the parameters in this template.</param>
```
ivannaranjo: Document that `resourceIds` cannot be `null`.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Although this method returns <c>null</c> if a name is passed in which doesn't match the template,
+        /// it still throws <see cref="ArgumentNullException"/> if <paramref name="name"/> is null, as this would
+        /// usually indicate a programming error rather than a data error.
+        /// </remarks>
+        /// <param name="name">The resource name to parse against this template.</param>
+        /// <returns>The parsed name as a <see cref="ResourceName"/>, or <c>null</c> if the name does not match this
+        /// template.</returns>
+        public ResourceName TryParseName(string name)
```
ivannaranjo: That is not the usual pattern in the BCL, it would typically be:
```C#
bool TryParseName(string name, out ResourceName result);
```
I don't particularly like that pattern either but is something to consider with respect of being idiomatic.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,439 @@
...
+
+        /// <summary>
+        /// Attempts to parse the given resource name against this template, returning <c>null</c> on failure.
+        /// </summary>
+        /// <remarks>
+        /// Although this method returns <c>null</c> if a name is passed in which doesn't match the template,
+        /// it still throws <see cref="ArgumentNullException"/> if <paramref name="name"/> is null, as this would
+        /// usually indicate a programming error rather than a data error.
+        /// </remarks>
+        /// <param name="name">The resource name to parse against this template.</param>
```
ivannaranjo: Document that `name` cannot be null.

src/Google.Api.Gax/Api/Gax/ResourceName.cs
```diff
@@ -0,0 +1,127 @@
...
+            set { this[GetParameterIndex(parameterName)] = value; }
+        }
+
+        /// <summary>
+        /// Creates a resource name with the given template and resource IDs.
+        /// The resource IDs are cloned, so later changes to <paramref name="resourceIds"/>
+        /// are ignored. This constructor does not populate the <see cref="ServiceName"/> property,
+        /// but that can be set after construction.
+        /// </summary>
+        public ResourceName(PathTemplate template, params string[] resourceIds)
```
ivannaranjo: No params documentation.

src/Google.Api.Gax/Api/Gax/ResourceName.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            Template = template;
```
ivannaranjo: No check that `template` is not `null`.

ivannaranjo:
Some comments, primarily on documentation and some style. The overall algorithm for parsing seems complex and I think grant another review on itself.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: The generated C# is in the same repo. Added comment to the .proto source.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Yes, although that doesn't really mean much.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: That doesn't signify as much though - `ReadOnlyCollection` really is read-only; you're not going to be able to cast it to anything else and add to it. `IReadOnlyList<T>` is really only "you can read from it, but you might be able to write to it too". If we were taking a dependency on the immutable collections, I'd use those here...

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: There may be in the future. Added a TODO.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done. Fortunately the calling public method always has a parameter called resourceIds :)

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done :)

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done (and validated...)

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done in ValidateResourceIds.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: In every case I can think of, that's used because the result can legitimately be a null reference *or* the return type is a value type. Will try to find some precedent/guidance about whether this case can be different.

src/Google.Api.Gax/Api/Gax/ResourceName.cs
jskeet: Done. (It was internal in a previous life...)

src/Google.Api.Gax/Api/Gax/ResourceName.cs
jskeet: Fixed, along with removing the embarrassing nature of the following lines...

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Uri.TryCreate uses an out parameter... amending this code to follow suit, much as I dislike it.

ivannaranjo:
LGTM :shipit:

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,440 @@
...
+    /// URL encoding is required.
+    /// </para>
+    /// </remarks>
+    public sealed class PathTemplate
+    {
+        /// <summary>
+        /// Just an array containing a single slash, to avoid constructing a new array every time we need
+        /// to split.
+        /// </summary>
+        private static readonly char[] SlashSplit = { '/' };
```
chrisdunelm: s_slashSplit
If we follow https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
```diff
@@ -0,0 +1,440 @@
...
+            PathWildcard,
+        }
+
+        /// <summary>
+        /// A segment of a path.
+        /// </summary>
+        private sealed class Segment
+        {
+            private static readonly Segment UnnamedWildcard = new Segment(SegmentKind.Wildcard, null);
+            private static readonly Segment UnnamedPathWildcard = new Segment(SegmentKind.PathWildcard, null);
```
chrisdunelm: s_unnamedWildcard
s_unnamedPathWildcard

jskeet: Done. I think I originally had these as internal properties, hence the different names...

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: After a bit of discussion, Ivan and I reckoned that `IReadOnlyList<T>` might be a better option after all, just so that we can update to immutable collections at a later date.

src/Google.Api.Gax/Api/Gax/PathTemplate.cs
jskeet: Done.

jskeet:
Have made the requested changes - PTAL, then I'll rebase and squash before merging, if you're then happy.

jskeet:
Actually, have rebased - but the commits are reasonably structured already, so will just merge with the 4 commits when Travis is happy.

PR #12, opened 2016-03-10
Helper code to reduce repetition in per-API libraries

jskeet:
(Needs tests, but wanted to get API shape right first)

### DIFF
--- a/src/Google.Api.Gax/Api/Gax/ClientHelper.cs
+++ b/src/Google.Api.Gax/Api/Gax/ClientHelper.cs
@@ -1,0 +1,162 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+using Google.Apis.Auth.OAuth2;
+using Grpc.Auth;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Common helper code shared by clients.
+    /// </summary>
+    public class ClientHelper
+    {
+        private readonly Func<DateTime?> _deadlineCalculator;
+
+        /// <summary>
+        /// Constructs a helper from the given settings. These are expected to be cloned
+        /// already, and not changed afterwards.
+        /// </summary>
+        /// <param name="settings">The service settings.</param>
+        public ClientHelper(ServiceSettingsBase settings)
+        {
+            _deadlineCalculator = GetDeadlineCalculator(settings);
+        }
+
+        /// <summary>
+        /// Returns a function which will compute a deadline suitable for the clock and timeout
+        /// currently in effect in the given settings.
+        /// </summary>
+        /// <remarks>
+        /// Later changes made to the clock and timeout in the settings will not affect the returned delegate.
+        /// </remarks>
+        /// <returns>A delegate which computes a deadline.</returns>
+        private static Func<DateTime?> GetDeadlineCalculator(ServiceSettingsBase settings)
+        {
+            var timeout = settings.Timeout;
+            if (timeout == null)
+            {
+                return () => null;
+            }
+            else
+            {
+                IClock clock = settings.Clock ?? SystemClock.Instance;
+                TimeSpan actualTimeout = timeout.Value;
+                // Important: this lambda expression doesn't capture settings
+                return () => clock.GetCurrentDateTimeUtc() + actualTimeout;
+            }
+        }
+
+        /// <summary>
+        /// Builds a suitable <see cref="CallOptions"/> taking service settings into account,
+        /// along with a call-specific cancellation token and override mechanism.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token to use for this call.</param>
+        /// <param name="callOptionsOverride">Optional options override delegate.</param>
+        /// <returns></returns>
+        public CallOptions BuildCallOptions(
+            CancellationToken cancellationToken,
+            Func<CallOptions, CallOptions> callOptionsOverride)
+        {
+            var callOptions = new CallOptions(
+                deadline: _deadlineCalculator(),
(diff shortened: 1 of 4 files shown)


src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <param name="suppliedCredentialScopes">The credential scopes supplied by the client, if any. These are cloned if not null.</param>
+        /// <param name="defaultCredentialScopes">The credential scopes to apply if none are supplied. These are not expected to change.</param>
+        /// <param name="clientFactory">The factory delegate which creates a client given the relevant credentials, service settings and endpoint.</param>
+        /// <returns>A task which, when completed, will have a result of the newly constructed client.</returns>
+        public static async Task<TClient> CreateFromDefaultCredentialsAsync<TClient, TSettings>(
+            TSettings settings,
+            ServiceEndpointSettings serviceEndpointSettings,
+            IEnumerable<string> suppliedCredentialScopes,
+            IEnumerable<string> defaultCredentialScopes,
+            Func<ChannelCredentials, TSettings, ServiceEndpointSettings, TClient> clientFactory)
```
chrisdunelm: Most of these parameters could be passed into the constructor - settings, serviceEndpointSettings, defaultCredentialScopes, clientFactory
The class would need making generic in TClient and TSettings, which seems fine to me.

src/Google.Api.Gax/Api/Gax/ServiceEndpointSettings.cs
```diff
@@ -60,5 +61,19 @@ public string Host
...
         };
+
+        /// <summary>
+        /// Creates a channel for the host and port in this object, with defaults for unspecified values.
+        /// </summary>
+        /// <param name="defaultHost">The host to use if <see cref="Host"/> is null. Must not be null.</param>
+        /// <param name="defaultPort">The port to use if <see cref="Port"/> is null.</param>
+        /// <param name="credentials">The credentials to use with the channel. Must not be null.</param>
+        /// <returns></returns>
+        public Channel CreateChannel(string defaultHost, int defaultPort, ChannelCredentials credentials)
```
chrisdunelm: Should this method be in the settings class? This puts functionality into a class that currently just expresses configuration.
I'd rather see this as a static method in ClientHelper.

As a user having this public method here might be confusing. ServiceEndpointSettings is a class that users are expected to directly interact with (via a concrete service-specific settings class, inherited from this), and having this public method here may well be confusing, not least because they might think this they are meant to call themselves to get a channel.
This method is purely for our internal use, so I think it's better suited on ClientHelper, which is also purely for our internal use.
(and on ServiceSettingsBase.GetDeadlineCalculator)

jskeet: Could do, sure. If it weren't for it being a public class that users would interact with, I think it would still make sense where it is - but yes, I'm happy to move it to avoid confusion.

jskeet: Done (pushing commit now.)

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -0,0 +1,146 @@
...
+            var timeout = settings.Timeout;
+            if (timeout == null)
+            {
+                return () => null;
+            }
+            else
+            {
+                IClock clock = settings.Clock ?? SystemClock.Instance;
+                TimeSpan actualTimeout = timeout.Value;
+                // Important: this lambda expression doesn't capture "this"
```
chrisdunelm: ...doesn't capture "settings"

src/Google.Api.Gax/Api/Gax/ServiceEndpointSettings.cs
chrisdunelm: Can this CreateChannel method be moved too? For the same reason as GetDeadlineCalculator

jskeet: Done, with corresponding commit in gcloud-dotnet as well.

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// Creates a channel for the host and port in this object, with defaults for unspecified values.
+        /// </summary>
+        /// <param name="endpoint">Endpoint settings to use. Must not be null.</param>
+        /// <param name="defaultHost">The host to use if <see cref="Host"/> is null. Must not be null.</param>
+        /// <param name="defaultPort">The port to use if <see cref="Port"/> is null.</param>
+        /// <param name="credentials">The credentials to use with the channel. Must not be null.</param>
+        /// <returns>A channel for the given endpoint with application default credentials.</returns>
+        public static Channel CreateChannel(ServiceEndpointSettings endpoint, string defaultHost, int defaultPort, ChannelCredentials credentials)
+        {
+            GaxPreconditions.CheckNotNull(defaultHost, nameof(endpoint));
```
chrisdunelm: endPoint, not defaultHost

jskeet: Gah! Thanks, fixing now.

src/Google.Api.Gax/Api/Gax/ServiceEndpointSettings.cs
```diff
@@ -5,6 +5,7 @@
  * https://developers.google.com/open-source/licenses/bsd
  */
 
+using Grpc.Core;
```
chrisdunelm: I guess this can be removed?

chrisdunelm:
LGTM

src/Google.Api.Gax/Api/Gax/ServiceEndpointSettings.cs
jskeet: Yup. I'll squash to one commit and remove any straggling bits like this or whitespace.

jskeet:
Okay, this now has tests where feasible - it would be nice to test the client setup, but that's tricky due to default credentials and channel setup. (We could use insecure credentials to test CreateChannel, but it's still going to make a network connection...)

chrisdunelm:
LGTM

PR #13, opened 2016-03-17
Provide CallSettings, and use it in ServiceBasesettings.

chrisdunelm:


src/Google.Api.Gax/Api/Gax/Expiration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        /// <summary>
+        /// If not null, the absolute deadline of this expiration.
+        /// </summary>
+        public DateTime? Deadline { get; }
+
+        /// <summary>
+        /// Does this <see cref="Expiration"/> contain a relative timeout?
+        /// </summary>
+        public bool IsTimeout => Timeout != null;
```
ivannaranjo: Nit: Would it be better to use the prefix _Has_ as in `HasTimeout` ?

src/Google.Api.Gax/Api/Gax/Expiration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        /// <summary>
+        /// An <see cref="Expiration"/> with no timeout or deadline.
+        /// </summary>
+        /// <remarks>
+        /// Indicates that no expiration is required.
+        /// </remarks>
+        public static Expiration None { get; } = new Expiration(null, null);
+
+        private Expiration(TimeSpan? timeout, DateTime? deadline)
```
ivannaranjo: Nit: Obvious disagreement here :) should have public constructors, having only the static methods is not idiomatic.

src/Google.Api.Gax/Api/Gax/MetadataExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    internal static class MetadataExtensions
+    {
+        public static void Add(this Metadata metadata, IEnumerable<Metadata.Entry> items)
```
ivannaranjo: List<T> has a method called `AddRange`, should we use that name? Not that it matters much since it is internal.

chrisdunelm: Calling it Add allows it to be used as a collection initializer, as in the Clone method below

jskeet: Side piece of trivia: this is the first time I've seen the C# 6 feature of having extension methods used by collection initializers. And I forgot to mention it in my recent talk :(

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -20,7 +20,9 @@ namespace Google.Api.Gax
     /// </summary>
     public class ClientHelper
     {
-        private readonly Func<DateTime?> _deadlineCalculator;
+        //private readonly Func<DateTime?> _deadlineCalculator;
```
jskeet: Remove?

src/Google.Api.Gax/Api/Gax/Expiration.cs
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+using System;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Expiration specified by relative timeout or absolute deadline.
+    /// </summary>
+    public class Expiration
```
jskeet: sealed

src/Google.Api.Gax/Api/Gax/Expiration.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        /// <summary>
+        /// Create an <see cref="Expiration"/> with a relative timeout.
+        /// </summary>
+        /// <param name="timeout">The relative timeout.</param>
+        /// <returns>An <see cref="Expiration"/> with the specified relative timeout.</returns>
+        /// <remarks>
+        /// Zero or negative timeouts are valid, and will cause immediate failure of the operation being performed.
+        /// </remarks>
+        public static Expiration WithTimeout(TimeSpan timeout) => new Expiration(timeout, null);
```
jskeet: "With" sounds like it's going to be an instance method to me. I'd use ForTimeout or FromTimeout. Or maybe even just Timeout.

src/Google.Api.Gax/Api/Gax/Expiration.cs
jskeet: I personally prefer the static methods here. If we exposed constructors, we'd want to throw an exception if both parameters are specified... if we only exposed two constructors (one for deadline, one for timeout) then you couldn't construct an expiration equivalent to None. We could expose a parameterless constructor as well, *just* for the sake of None, but that's pretty weird.

jskeet: Counterexample to "no constructors": everyone's used to calling `WebRequest.Create`

test/Google.Api.Gax.Tests/ClientHelperTest.cs
```diff
@@ -60,17 +67,76 @@ public void BuildCallOptions_PopulatesCancellationToken()
...
+            {
+                Headers = new Metadata { new Metadata.Entry("1", "one") },
+                Expiration = Expiration.WithDeadline(new DateTime(78L)),
+                CancellationToken = cts.Token,
+                WriteOptions = new WriteOptions(WriteFlags.BufferHint),
+                PropagationToken = null, // Un-creatable, un-mockable
+                Credentials = null, // Un-creatable, un-mockable
+            };
+            CallOptions callOptions = helper.BuildCallOptions(null, callSettings);
+            Assert.Equal(callSettings.Headers, callOptions.Headers);
```
jskeet: This should fail, because we should have cloned the headers.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
```diff
@@ -60,17 +67,76 @@ public void BuildCallOptions_PopulatesCancellationToken()
...
+                }
+            });
+            var overrideCancellationToken = new CancellationTokenSource().Token;
+            var callSettingsOverride = new CallSettings
+            {
+                CancellationToken = overrideCancellationToken
+            };
+            var argCancellationToken = new CancellationTokenSource().Token;
+
+            CallOptions callOptionsNoOverride = helper.BuildCallOptions(null, null);
```
jskeet: This feels like it's three tests, really...

jskeet:
Generally happy, but with the given nits. Sounds like we need to consult further around the expiration constructor vs factory methods.

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
chrisdunelm: Oops, done.

src/Google.Api.Gax/Api/Gax/Expiration.cs
chrisdunelm: Done

src/Google.Api.Gax/Api/Gax/Expiration.cs
chrisdunelm: Done with FromTimeout

src/Google.Api.Gax/Api/Gax/Expiration.cs
chrisdunelm: Not sure, using "Has" wouldn't really make sense for "HasNone", so using Has for the other two would be slightly inconsistent.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
chrisdunelm: Headers are now cloned.
Although this Assert.Equal() method is collection-aware, and compares the contents.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
chrisdunelm: OK, OK, now split into three tests.

jskeet:
LGTM modulo cloning which we'll discuss further.

PR #14, opened 2016-03-18
GetOrCreateHelper, for use in SimplePubsub in gcloud-dotnet

jskeet:
(I'm certain this will be useful beyond Pusub.)

// cc @ivannaranjo if you want to review it instead...

### DIFF
--- a/src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
+++ b/src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
@@ -1,0 +1,154 @@
+﻿/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Grpc.Core;
+using System;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Helper methods for client library code which provides "get or create" functionality,
+    /// i.e. fetching an existing resource if it exists, and creating it otherwise.
+    /// </summary>
+    public static class GetOrCreateHelper
+    {
+        private const int DefaultMaxAttempts = 3;
+
+        /// <summary>
+        /// Synchronously retrieves an existing resource or creates a new one if the 
+        /// resource doesn't already exist.
+        /// </summary>
+        /// <remarks>
+        /// <para>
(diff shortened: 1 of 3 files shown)


src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    try
+                    {
+                        return creator();
+                    }
+                    catch (RpcException creationException) when (i < maxAttempts - 1 && creationException.Status.StatusCode == StatusCode.AlreadyExists)
+                    {
+                        // Probably a race condition... someone else tried to create it just when we did.
+                        // Go back through the loop and check we can get it this time, failing after this happens too often.
+                    }
```
ivannaranjo: Should we have a bit of a wait before going back to see if we can fetch the result? Wondering there can be a bit of a delay between the creation of a resource and been able to fetch it.

src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
```diff
@@ -0,0 +1,153 @@
...
+            int maxAttempts = DefaultMaxAttempts)
+        {
+            GaxPreconditions.CheckNotNull(fetcher, nameof(fetcher));
+            GaxPreconditions.CheckNotNull(creator, nameof(creator));
+            GaxPreconditions.CheckArgumentRange(maxAttempts, nameof(maxAttempts), 1, int.MaxValue);
+            for (int i = 0; i < maxAttempts; i++)
+            {
+                try
+                {
+                    var resource = await fetcher().ConfigureAwait(false);
```
ivannaranjo: Since you are calling user provided callbacks here, we should either document the fact that they can be called on different threads and/or don't user `ConfigureAwait` to ensure that they are called in the same context.

ivannaranjo:
Few comments, one about threading, otherwise looking good.

src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
jskeet: No, I'd expect this to be okay, to be honest - an update not being applied universally is one thing but the presence/absence I'd expect to be atomic. Can tweak later if necessary though :)

src/Google.Api.Gax/Api/Gax/GetOrCreateHelper.cs
jskeet: Righto - will add something into the remarks.

PR #15, opened 2016-03-18
Update to the latest Google.Apis.Auth

jskeet:


jskeet:
(Self-merging when Travis is green in the hope of fixing the myget build)

PR #16, opened 2016-03-22
Removeclones

chrisdunelm:


src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -24,8 +24,8 @@ public class ClientHelper
         private readonly CallSettings _globalCallSettings;
 
         /// <summary>
-        /// Constructs a helper from the given settings. These are expected to be cloned
-        /// already, and not changed afterwards.
+        /// Constructs a helper from the given settings.
+        /// Behaviour is undefined if settings are changed afterwards
```
jskeet: Bring back the full stop :)
(I'd also say "after construction" rather than just "afterwards")

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
```diff
@@ -103,14 +103,13 @@ public static Channel CreateChannel(ServiceEndpointSettings endpoint, string def
         }
 
         /// <summary>
-        /// Creates a client using application default credentials, taking care of generic cloning and
-        /// credential scoping.
+        /// Creates a client using application default credentials, taking care of credential scoping.
```
jskeet: I would remove the parameter-specific documentation, and just repeat the "The behavior undefined if..."

jskeet:
Looks good so far - as we've discussed, we can remove the abstract base class from service settings now.

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
chrisdunelm: Done.
(although I was hoping to subtly show its UK heritage here)

src/Google.Api.Gax/Api/Gax/ClientHelper.cs
chrisdunelm: Done

chrisdunelm:
Generic ServiceSettingsBase class now removed

jskeet:
LGTM. I wouldn't be *shocked* if we ended up deciding to put the generic base class back again, but I'm definitely in favour of reverting to simplicity for the moment.

PR #19, opened 2016-03-31
Remove bundling for now.

jskeet:
It's likely we'll revert this change later, but while we're not *using* bundling, it's simpler for it to not be in the API.


chrisdunelm:
LGTM

PR #20, opened 2016-03-31
First pass at retry mechanism.

jskeet:
This changes the definition of ApiCallable to accept a CallSettings.
We may well want to split ApiCallable into one synchronous and one asynchronous
delegate at some point. This commit is the asynchronous side.

This also introduces a fake clock and a fake scheduler, in a new Testing package.


src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// <summary>
+        /// The multiplier to apply to the delay on each iteration; must be greater than or equal to 1.0.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The default value of this property is 1.0.
+        /// </para>
+        /// <para>
+        /// As an example, a mulitplier of 2.0 with a delay of 0.1s on an RPC which fails three times before
+        /// succeeding would lead to an initial delay between the first response and the second request
```
ivannaranjo: mulitplier => multiplier

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        /// <summary>
+        /// Works out the next delay from the current one, based on the multiplier and maximum.
+        /// </summary>
+        internal TimeSpan NextDelay(TimeSpan currentDelay)
+        {
+            checked
+            {
+                TimeSpan next = new TimeSpan((long) (currentDelay.Ticks * DelayMultiplier));
+                return next < MaxDelay ? next : MaxDelay;
```
ivannaranjo: May be `return Math.Min(next, MaxDelay);`

src/Google.Api.Gax/Api/Gax/GaxPreconditions.cs
```diff
@@ -189,5 +190,12 @@ public static void CheckArgument<T>(bool condition, string paramName, string for
                 throw new ArgumentException(string.Format(format, arg0, arg1), paramName);
             }
         }
+
+        internal static void CheckNotNull<TRequest, TResponse>(ApiCallable<TRequest, TResponse> originalCall)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            throw new NotImplementedException();
```
ivannaranjo: This is temporary, wondering if this should have also the name of the parameter?

src/Google.Api.Gax/Api/Gax/IScheduler.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Abstraction of scheduler-like operations, used for testability.
+    /// </summary>
+    /// <remarks>
+    /// Note that this is different to <see cref="TaskScheduler"/>, which is really involved
+    /// with assigning tasks to threads rather than any sort of delay.
+    /// </remarks>
+    public interface IScheduler
```
ivannaranjo: How about `IDelayScheduler` to not be confused with a general purpose scheduler?

src/Google.Api.Gax/Api/Gax/ApiCallableExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+            RetrySettings settings,
+            IClock clock,
+            IScheduler scheduler)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            GaxPreconditions.CheckNotNull(originalCall, nameof(originalCall));
+            GaxPreconditions.CheckNotNull(clock, nameof(clock));
+            scheduler = scheduler ?? SystemScheduler.Instance;
+            GaxPreconditions.CheckNotNull(settings, nameof(settings));
```
chrisdunelm: Put this precondition check with the others? And in the same order as the parameters?

src/Google.Api.Gax/Api/Gax/IScheduler.cs
jskeet: Not sure about *that* alternative name, but definitely happy to consider alternatives.

src/Google.Api.Gax/Api/Gax/GaxPreconditions.cs
jskeet: Ick, this was a mistake - not sure how it happened at all. Will fix.

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
jskeet: Done, thanks.

src/Google.Api.Gax/Api/Gax/RetrySettings.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// A jitter which simply returns the specified maximum delay.
+        /// </summary>
+        public static IJitter NoJitter { get; } = new NoJitterImpl();
+
+        // TODO: Is this a reasonable default? Does it make sense to have a default at all?
+
+        /// <summary>
+        /// The default retry filter, which retries operations which fail due to a status code of "not found".
+        /// </summary>
+        public static Predicate<RpcException> DefaultFilter { get; } = FilterForStatusCodes(StatusCode.NotFound);
```
ivannaranjo: I thought there was also a status code that meant try again or something like that. Not found I guess it means 404? That doesn't seem like a good general filter though, if a file doesn't exist it doesn't necessary mean that it is going to exist later.

src/Google.Api.Gax/Api/Gax/ApiCallableExtensions.cs
jskeet: Reordered now - it was previously in the right order, then I reordered the parameters but not the validation.

src/Google.Api.Gax/Api/Gax/RetrySettings.cs
```diff
@@ -0,0 +1,155 @@
...
+            // TODO: Take a copy? Optimize for common cases?
+            return ex => statusCodes.Contains(ex.Status.StatusCode);
+        }
+
+        private sealed class RandomJitterImpl : IJitter
+        {
+            private static int seed = Environment.TickCount;
+
+            private static ThreadLocal<Random> randomWrapper = new ThreadLocal<Random>(() =>
+                new Random(Interlocked.Increment(ref seed))
```
jskeet: It means that if lots of instances are created in very swift succession, we don't end up with them all having the same seed.

src/Google.Api.Gax/Api/Gax/ApiCallableExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// <typeparam name="TRequest">Request message type</typeparam>
+        /// <typeparam name="TResponse">Response message type</typeparam>
+        /// <param name="originalCall">Original callable delegate</param>
+        /// <param name="settings">Retry settings</param>
+        /// <param name="clock">Clock to consult for retry timing</param>
+        /// <param name="scheduler">Scheduler to use for delays</param>
+        /// <returns>A retrying equivalent of <paramref name="originalCall"/></returns>
+        public static ApiCallable<TRequest, TResponse> WithRetry<TRequest, TResponse>(
+            this ApiCallable<TRequest, TResponse> originalCall,
+            RetrySettings settings,
```
chrisdunelm: Could this be called "retrySettings" rather than just "settings"? I find it a little confusing with all the different types of settings!

jskeet: Good plan, will do.

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+                _maxDelay = value;
+            }
+        }
+
+        /// <summary>
+        /// Creates a clone of this instance.
+        /// </summary>
+        /// <returns>A clone of this instance.</returns>
+        public BackoffSettings Clone() =>
+            new BackoffSettings
```
chrisdunelm: nit : Don't line-split here?
public BackoffSettings Clone() => new BackoffSettings

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// between the third response and the fourth request.
+        /// </para>
+        /// </remarks>
+        /// <exception cref="ArgumentOutOfRangeException">The value is not a real number, or is less than 1.</exception>
+        public double DelayMultiplier
+        {
+            get { return _delayMultiplier; }
+            set
+            {
+                if (value < 1 || double.IsInfinity(value) || double.IsNaN(value))
```
chrisdunelm: nit: 1.0 as it's a double?
nit: Don't check for infinity here? Something like Double.MaxValue would have same effect as infinity, so why single out infinity to be invalid?

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
jskeet: My evolved convention (which you can probably find exceptions for :) is

"If it all fits on one line, do so. Otherwise, break after =>"

I don't think there's anything like a widespread convention for this yet.

src/Google.Api.Gax/Api/Gax/GaxPreconditions.cs
```diff
@@ -6,6 +6,7 @@
  */
 
 using System;
+using Google.Protobuf;
```
chrisdunelm: Why has this been added? There don't appear to be any other changes in this file?

src/Google.Api.Gax/Api/Gax/IScheduler.cs
```diff
@@ -0,0 +1,44 @@
+﻿/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+using System.Runtime.CompilerServices;
```
chrisdunelm: Is CompilerServices required?

jskeet: I think that was when we were experimenting with TaskAwaitable (or whatever). Will zap.

src/Google.Api.Gax/Api/Gax/GaxPreconditions.cs
jskeet: Humbug. That was due to a *different* change that was unwanted. After squashing, we should end up not having any changes in here. Sorry!

src/Google.Api.Gax/Api/Gax/IScheduler.cs
```diff
@@ -0,0 +1,44 @@
...
+    public interface IScheduler
+    {
+        /// <summary>
+        /// Returns a task which will complete after the given delay. Whether the returned
+        /// awaitable is configured to capture the current context or not is implementation-specific.
+        /// (A test implementation may capture the current context to enable reliable testing.)
+        /// </summary>
+        /// <param name="delay">Time to delay for. Must not be negative.</param>
+        /// <returns>A task which will complete after the given delay.</returns>
+        Task Delay(TimeSpan delay);
```
chrisdunelm: This method isn't strictly required, as it looks like the Schedule() method could be used with an empty action.
Just want to check that you consider it's definitely worth keeping this method.

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
jskeet: Fair and fair. I suspect we may have issues if the value is too large in various ways, but they wouldn't be restricted to infinity.

src/Google.Api.Gax/Api/Gax/RetrySettings.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <returns>A retry filter based on status codes.</returns>
+        public static Predicate<RpcException> FilterForStatusCodes(params StatusCode[] statusCodes) =>
+            FilterForStatusCodes((IEnumerable<StatusCode>) statusCodes);
+
+        /// <summary>
+        /// Creates a retry filter based on status codes.
+        /// </summary>
+        /// <param name="statusCodes">The status codes to retry. Must not be null.</param>
+        /// <returns>A retry filter based on status codes.</returns>
+        public static Predicate<RpcException> FilterForStatusCodes(IEnumerable<StatusCode> statusCodes)
```
chrisdunelm: Strictly speaking, this might be better as an IReadOnlySet<StatusCode>, if only such a thing existed.

src/Google.Api.Gax/Api/Gax/IScheduler.cs
jskeet: Hmm. I think it's worth keeping, as it makes both the prod implementation (`Task.Delay`) and calling code simpler.

src/Google.Api.Gax/Api/Gax/SystemScheduler.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Singleton implementation of <see cref="IScheduler"/> which uses <see cref="Task.Delay"/>
+    /// and <see cref="Task.Run"/> internally.
+    /// </summary>
+    public sealed class SystemScheduler : IScheduler
+    {
+        public static SystemScheduler Instance { get; } = new SystemScheduler();
+
+        private SystemScheduler()
+        {
+        }
```
chrisdunelm: Can we put all this on one line, in the case of empty constructors?

src/Google.Api.Gax/Api/Gax/RetrySettings.cs
jskeet: It doesn't, and in particular I'd expect the set to be really small (1-3 codes), so just a linear search should be fine.

src/Google.Api.Gax/Api/Gax/SystemScheduler.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public Task Delay(TimeSpan timeSpan)
```
chrisdunelm: expression-bodied method?

jskeet: Done.

src/Google.Api.Gax/Api/Gax/SystemScheduler.cs
jskeet: We certainly can. I'm not a massive fan myself, but I don't mind it...

src/Google.Api.Gax/Api/Gax/BackoffSettings.cs
chrisdunelm: Hmm, OK, my personal convention is currently
"If the code after the => nicely fits on one line and looks nice, then stick it all on one line"
At some point soon-ish we'll probably want our conventions to converge!

src/Google.Api.Gax/Api/Gax/IScheduler.cs
chrisdunelm: OK, fair enough.

src/Google.Api.Gax/Api/Gax/RetrySettings.cs
chrisdunelm: OK

chrisdunelm:
LGTM

jskeet:
Rebased, although I'm not sure why it thinks there are 2 commits instead of 1...

PR #21, opened 2016-03-31
Move source files around

jskeet:
Given that most source editing is the manually written code in Google.Api.Gax, it makes sense for those files to be at the top-level,
and all the generated code be more nested.

We could potentially move the generated folder to be a peer of src, but that would be a little odd.

(The content of the source files have not changed.)


chrisdunelm:
Or we could use two top-level directories, "generated" and "hand-coded (or whatever)"?
This more cleanly separates the two, although I'm not sure it's better.

jskeet:
Yes, I'd considered that - but I think I prefer this approach. Aside from anything else, the location of the files in `src` mirrors the location of tests in `test` - where obviously there aren't any tests for generated code.

chrisdunelm:
LGTM

PR #25, opened 2016-04-07
Make GAX use Grpc 0.13.1, which has better native extension handling

jskeet:
The direct dependency on the native library is unnecessary because Grpc.Core depends on it already.


chrisdunelm:
LGTM

PR #26, opened 2016-04-13
Further support for retry

chrisdunelm:
Change APICallable delegateto ApiCall class, an make it handle async and
sync calls.
Extend retry to provide a sync method as well as async.
Provide IScheduler.Sleep() call (sync version of existing Delay).
Alter PageStreaming to simplify code-gen'ed code.
Fix FakeScheduler bug.
Update tests.


src/Google.Api.Gax/ApiCall.cs
```diff
@@ -0,0 +1,49 @@
...
+        where TRequest : class, IMessage<TRequest>
+        where TResponse : class, IMessage<TResponse>
+    {
+        /// <summary>
+        /// Delegate representing a single asynchronous API call with a request and <see cref="CallSettings"/>.
+        /// </summary>
+        /// <param name="request">Request to make to the service.</param>
+        /// <param name="callSettings">The (optional) settings to apply to this call.</param>
+        /// <returns>A task representing the API call; the result of the completed task will be the API response.</returns>
+        public delegate Task<TResponse> AsyncCall(TRequest request, CallSettings callSettings);
```
jskeet: Let's see whether it's nicer having them nested like this, or generic top-level delegates.

src/Google.Api.Gax/ApiCall.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <typeparam name="TResponse">Response type, which must be a protobuf message.</typeparam>
+    public sealed class ApiCall<TRequest, TResponse>
+        where TRequest : class, IMessage<TRequest>
+        where TResponse : class, IMessage<TResponse>
+    {
+        /// <summary>
+        /// Delegate representing a single asynchronous API call with a request and <see cref="CallSettings"/>.
+        /// </summary>
+        /// <param name="request">Request to make to the service.</param>
+        /// <param name="callSettings">The (optional) settings to apply to this call.</param>
```
jskeet: Rather than "(optional)", I'd say that the value can be null, and what the effect of that is.

src/Google.Api.Gax/ApiCall.cs
```diff
@@ -0,0 +1,49 @@
...
+        public delegate Task<TResponse> AsyncCall(TRequest request, CallSettings callSettings);
+        /// <summary>
+        /// Delegate representing a single synchronous API call with a request and <see cref="CallSettings"/>.
+        /// </summary>
+        /// <param name="request">Request to make to the service.</param>
+        /// <param name="callSettings">The (optional) settings to apply to this call.</param>
+        /// <returns>The API response.</returns>
+        public delegate TResponse SyncCall(TRequest request, CallSettings callSettings);
+
+        public ApiCall(AsyncCall async, SyncCall sync)
```
jskeet: I'd suggest changing the names to asyncCall and syncCall, partly to avoid using a contextual keyword as an identifier. (It works, but it can screw up syntax highlighting and probably mental reading...)

src/Google.Api.Gax/ApiCall.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Async = async;
+            Sync = sync;
+        }
+
+        /// <summary>
+        /// The asynchronous API call.
+        /// </summary>
+        public AsyncCall Async { get; }
```
jskeet: Possibly call this AsyncCall as well? (Can't do that while the delegate is nested of course.) Ditto SyncCall.

src/Google.Api.Gax/ApiCall.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Delegate representing a single synchronous API call with a request and <see cref="CallSettings"/>.
+        /// </summary>
+        /// <param name="request">Request to make to the service.</param>
+        /// <param name="callSettings">The (optional) settings to apply to this call.</param>
+        /// <returns>The API response.</returns>
+        public delegate TResponse SyncCall(TRequest request, CallSettings callSettings);
+
+        public ApiCall(AsyncCall async, SyncCall sync)
+        {
+            Async = async;
```
jskeet: Check for nullity, and document non-nullity on properties.

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,103 @@
+﻿using Google.Protobuf;
```
jskeet: License header, here and on other files.

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    public static class ApiCallRetryExtensions
```
jskeet: We should probably discuss the layout/naming of the extension methods at some point, but that can come later. (It'll be a binary breaking change, but not a source one.)

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        public static ApiCall<TRequest, TResponse> WithRetry<TRequest, TResponse>(
+            this ApiCall<TRequest, TResponse> originalCall,
+            RetrySettings retrySettings,
+            IClock clock,
+            IScheduler scheduler)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            if (retrySettings == null)
```
jskeet: This makes the precondition a few lines later a bit odd. Do we ever actually want to call this without a retry settings?

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+                return originalCall;
+            }
+
+            GaxPreconditions.CheckNotNull(originalCall, nameof(originalCall));
+            GaxPreconditions.CheckNotNull(retrySettings, nameof(retrySettings));
+            retrySettings.Validate(nameof(retrySettings));
+            GaxPreconditions.CheckNotNull(clock, nameof(clock));
+            scheduler = scheduler ?? SystemScheduler.Instance;
+
+            ApiCall<TRequest, TResponse>.AsyncCall asyncCall = async (request, callSettings) =>
```
jskeet: I'd possibly add a comment explaining the code duplication. (It's probably still the right decision, but worth explaining it.)

src/Google.Api.Gax/ClientHelper.cs
```diff
@@ -54,33 +55,51 @@ public ClientHelper(ServiceSettingsBase settings)
...
+        /// <param name="asyncGrpcCall">The underlying synchronous gRPC call.</param>
+        /// <param name="syncGrpcCall">The underlying asynchronous gRPC call.</param>
+        /// <returns></returns>
+        public ApiCall<TRequest, TResponse> BuildApiCall<TRequest, TResponse>(
+            Func<TRequest, CallOptions, AsyncUnaryCall<TResponse>> asyncGrpcCall,
+            Func<TRequest, CallOptions, TResponse> syncGrpcCall)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            ApiCall<TRequest, TResponse>.AsyncCall asyncCall = (request, callSettings) =>
```
jskeet: We might consider making BuildCallOptions a static call, and hoisting the _clientCallSettings variable into a local variable here, to make the lambda expression capture more efficient. Not just yet though :)

src/Google.Api.Gax/PageStreamer.cs
```diff
@@ -66,19 +69,23 @@ public sealed class PageStreamer<TResource, TRequest, TResponse, TToken>
...
         /// that token is maintained. This is also passed to the request provider along with a token
         /// to provide another request. (The provider may build a new object, or modify the existing request.)</param>
         /// <param name="responseFetcher">A function to fetch a response containing a page of results given a request.</param>
-        /// <returns>A sequence of resources, which are fetched a page at a time. Must not be null.</returns>
-        public IEnumerable<TResource> Fetch(TRequest initialRequest, Func<TRequest, TResponse> responseFetcher)
+        /// <returns>A sequence of resources, which are fetched a page at a time. Will not be null.</returns>
+        public IEnumerable<TResource> Fetch(
+            CallSettings callSettings,
+            TRequest initialRequest,
+            ApiCall<TRequest, TResponse>.SyncCall responseFetcher)
```
jskeet: It's quite tempting to make both methods just take an `ApiCall<TRequest, Response>`...

src/Google.Api.Gax/PageStreamer.cs
```diff
@@ -91,25 +98,35 @@ public IEnumerable<TResource> Fetch(TRequest initialRequest, Func<TRequest, TRes
...
             GaxPreconditions.CheckNotNull(responseFetcher, nameof(responseFetcher));
             TRequest nextRequest = initialRequest;
             bool done = false;
+            CallSettings rpcCallSettings = callSettings?.Clone() ?? new CallSettings();
             return new PageStreamingAsyncEnumerable<TResource>(async cancellationToken =>
             {
-                cancellationToken.ThrowIfCancellationRequested();
+                // If MoveNext() has been passed a non-default cancellation-token, then use it.
+                // Otherwise, use the callSettings cancellation-token (if it has one).
+                CancellationToken? rpcCancellationToken = cancellationToken != default(CancellationToken) ?
```
jskeet: I suspect we want this check elsewhere, too - we might want to add extension methods to make it simpler. Not yet though.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
```diff
@@ -20,11 +20,11 @@ public void BuildCallOptions_NoTimeout()
         {
             var mockClock = new Mock<IClock>();
             var helper = new ClientHelper(new DummySettings { Clock = mockClock.Object });
-            var options = helper.BuildCallOptions(CancellationToken.None, null);
+            var options = helper.BuildCallOptions(null);
             Assert.Null(options.Deadline);
             mockClock.Verify(c => c.GetCurrentDateTimeUtc(), Times.Never);
         }
-
+
```
jskeet: Remove whitespace-only changes? I don't care too much, to be honest.

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: The "Call" suffix is possibly unnecessary here as they are inside a class ApiCall, which already has "Call" in its name.

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Acknowledged.

src/Google.Api.Gax/ClientHelper.cs
chrisdunelm: Acknowledged.

src/Google.Api.Gax/PageStreamer.cs
chrisdunelm: Acknowledged.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
chrisdunelm: Done. Added this whitespace by accident.

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: Done

test/Google.Api.Gax.Tests/RetryTest.cs
```diff
@@ -31,8 +32,30 @@ public class RetryTest
             MaxDelay = TimeSpan.FromTicks(6000) // Irrelevant
         };
 
-        [Fact]
-        public void FirstCallSucceeds()
+        private static Task<TResponse> Call<TRequest, TResponse>(
```
jskeet: Again, I'd avoid calling the parameter async if possible. (Even just `asynchronous` would be better.) Maybe I'm being too picky though.

testing/Google.Api.Gax.Testing/FakeScheduler.cs
```diff
@@ -49,8 +49,15 @@ public FakeScheduler() : this(new FakeClock(0L))
...
         public Task Delay(TimeSpan delay)
         {
-            var tcs = new TaskCompletionSource<int>();
-            return Schedule(() => tcs.SetResult(0), delay);
+            Task task = Schedule(delegate { }, delay);
+            return task;
+        }
+
+        /// <inheritdoc />
+        public void Sleep(TimeSpan delay)
```
jskeet: As discussed, this isn't quite right :)

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: It'll make the code-gen'ed code slightly nicer if it transparently handles nulls.
E.g. It'll be able to do this:
_createTopic = _clientHelper.BuildApiCall(...).WithRetry(effectiveSettings.CreateTopicRetry)
And if effectiveSettings.CreateTopicRetry has been set to null (by user code) then it all "just works" correctly.

testing/Google.Api.Gax.Testing/FakeScheduler.cs
chrisdunelm: Maybe so, but it works ;)
(but I'll fix it anyway)

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Re-arranged precondition checks slightly to make more sense.

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: OK to leave like this for now?
We can change it later if this gets unwieldy.

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/PageStreamer.cs
chrisdunelm: Agreed :) Done.

src/Google.Api.Gax/ApiCall.cs
jskeet: Sure.

src/Google.Api.Gax/ApiCall.cs
jskeet: True - it just makes for an odd property name, I think. Let's leave it for now, but consider revisiting later.

test/Google.Api.Gax.Tests/RetryTest.cs
chrisdunelm: In my very humble opinion you're being perhaps maybe just a tiny bit too picky ;)
But please feel free to correct me on this -  and I'll change to code faster than a speeding async call.

chrisdunelm:
PTAL

test/Google.Api.Gax.Tests/FakeSchedulerTest.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [Fact]
+        public void SynchronousSleep()
+        {
+            var scheduler = new FakeScheduler();
+            scheduler.Run<Object>(() =>
+            {
+                scheduler.Sleep(TimeSpan.FromSeconds(1));
+                scheduler.Run();
```
jskeet: We don't want this Run call, do we?

test/Google.Api.Gax.Tests/FakeSchedulerTest.cs
```diff
@@ -0,0 +1,45 @@
...
+                return null;
+            });
+            // Passes if it completes.
+        }
```
jskeet: Maybe assert that the fake clock is at 1 second?

jskeet:
Just a couple more comments.

PR #28, opened 2016-04-18
First pass at user agent work. No codegen changes required :)

jskeet:


src/Google.Api.Gax/ServiceSettingsBase.cs
```diff
@@ -44,6 +58,7 @@ public abstract class ServiceSettingsBase
         {
             settings.CallSettings = CallSettings.Clone();
             settings.Clock = Clock;
+            settings.UserAgent = UserAgent;
```
chrisdunelm: I suspect this line isn't actually required, as the constructor will already have set it.
Could possibly replace this line with a comment saying why it's not necessary?

src/Google.Api.Gax/UserAgentBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// containing the given type.
+        /// </summary>
+        internal UserAgentBuilder AppendAssemblyVersion(string name, System.Type type)
+            => AppendVersion(name, FormatAssemblyVersion(type));
+
+        /// <summary>
+        /// Appends the .NET environment information to the list.
+        /// </summary>
+        internal UserAgentBuilder AppendDotNetEnvironment()
+            => AppendVersion("dotnet", FormatVersion(Environment.Version));
```
chrisdunelm: Comment that this will need reworking slightly to support Core?

src/Google.Api.Gax/ServiceSettingsBase.cs
jskeet: Ah, hadn't spotted that - yes, of course. Bit of a pain that it will have to regenerate it on each clone though. Can we avoid that? How often do we clone service settings?

src/Google.Api.Gax/ServiceSettingsBase.cs
```diff
@@ -19,6 +20,19 @@ namespace Google.Api.Gax
     /// </remarks>
     public abstract class ServiceSettingsBase
     {
+        protected ServiceSettingsBase()
+        {
+            UserAgent = new UserAgentBuilder()
+                .AppendDotNetEnvironment()
+                .AppendAssemblyVersion("gax", typeof(CallSettings))
```
jskeet: Just thinking - I wonder if this should be gax-dotnet. Might want to check with Omar... ditto grpc

src/Google.Api.Gax/UserAgentBuilder.cs
jskeet: Done (in unpushed code).

src/Google.Api.Gax/ServiceSettingsBase.cs
jskeet: Thinking about this more, we could remove CloneInto but add two constructors: one parameterless one, and one that accepts a ServiceSettingsBase to clone from.

chrisdunelm: Yes, that sounds like the best way to do it. The code-gen will have to change, but only slightly.

jskeet:
@chrisdunelm See what you think of the second commit.

test/Google.Api.Gax.Tests/ClientHelperTest.cs
```diff
@@ -127,7 +127,13 @@ private string RemoveUserAgent(Metadata metadata)
 
         private class DummySettings : ServiceSettingsBase
         {
-            public DummySettings Clone() => CloneInto(new DummySettings());
+            public DummySettings()
+            {
```
chrisdunelm: Nit: Put open+close curly brackets on line above, as with the constructor below.

test/Google.Api.Gax.Tests/ServiceSettingsBaseTest.cs
```diff
@@ -15,7 +15,13 @@ public class ServiceSettingsBaseTest
     {
         class TestSettings : ServiceSettingsBase
         {
-            public TestSettings Clone() => CloneInto(new TestSettings());
+            public TestSettings()
+            {
```
chrisdunelm: Nit: Put open+close curly brackets on line above, as with the constructor below.

chrisdunelm:
LGTM

PR #31, opened 2016-05-03
Retry & settings refactor

chrisdunelm:
Include retry settings in CallSettings.
Support per-method CallSettings.


src/Google.Api.Gax/ApiCall.cs
```diff
@@ -6,54 +6,75 @@
...
-        public delegate TResponse SyncCall(TRequest request, CallSettings callSettings);
-
-        /// <summary>
-        /// Initializes a new <see cref="ApiCall"/> with the specified asynchronous and synchronous calls.
-        /// </summary>
-        /// <param name="asyncCall">The asynchronous API call. Must not be null.</param>
-        /// <param name="syncCall">The synchronous API call. Must not be null.</param>
-        /// <exception cref="ArgumentNullException">Either specified call is null.</exception>
-        public ApiCall(AsyncCall asyncCall, SyncCall syncCall)
+        internal ApiCall(
```
jskeet: Thought for another round of refactoring - maybe we should call this `ApiOperation` or even `ApiMethod`, then CallSync and CallAsync methods, so they're verbs again. Not for now though :)

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -7,89 +7,87 @@
 
 using Google.Protobuf;
 using Grpc.Core;
+using System;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Google.Api.Gax
 {
     public static class ApiCallRetryExtensions
```
jskeet: Maybe move this into ApiCall now?

src/Google.Api.Gax/ApiCallExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    static class ApiCallExtensions
```
jskeet: Nit - presumably you want this to be internal (given that it's not public). Let's make that explicit.

src/Google.Api.Gax/ApiCallExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static Func<A, CallSettings, R> MapArg<A, B, R>(
```
jskeet: Docs and make them internal?

src/Google.Api.Gax/ApiCallExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Grpc.Core;
```
jskeet: Licence header.

src/Google.Api.Gax/ApiCallExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Google.Api.Gax
+{
+    static class ApiCallExtensions
+    {
+        public static Func<A, CallSettings, R> MapArg<A, B, R>(
+            this Func<A, B, R> f, Func<CallSettings, B> fMap) =>
+            (a, callSettings) => f(a, fMap(callSettings));
+
+        public static Func<TRequest, CallOptions, Task<TResponse>>
```
jskeet: Unwrap the method name?

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -62,6 +73,65 @@ public sealed class CallSettings
             WriteOptions = WriteOptions,
             PropagationToken = PropagationToken,
             Credentials = Credentials,
+            RetrySettings = RetrySettings?.Clone(),
         };
+
+        // Merge other into this.
+        // Other is unchanged. Returns this
+        // Internal? Or Public?
+        public CallSettings Merge(CallSettings other)
```
jskeet: Probably internal for now.

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -62,6 +73,65 @@ public sealed class CallSettings
             WriteOptions = WriteOptions,
             PropagationToken = PropagationToken,
             Credentials = Credentials,
+            RetrySettings = RetrySettings?.Clone(),
         };
+
+        // Merge other into this.
```
jskeet: Will we always have one of the two as non-null? I was wondering whether a static method might be more flexible.
Also, I'd expected (in my head) for Merge to return a new one instead of merging it into an existing one. I suspect that would allow less defensive cloning - but I admit I haven't looked through carefully. Probably something for another PR after pondering.

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -62,6 +73,65 @@ public sealed class CallSettings
...
+        // Other is unchanged. Returns this
+        // Internal? Or Public?
+        public CallSettings Merge(CallSettings other)
+        {
+            if (other == null)
+            {
+                return this;
+            }
+            if (other.Headers != null)
```
jskeet: I thought our header merging was going to be additive?

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -62,6 +73,65 @@ public sealed class CallSettings
...
+            {
+                Headers = other.Headers.Clone();
+            }
+            if (other.Expiration != null)
```
jskeet: All of the replace properties could be written as

```csharp
Expiration = other.Expiration ?? Expiration;
// etc
```

... which has a pointless assignment, but would be a lot shorter, and easier to see ones which weren't just replacements.

(Of course, if we make merge return a new instance instead, the assignment wouldn't be pointless...)

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -62,6 +73,65 @@ public sealed class CallSettings
...
+
+        internal CallOptions ToCallOptions(IClock clock) => new CallOptions(
+            headers: Headers,
+            deadline: Expiration.CalculateDeadline(clock),
+            cancellationToken: CancellationToken ?? default(CancellationToken),
+            writeOptions: WriteOptions,
+            propagationToken: PropagationToken,
+            credentials: Credentials);
+
+        internal CallSettings AddUserAgent(string userAgent)
```
jskeet: I can't remember where this gets called from now, but it's likely to be relevant with the header overwriting.

src/Google.Api.Gax/ClientHelper.cs
```diff
@@ -100,15 +54,17 @@ internal CallOptions BuildCallOptions(CallSettings callSettingsOverride)
...
-                asyncGrpcCall(request, BuildCallOptions(callSettings)).ResponseAsync;
-            ApiCall<TRequest, TResponse>.SyncCall syncCall = (request, callSettings) =>
-                syncGrpcCall(request, BuildCallOptions(callSettings));
-            return new ApiCall<TRequest, TResponse>(asyncCall, syncCall);
+            CallSettings baseCallSettings =
+                (perMethodCallSettings?.Clone() ?? new CallSettings())
+                .Merge(_clientCallSettings);
+            return ApiCall.Create(asyncGrpcCall, syncGrpcCall, baseCallSettings, _clock)
+                .WithRetry(_clock, SystemScheduler.Instance)
+                .WithUserAgent(_userAgent);
```
jskeet: Ah, and there it is :)

src/Google.Api.Gax/Expiration.cs
```diff
@@ -9,6 +9,13 @@
 
 namespace Google.Api.Gax
 {
+    public enum ExpirationType
```
jskeet: Docs for all these public members :)

src/Google.Api.Gax/RetrySettings.cs
```diff
@@ -61,8 +61,8 @@ public sealed class RetrySettings
             {
                 RetryBackoff = RetryBackoff?.Clone(),
                 TimeoutBackoff = TimeoutBackoff?.Clone(),
+                RetryFilter = RetryFilter,
```
jskeet: Any reason for this change? Only one in the file, so probably worth reverting for a simpler commit.

test/Google.Api.Gax.Tests/CallSettingsTest.cs
```diff
@@ -56,5 +58,60 @@ public void CloneNullHeaders()
...
+            CallSettings callSettings = new CallSettings();
+            var options = callSettings.ToCallOptions(mockClock.Object);
+            Assert.Null(options.Deadline);
+            mockClock.Verify(c => c.GetCurrentDateTimeUtc(), Times.Never);
+        }
+
+        [Fact]
+        public void ToCallOptions_ExpirationTimeout()
+        {
+            var now = new DateTime(2015, 6, 19, 5, 2, 3, DateTimeKind.Utc);
```
jskeet: FakeClock is designed to make this sort of thing easier, unless you *really* care how often the clock is being consulted. (I don't, personally...)

test/Google.Api.Gax.Tests/ClientHelperTest.cs
```diff
@@ -15,104 +15,8 @@
 
 namespace Google.Api.Gax.Tests
 {
-    public class ClientHelperTest
+    /*public class ClientHelperTest
```
jskeet: ??? Delete the file if it's no longer useful?

src/Google.Api.Gax/ApiCallRetryExtensions.cs
```diff
@@ -7,89 +7,87 @@
...
-            where TRequest : class, IMessage<TRequest>
-            where TResponse : class, IMessage<TResponse>
-        {
-            GaxPreconditions.CheckNotNull(originalCall, nameof(originalCall));
-            GaxPreconditions.CheckNotNull(clock, nameof(clock));
+        // By design, the code is mostly duplicated between the async and sync calls.
 
-            if (retrySettings == null)
+        // Async retry
+        public static Func<TRequest, CallSettings, Task<TResponse>> WithRetry<TRequest, TResponse>(
```
jskeet: It seems very odd to me that WithRetry doesn't take any retry settings. Do we definitely need to have this as a separate concept? Given that it's baked into CallSettings, maybe it should be baked into ApiCall as well? I realize there's a *sort* of orthogonality here... but it feels a bit fake somehow.

jskeet:
Various nits and some reservations about where a few things live, but I'd be quite happy for it to go in after the nits are addressed, then maybe have another round of refactoring. (I'd be happy to tinker and see what I can do, and possibly admit defeat :)

src/Google.Api.Gax/ApiCall.cs
chrisdunelm: Acknowledged

src/Google.Api.Gax/ApiCallExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallExtensions.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: At the moment I like the separation of concerns that this provides.
I'd quite like to leave it as it is for now, and refactor later if it turns out to not be suitable.

jskeet: Sure. One thing to help that later on - can we make most of this internal? That way we can refactor after 1.0 more easily.

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Could do, although it'll make the code in ApiCall very slightly less attractive (in my opinion, that is).
I like the current "_asyncCall.WithRetry(clock, scheduler)" that this allows.
Putting it in ApiCall would require "WithRetry(_asyncCall, clock, scheduler)"
A small difference I know.
I also quite like it being in a separate file, as these retry methods are entirely separate from the other logic in ApiCall, although this could be done as a partial class.

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: Good point, yes, later we should check to see if allowing null and making it a static method could give less cloning.

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: Done

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: Yes, we did discuss this.
However (aside from the question of consistency with other merging properties), that would mean it's impossible to remove headers during an override, which I feel uncomfortable about.
I'll put a TODO in, and we should probably discuss this further and/or see what other code does in similar situations.

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: I was indeed avoiding the pointless assignment ;)
I'll change it, I agree your way is much easier on the eyes :)

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: This is done after all header overwriting has already been performed.
When a call happens, all the CallSettings are merged first, then the code goes into the AddUserAgent and Retry operations.

src/Google.Api.Gax/ClientHelper.cs
chrisdunelm: :)

src/Google.Api.Gax/Expiration.cs
chrisdunelm: Done

src/Google.Api.Gax/RetrySettings.cs
chrisdunelm: Done

test/Google.Api.Gax.Tests/ClientHelperTest.cs
chrisdunelm: Oops, done :)

test/Google.Api.Gax.Tests/CallSettingsTest.cs
chrisdunelm: Done

src/Google.Api.Gax/ApiCallRetryExtensions.cs
chrisdunelm: Done.

jskeet:
LGTM.

PR #34, opened 2016-05-04
Integration tests for GAX

jskeet:
Initial commit includes ChannelPool tests.

These should *not* be run on Travis, as they will require application default credentials.


chrisdunelm:
LGTM

PR #37, opened 2016-05-17
First pass at .NET Core RC2

jskeet:


.travis.yml
```diff
@@ -1,14 +1,30 @@
...
+
+addons:
+  apt:
+    packages:
+    - gettext
+    - libcurl4-openssl-dev
+    - libicu-dev
+    - libssl-dev
+    - libunwind8
+
```
ivannaranjo: Looking at the core-deps Docker image it seems that they might need more packages than this, https://github.com/dotnet/dotnet-docker/blob/master/1.0.0-rc2/core-deps/Dockerfile

jskeet: That was all I needed to get as far as the dotnet tool running, and successfully restoring and building packages. It's only the test that fails.

ivannaranjo: Not that this is the cause of the test failure, but I will assume that those dependencies are required by some parts of the runtime one way or another, so eventually they should be added to the setup.

jskeet: Well they may well be required by ASP.NET but not by just the core. I suspect all of this will change over time, to be honest.

jskeet:
Closing as we have a better PR now.

PR #38, opened 2016-05-19
Update GAX to protobuf 3.0.0-beta3 and gRPC 0.14.0

jskeet:
Will push this through myget, then we'll need to regenerate the source in gcloud-dotnet and target the new build.


src/Google.Api.Gax/project.json
```diff
@@ -12,9 +12,10 @@
   },
   "dependencies": {
     "Google.Apis.Auth": "1.11.0",
-    "Google.Protobuf": "3.0.0-beta2",
-    "Grpc.Auth": "0.13.1",
-    "Grpc.Core": "0.13.1",
+    "Google.Protobuf": "3.0.0-beta3",
+    "Google.Protobuf.Tools": { "version": "3.0.0-beta3", "type": "build" },
```
chrisdunelm: Why is this required?

jskeet: As per the commit comment - it's not required, but it's a very simple way of ensuring that protoc is readily available. It doesn't affect the eventual output. (It won't be a dependency.)

chrisdunelm: Oops, sorry, I should read the commit notes!

chrisdunelm:
LGTM

PR #40, opened 2016-05-24
Retry again

chrisdunelm:
"Retry" and "Simple expiration" of RPCs made clearer, with a new CallTimings settings class.

### DIFF
--- a/src/Google.Api.Gax/CallSettings.cs
+++ b/src/Google.Api.Gax/CallSettings.cs
@@ -22,12 +22,13 @@ namespace Google.Api.Gax
             if (other != null)
             {
                 Headers = other.Headers?.Clone();
-                Expiration = other.Expiration;
+                //Expiration = other.Expiration;
                 CancellationToken = other.CancellationToken;
                 WriteOptions = other.WriteOptions;
                 PropagationToken = other.PropagationToken;
                 Credentials = other.Credentials;
-                RetrySettings = other.RetrySettings?.Clone();
+                //RetrySettings = other.RetrySettings?.Clone();
+                CallTiming = other.CallTiming?.Clone();
             }
         }
         /// <summary>
@@ -35,11 +36,6 @@ namespace Google.Api.Gax
         /// </summary>
         //public Metadata Headers
         public Metadata Headers { get; set; }
-
-        /// <summary>
-        /// Call expiration.
-        /// </summary>
-        public Expiration Expiration { get; set; }
 
         /// <summary>
         /// Cancellation token that can be used for cancelling the call.
@@ -61,15 +57,7 @@ namespace Google.Api.Gax
         /// </summary>
         public CallCredentials Credentials { get; set; }
 
-        private RetrySettings _retrySettings;
-        /// <summary>
-        /// <see cref="RetrySettings"/> to use, or null for no retry.
-        /// </summary>
-        public RetrySettings RetrySettings
-        {
-            get { return _retrySettings; }
-            set { _retrySettings = value?.Validate(nameof(value)); }
-        }
+        public CallTiming CallTiming { get; set; }
 
         /// <summary>
         /// Creates a clone of this object, with all the same property values.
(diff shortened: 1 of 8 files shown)


src/Google.Api.Gax/CallSettings.cs
```diff
@@ -98,12 +90,13 @@ internal CallSettings Merge(CallSettings other)
             // Should a merge of Headers be additive, instead of overridding?
             // If additive, how to remove headers during an override?
             Headers = other.Headers ?? Headers;
-            Expiration = other.Expiration ?? Expiration;
+            //Expiration = other.Expiration ?? Expiration;
```
jskeet: Delete?

src/Google.Api.Gax/CallSettings.cs
```diff
@@ -98,12 +90,13 @@ internal CallSettings Merge(CallSettings other)
...
             CancellationToken = other.CancellationToken ?? CancellationToken;
             WriteOptions = other.WriteOptions ?? WriteOptions;
             PropagationToken = other.PropagationToken ?? PropagationToken;
             Credentials = other.Credentials ?? Credentials;
-            RetrySettings = other.RetrySettings ?? RetrySettings;
+            //RetrySettings = other.RetrySettings ?? RetrySettings;
```
jskeet: Delete?

src/Google.Api.Gax/CallTiming.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    public enum CallTimingType
```
jskeet: Docs :)

src/Google.Api.Gax/CallTiming.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        Retry,
+
+        /// <summary>
+        /// Call does not use retry; call expiration is specified with an <see cref="Google.Api.Gax.Expiration"/> .
+        /// </summary>
+        Expiration,
+    }
+
+    public sealed class CallTiming
```
jskeet: And more docs...

test/Google.Api.Gax.Tests/RetryTest.cs
```diff
@@ -302,7 +302,7 @@ public void AssertCallTimes(params long[] ticks)
             public void AssertDeadlines(params long[] ticks)
             {
                 // Note that this effectively asserts we always end up with a CallSettings with a deadline.
-                Assert.Equal(ticks, CallSettingsReceived.Select(cs => cs.Expiration.Deadline.Value.Ticks).ToArray());
+                Assert.Equal(ticks, CallSettingsReceived.Select(cs => cs.Timing.Expiration.Deadline.Value.Ticks).ToArray());
```
jskeet: Law of Demeter? We don't need no stinkin' Law of Demeter.

jskeet:
Yup, this looks clearer. Some XML docs to add, but then I'm happy.

src/Google.Api.Gax/CallTiming.cs
garrettjonesgoogle: The name "CallTiming" implies to me that it gives control around when the call is made. I have been thinking about what to name the combined concept of retry + timeout settings, and my current thought is "failure handling settings". None of these settings should cause anything to happen if an RPC succeeds and returns in the expected timeframe.

jskeet: But "the expected timeframe" is part of these settings - so it's not like these settings are *only* used when there's a failure.

garrettjonesgoogle: True... they also expand what is considered a failure, beyond the failures explicitly returned by the service (meaning the timeout case). 

The thing about the name "CallTiming" is that it doesn't capture the fact that failures are being handled at all. 

I've tried to find an overarching term that captures retry and timeout, but I don't know that there is a satisfying one. Maybe it just needs to be called "TimeoutAndRetrySettings." (I'm looking at this from the cross-language perspective; I think that the current Java model needs to be readjusted.)

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: Done

src/Google.Api.Gax/CallSettings.cs
chrisdunelm: Done

src/Google.Api.Gax/CallTiming.cs
chrisdunelm: Done

test/Google.Api.Gax.Tests/RetryTest.cs
chrisdunelm: :)

src/Google.Api.Gax/CallTiming.cs
chrisdunelm: Docs added.
Will leave it named CallTiming for now. Can reconsider the name later as required.

PR #47, opened 2016-06-15
Page streaming update - option 1

jskeet:
(At least one more option to come...)


jskeet:
We've decided against this, in favour of #48.

PR #53, opened 2016-06-28
Update GAX to gRPC 0.15.

jskeet:
This will help fix a few of the issues logged in gcloud-dotnet, but will require all the services to be regenerated against gRPC 0.15 as well.

Note that I *haven't* updated the Ix-Async dependency to System.Interactive.Async (as I'd have liked to) because gRPC still depends on Ix-Async. I'm tracking this at https://github.com/grpc/grpc/issues/7103 for gRPC.


chrisdunelm:
LGTM

PR #54, opened 2016-07-04
Another pass at dotnet cli

jskeet:


.travis.yml
```diff
@@ -1,14 +1,19 @@
 language: csharp
 
+dist: trusty
```
chrisdunelm: According to: https://docs.travis-ci.com/user/trusty-ci-environment/
You're also meant to have "sudo: required".

myget.bat
```diff
@@ -1,31 +1,24 @@
...
+  dotnet build --configuration %Configuration% %%P || goto error
 )
 
 for %%P in (%TESTS%) do (
   echo Testing %%P
-  call dnu build --configuration %Configuration% %%P
-  IF ERRORLEVEL 1 goto error
-  dnx --configuration %Configuration% -p %%P test
-  IF ERRORLEVEL 1 goto error
+  dotnet build --configuration %Configuration% %%P || goto error
```
chrisdunelm: This isn't strictly required, as "test" will also build.

.travis.yml
jskeet: Have never needed that... it does say "All Trusty images have sudo enabled" so maybe that explains why it's working? I'm happy to modify it if you'd like, but it does seem to be okay...

myget.bat
chrisdunelm: (This is a nit, I'm happy either way...)

jskeet: True. I previously wanted to separate them to make it clear which bit had failed, but I'm happy to embrace a smaller build script.

.travis.yml
chrisdunelm: It's fine, leave it as it is

chrisdunelm:
LGTM :)

myget.bat
jskeet: Humbug - intended to remove the build step, but failed on the commit amend. Will leave it for now.

PR #55, opened 2016-07-07
Separate out the common protos from GAX

### LINKED ISSUE #39: Separate "API proto types" from GAX?
Currently we include the generated code for google.api.* etc in Google.Gax.
We should consider putting it in a separate package instead. More dependencies, admittedly... but it would make for more sensible versioning.

jskeet:
Fixes #39.


chrisdunelm:
LGTM

PR #56, opened 2016-07-11
Update package and dependency versions

jskeet:
This is not for 1.0.0 GA, but initially 1.0.0-beta1.
Continuous integration will create 1.0.0-CI00070 builds etc.


chrisdunelm:
LGTM

PR #59, opened 2016-07-11
Update myget script

jskeet:
This now uses the configuration when building and testing, and makes it much easier to add new projects.


chrisdunelm:
Much nicer. LGTM.
Dumb question: Where does %Configuration% come from?

jskeet:
Myget provides %Configuration% itself, handily.
I'm running this through manually before I merge. Once this is working, I'll get the build service for the beta feed working, then move on to google-cloud-dotnet.

PR #64, opened 2016-07-29
Make ChannelPool apply scopes where appropriate

jskeet:
This is a PR which would be part of addressing
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/294
if it turns out the fix really should be in our libraries.

This is a breaking change - it requires support in the code generator as well, and we'd need to be careful not to release the GAX package separately, as anyone upgrading to that without upgrading the other packages would be broken. That *could* be addressed by adding a parameterless constructor here, but that feels a bit of a hack.

### DIFF
--- a/src/Google.Api.Gax/ChannelPool.cs
+++ b/src/Google.Api.Gax/ChannelPool.cs
@@ -22,20 +22,46 @@ namespace Google.Api.Gax
     /// </summary>
     public sealed class ChannelPool
     {
+        private readonly IEnumerable<string> _scopes;
+
         /// <summary>
         /// Lazily-created task to retrieve the default application channel credentials. Once completed, this
         /// task can be used whenever channel credentials are required. The returned task always runs in the
         /// thread pool, so its result can be used synchronously from synchronous methods without risk of deadlock.
-        /// The same channel credentials are used by all pools.
+        /// The same channel credentials are used by all pools. The field is initialized in the constructor, as it uses
+        /// _scopes, and you can't refer to an instance field within an instance field initializer.
         /// </summary>
-        private static readonly Lazy<Task<ChannelCredentials>> s_lazyDefaultChannelCredentials
-            = new Lazy<Task<ChannelCredentials>>(
-                () => Task.Run(async () => (await GoogleCredential.GetApplicationDefaultAsync()).ToChannelCredentials()));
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax/ChannelPool.cs
```diff
@@ -32,11 +32,50 @@ public sealed class ChannelPool
...
 
+        private readonly IEnumerable<string> _scopes;
+        private readonly Lazy<Task<ChannelCredentials>> _lazyScopedDefaultChannelCredentials;
+
         // TODO: See if we could use ConcurrentDictionary instead of locking. I suspect the issue would be making an atomic
         // "clear and fetch values" for shutdown.
         private readonly Dictionary<ServiceEndpoint, Channel> _channels = new Dictionary<ServiceEndpoint, Channel>();
         private readonly object _lock = new object();
 
+        // Constructor which exists for ease of testing as much as anything else.
```
chrisdunelm: This should go?

jskeet: The comment, or the constructor as well? I was thinking about this, and I wonder whether it would be better to remove it - very little test code is going to create a ChannelPool itself, and we can fix that up.

src/Google.Api.Gax/ChannelPool.cs
```diff
@@ -32,11 +32,50 @@ public sealed class ChannelPool
             = new Lazy<Task<ChannelCredentials>>(
                 () => Task.Run(async () => (await GoogleCredential.GetApplicationDefaultAsync()).ToChannelCredentials()));
 
+        private readonly IEnumerable<string> _scopes;
+        private readonly Lazy<Task<ChannelCredentials>> _lazyScopedDefaultChannelCredentials;
```
chrisdunelm: Nothing appears to use this.
Shouldn't GetChannel() be changed to use this as it's credential, rather than s_lazyDefaultChannelCredentials?
And can the non-scoped credentials be removed?

jskeet: Gah! Sorry, yes. Brain fail.

chrisdunelm:
LGTM

jskeet:
(LGTM received out-of-band.)

PR #69, opened 2016-08-31
Bring GAX build in line with google-cloud-dotnet

jskeet:
- Use bash for the build on Windows and Linux
- Make .travis.yml and myget.bat just trivial script runners
- Update the version in project.json to include beta tag


build.sh
```diff
@@ -0,0 +1,71 @@
...
+set -e
+
+if [ $(dotnet --info | grep "OS Platform" | grep -c Windows) -ne 0 ]
+then
+  OS=Windows
+else
+  OS=Linux
+fi
+
+FIND=/usr/bin/find
```
ivannaranjo: Nit: As a general note in style, keep the non-exported (local) variables lowercase.

jskeet: I suspect it'll be worth doing a second pass at these for style, yes - after initial merge.

build.sh
```diff
@@ -0,0 +1,71 @@
...
+# TODO: Work out all projects we can test with dotnet test
+# automatically
+
+dotnet test -f netcoreapp1.0 $DOTNET_TEST_ARGS test/Google.Api.Gax.Rest.Tests
+
+echo Packing
+
+# Assume each packagable project contains something like "version": "1.0.0-*"
+# and no other projects do.
+for package in `$FIND . -name project.json | xargs grep -le 'version.*-\*' | sed 's/\/project.json//g'`
```
ivannaranjo: Are you trying to remove the project.json at the end of the path? I believe all dotnet commands accept the path to either the directory or project.json.

ivannaranjo:
Minor nits. LGTM

build.sh
jskeet: Ooh, that's interesting - I didn't know that. Will give it a try later (as that change would impact non-GAX too).

PR #71, opened 2016-09-09
Include -* suffix on project versions

jskeet:
This is so we can pick up the CI release number etc.

This was the intention to start with, and it's what we already have in the other repo.


chrisdunelm:
LGTM

PR #72, opened 2016-09-09
Add a user agent helper to the REST version of GAX

jskeet:
This avoids us baking logic into multiple libraries
(Storage+Bigquery).

(After merging this, I'll promote the resulting myget build, and use it in the libraries...)


chrisdunelm:
LGTM

PR #73, opened 2016-09-12
Update version numbers ready for a beta02 release.

jskeet:
Dependencies are changed to 1.16 to avoid unnecessary binding redirects in some cases.

(Once this is in, I'll create and publish the release, then update google-cloud-dotnet libraries in a separate PR.)


chrisdunelm:
LGTM

PR #78, opened 2016-09-20
Enable easy faking of ApiCall<,>.

### LINKED ISSUE #77: Add a way of constructing an ApiCall with an async func
Currently `ApiCall<,>` is easy enough to construct in terms of the synchronous call, but the async call is tricky - as it needs to return an `AsyncUnaryResponse`. We only really need a `Task<TResponse>`, so we could expose (in Gax.Testing?) a call to create an `ApiCall<,>` from a `Func<TRequest, TResponse>` and a `Func<TRequest, Task<TResponse>>`.

jskeet:
This could make it reasonably easy to bypass gRPC if we wanted to, constructing a FooClientImpl with appropriate API calls but no gRPC client. Something for later consideration.

Fixes #77.

### DIFF
--- a/testing/Google.Api.Gax.Testing/FakeApiCall.cs
+++ b/testing/Google.Api.Gax.Testing/FakeApiCall.cs
@@ -1,0 +1,76 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Google.Protobuf;
+using Grpc.Core;
+using System;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Testing
+{
+    /// <summary>
+    /// Helper class to use when faking services; enables <see cref="ApiCall{TRequest, TResponse}"/> objects
+    /// to be created without gRPC 
+    /// </summary>
+    public static class FakeApiCall
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


testing/Google.Api.Gax.Testing/FakeApiCall.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="clock">The clock to use; defaults to a new <see cref="FakeClock"/>.</param>
+        /// <returns>The new API call.</returns>
+        public static ApiCall<TRequest, TResponse> Create<TRequest, TResponse>(
+            Func<TRequest, CallOptions, TResponse> syncCall,
+            CallSettings baseCallSettings = null,
+            IClock clock = null)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse> =>
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+            Create(async (request, response) => syncCall(request, response),
```
chrisdunelm: Might this be slightly clearer:
Create((request, response) => Task.FromResult(syncCall(request, response))
?

jskeet: That was my first thought - but that would throw exceptions synchronously. This way, if the syncCall throws, you end up with a faulted task, which is more realistic.

chrisdunelm: Good point, worth a comment?

jskeet: Okay, will do.

PR #88, opened 2016-09-30
Generate XML docs

### LINKED ISSUE #87: Include XML docs in packages
For some reason we don't have

```
"xmlDoc": true
```

in our project.json files. We should do so for all packages that will be shipped (so typically just the API package, but also any testing packages).

jskeet:
This just adds `"xmlDoc": true` to all the relevant project.json files, then fixes up the code so that everything is actually documented.

The only code change is to make `ApiCall` internal, given that a static class with no public members isn't really useful.

This fixes #87.

### DIFF
--- a/src/Google.Api.CommonProtos/project.json
+++ b/src/Google.Api.CommonProtos/project.json
@@ -12,7 +12,8 @@
   },
 
   "buildOptions": {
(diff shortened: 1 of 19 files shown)


jskeet:
Not sure why I didn't assign this before...

PR #89, opened 2016-10-10
Use gRPC 1.0.1-pre1 to support .NET Core

jskeet:


src/Google.Api.Gax/UserAgentBuilder.cs
```diff
@@ -43,19 +43,23 @@ internal UserAgentBuilder AppendVersion(string name, string version)
...
             => AppendVersion(name, FormatAssemblyVersion(type));
 
-        // TODO: This probably won't work as-is on .NET Core.
         /// <summary>
         /// Appends the .NET environment information to the list.
         /// </summary>
         internal UserAgentBuilder AppendDotNetEnvironment()
+#if NETSTANDARD1_5
+            // TODO: Improve this if we can.
+            => AppendVersion("dotnet", "unknown");
```
chrisdunelm: Mention "core" or "netstandard" until it can be improved?

jskeet: Not necessarily, as all it means is that we're using the .NET Core build - it could still be running on the regular framework, just picked up by a dependency from something that only targets netstandard.

PR #90, opened 2016-10-10
Moved PollSettings from Google.LongrunningOperations

jskeet:
Also moved the code to perform the repeated polling, so
that it's easy to use from other code.

PollSettings is now mutable, like all other settings classes - but
as normal, it should not be mutated when in use.

### DIFF
--- a/src/Google.Api.Gax/PollSettings.cs
+++ b/src/Google.Api.Gax/PollSettings.cs
@@ -13,11 +13,10 @@ namespace Google.Api.Gax
     /// Settings controlling repeated polling, for example when waiting for a long-running operation
     /// to complete.
     /// </summary>
-    public class PollSettings
+    public sealed class PollSettings
     {
         /// <summary>
-        /// Settings to override the <see cref="OperationsClient"/> call settings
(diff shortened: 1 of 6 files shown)


jskeet:
(Unit tests TBD - I basically need to extract them from the Operations code, but there they're attached to server code...)

src/Google.Api.Gax/PollSettings.cs
```diff
@@ -0,0 +1,64 @@
...
+    public class PollSettings
+    {
+        /// <summary>
+        /// Settings to override the <see cref="OperationsClient"/> call settings
+        /// when polling, if any.
+        /// </summary>
+        public CallSettings CallSettings { get; set; }
+
+        /// <summary>
+        /// How long to wait before giving up. This may be null, which is equivalent to <see cref="Expiration.None"/>
```
chrisdunelm: I do wonder if Expiration should be a struct

jskeet:
Now with tests :) Will merge when green.

chrisdunelm:
Great :)

PR #94, opened 2016-10-13
Split the gRPC-specific parts of GAX out from non-gRPC-specific

jskeet:
This should allow a lot more code reuse, e.g. for polling from the BigQuery library.

It makes things a bit more complex, admittedly, but it has user benefits in terms of there not being multiple preconditions classes, interfaces named the same thing etc.

If we go ahead with this, we'll need changes to codegen and all the libraries, of course.


src/Google.Api.Gax/project.json
```diff
@@ -17,16 +17,14 @@
...
-    "Grpc.Auth": "1.0.1-pre1",
-    "Grpc.Core": "1.0.1-pre1",
-    "Grpc.Tools": { "version": "1.0.0", "type": "build" },
     "System.Interactive.Async": "3.0.0"
   },
   "frameworks": {
-    "net45": { },
-    "netstandard1.5": { }
+    "net45": {},
+    "netstandard1.3": {
```
jskeet: Nope, we can target 1.3 for Google.Api.Gax, which means we can keep targeting 1.3 for Google.Storage.V1 and Google.Bigquery.V1.

But Google.Api.Gax.Grpc still targets 1.5, because that's all that gRPC supports.

PR #101, opened 2016-10-27
Alternative approach for IPagedEnumerable

jskeet:
This mostly affects WithFixedPageSize becoming GetCompletePage;
the rest of the public API change is a matter of removing code.


jskeet:
Feedback requested:

- Rename FixedSizePage?
- Rename GetCompletePage?

src/Google.Api.Gax/IPagedEnumerable.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <remarks>
+        /// "Natural" pages returned by the API may contain a smaller number of resources than requested.
+        /// For example, a request for a page with 100 resources may return a page with 80 resources but
+        /// a next page token for more to be retrieved. This is suitable for batch-processing, but not
+        /// for user-visible paging such as in a web application, where fixed-size pages are expected.
+        /// This method may make more than one API call in order to fill the page.
+        /// </remarks>
+        /// <param name="pageSize">The page size. Must be greater than 0.</param>
+        /// <returns>An asynchronous operation, the result of which is a page of resources.</returns>
+        FixedSizePage<TResource> GetCompletePage(int pageSize);
```
chrisdunelm: GetFirstPage(int pageSize) perhaps?

chrisdunelm: Wondering if we should also have:
IEnumerable&lt;FixedSizePage<TResource>> GetPages(int pageSize) ?
I know we're trying to simplify, and this is probably an uncommon use-case, but it feels sort of incomplete without it.

src/Google.Api.Gax/FixedSizePage.cs
```diff
@@ -19,7 +19,14 @@ public sealed class FixedSizePage<TResource> : IEnumerable<TResource>
     {
```
chrisdunelm: This could just be called "Page"?
Or "ResourcePage"?
Not sure.

jskeet: Yes, this is definitely something to think about. I'm not even sure we should be using Page for this, given that it's not necessarily a natural page from the API perspective.

src/Google.Api.Gax/IPagedEnumerable.cs
jskeet: Hmm. This is definitely an area where if we made this an abstract class instead of an interface, we'd be in a better position... we could add it later instead.

chrisdunelm: Yes, an abstract class may well be a better option.

src/Google.Api.Gax/FixedSizePage.cs
chrisdunelm: But the method that returns this is currently called GetCompletePage(). Having the two match seems like a good idea.

jskeet: Absolutely. I don't like either name at the moment :)

src/Google.Api.Gax.Grpc/PageStreamingAsync.cs
```diff
@@ -43,7 +44,12 @@ public sealed class PagedAsyncEnumerable<TRequest, TResponse, TResource> : IPage
         }
 
         /// <inheritdoc/>
-        public IResponseAsyncEnumerable<TResponse, TResource> AsPages() => _pages;
+        public IAsyncEnumerable<TResponse> AsResponses() => _pages;
+
+        /// <inheritdoc/>
+        public Task<FixedSizePage<TResource>> GetCompletePageAsync(
```
SurferJeffAtGoogle: I can't figure out what 'Complete' means in this context.

src/Google.Api.Gax.Grpc/PageStreamingAsync.cs
```diff
@@ -43,7 +44,12 @@ public sealed class PagedAsyncEnumerable<TRequest, TResponse, TResource> : IPage
         }
 
         /// <inheritdoc/>
-        public IResponseAsyncEnumerable<TResponse, TResource> AsPages() => _pages;
+        public IAsyncEnumerable<TResponse> AsResponses() => _pages;
```
SurferJeffAtGoogle: Why change the name?

src/Google.Api.Gax.Grpc/PageStreamingAsync.cs
```diff
@@ -71,16 +77,47 @@ internal sealed class ResponseAsyncEnumerable<TRequest, TResponse, TResource> :
             _apiCall = apiCall;
         }
 
-        public IResponseAsyncEnumerator<TResponse> GetEnumerator() =>
+        /// <inheritdoc />
+        public IAsyncEnumerator<TResponse> GetEnumerator() =>
             new ResponseAsyncEnumerator(_callSettings, _request.Clone(), _apiCall);
 
-        IAsyncEnumerator<TResponse> IAsyncEnumerable<TResponse>.GetEnumerator() => GetEnumerator();
+        internal async Task<FixedSizePage<TResource>> GetCompletePageAsync(
```
SurferJeffAtGoogle: Again, I don't understand what the word 'Complete' is trying to tell me.

src/Google.Api.Gax.Grpc/PageStreamingSync.cs
```diff
@@ -35,135 +37,63 @@ public sealed class PagedEnumerable<TRequest, TResponse, TResource> : IPagedEnum
...
-
-        /// <inheritdoc />
-        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
-        /// <inheritdoc />
-        IEnumerator<TResponse> IEnumerable<TResponse>.GetEnumerator() => GetEnumerator();
-
-        /// <inheritdoc />
-        public IEnumerable<FixedSizePage<TResource>> WithFixedSize(int pageSize)
+        /// <inheritdoc/>
+        public IEnumerable<TResponse> AsResponses()
```
SurferJeffAtGoogle: Does this make an API call?  I do not expect a method named As<Anything> to make an API call.

test/Google.Api.Gax.Grpc.Tests/PageStreamingTest.cs
```diff
@@ -157,7 +157,7 @@ public void NaturalPages(PagedResource pagedResource)
             var server = new FakeServer(pagedResource);
             var request = new PageStreamingRequest { PageSize = 0 };
             var paged = server.PagedSync(null, null, request);
-            Assert.Equal(pagedResource.Resource, paged.AsPages().Select(x => x.ToArray()).ToArray());
+            Assert.Equal(pagedResource.Resource, paged.AsResponses().Select(x => x.ToArray()).ToArray());
```
SurferJeffAtGoogle: Is there a semantic difference between AsPages() and AsResponses()?

src/Google.Api.Gax.Grpc/PageStreamingAsync.cs
jskeet: To clarify that they really are API responses - that's the point of them.

src/Google.Api.Gax.Grpc/PageStreamingSync.cs
jskeet: No it doesn't. It just means that when you start iterating over the result, you'll get back responses rather than the resources.

src/Google.Api.Gax.Grpc/PageStreamingAsync.cs
jskeet: It's complete in terms of "it'll give you everything you asked for if it possibly can". But I definitely want a better name :)

jskeet:
Now updated with proposed final names.

@SurferJeffAtGoogle We weren't sure if `ReadPage` should be `ReadOnePage`, `ReadSinglePage` etc. Thoughts welcome.

jskeet:
Removed obsolete comments - will merge when green.

PR #107, opened 2016-11-16
Add a ScopedCredentialProvider

jskeet:
This caches the scoped credentials for the default application
credentials, to avoid a new auth token being requested each time.

It would be nice to use this within ChannelPool, but because there's
the extra transformation of creating ChannelCredentials, and because
the Lazy<T> is of a task, it doesn't quite work out. Making it work
would probably require more new code than the removed redundancy...

Background: we want this for Storage and Bigquery so that creating new clients regularly doesn't have as much overhead. It will also make the code simpler in each of those projects.

### DIFF
--- a/src/Google.Api.Gax/ScopedCredentialProvider.cs
+++ b/src/Google.Api.Gax/ScopedCredentialProvider.cs
@@ -1,0 +1,95 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+using Google.Apis.Auth.OAuth2;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Simple factory of scoped credentials, which caches a scoped version of the
+    /// default application credentials to avoid repeated authentication.
+    /// </summary>
+    public sealed class ScopedCredentialProvider
+    {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Google.Api.Gax/ScopedCredentialProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                // No need to apply scopes here - they're already applied.
+                return _lazyScopedDefaultCredentials.Value.Result;
+            }
+            catch (AggregateException e)
+            {
+                // Unwrap the first exception, a bit like await would.
+                // It's very unlikely that we'd ever see an AggregateException without an inner exceptions,
+                // but let's handle it relatively gracefully.
+                throw e.InnerExceptions.FirstOrDefault() ?? e;
```
chrisdunelm: nit: perhaps:
throw e.InnerExceptions.Count == 1 ? e.InnerExceptions[0] : e;

src/Google.Api.Gax/project.json
```diff
@@ -17,6 +17,7 @@
   },
 
   "dependencies": {
+    "Google.Apis.Auth": "1.18.0",
```
chrisdunelm: I take it we're happy taking a dependency on Auth here in Gax?

src/Google.Api.Gax/ScopedCredentialProvider.cs
jskeet: That wouldn't match what await does. (This is the code we have everywhere else, btw.)

src/Google.Api.Gax/project.json
jskeet: I think so - it's common between REST and gRPC.

jskeet:
Merging despite Travis failure - that's a known issue: https://github.com/travis-ci/travis-ci/issues/6881

PR #108, opened 2016-11-22
Update System.Interactive.Async to 3.1.1

jskeet:
This avoids taking a dependency on NETStandard.Library when
using net45.


jskeet:
Basically, we don't want a net45 user to end up pulling in NetStandard.Library unnecessarily, which they would with 3.1.0. This was fixed in https://github.com/Reactive-Extensions/Rx.NET/commit/cb5686b25248e6be4976928214be7d09c0dc703f

PR #111, opened 2016-11-25
Add cancellation to IScheduler

### LINKED ISSUE #84: IScheduler.Delay should take a cancellation token
(And we should use that in retry.)

jskeet:
At the same time, the synchronous Sleep method has been removed from
IScheduler, and implemented as an extension method (with a
cancellation token).

Fixes #84.

### DIFF
--- a/src/Google.Api.Gax.Grpc/ApiCallRetryExtensions.cs
+++ b/src/Google.Api.Gax.Grpc/ApiCallRetryExtensions.cs
@@ -47,7 +47,7 @@ namespace Google.Api.Gax.Grpc
                         {
                             throw;
                         }
-                        await scheduler.Delay(actualDelay);
+                        await scheduler.Delay(actualDelay, callSettings.CancellationToken.GetValueOrDefault());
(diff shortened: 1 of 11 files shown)


src/Google.Api.Gax.Grpc/ApiCallRetryExtensions.cs
```diff
@@ -86,7 +86,7 @@ internal static class ApiCallRetryExtensions
                         {
                             throw;
                         }
-                        scheduler.Sleep(actualDelay);
+                        scheduler.Delay(actualDelay, callSettings.CancellationToken.GetValueOrDefault()).Wait();
```
chrisdunelm: schduler.Sleep() extension method?

jskeet: Doh - thought I'd got all of them, but no... because I wasn't calling WaitAndUnwrapExceptions, which I should have been :) Will fix.

PR #113, opened 2016-11-25
Remove flaky cancellation tests and warn about FakeScheduler

jskeet:
It's just not right yet.

### DIFF
--- a/test/Google.Api.Gax.Tests/FakeSchedulerTest.cs
+++ b/test/Google.Api.Gax.Tests/FakeSchedulerTest.cs
@@ -15,26 +15,15 @@ namespace Google.Api.Gax.Tests
 {
     public class FakeSchedulerTest
     {
+        // Note: no tests for cancellation at the moment as they're fiendishly difficult to write in a robust way.
+        // We need to revisit FakeScheduler yet again...
+
         [Fact]
         public async Task SimpleDelay()
         {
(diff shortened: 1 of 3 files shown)


test/Google.Api.Gax.Tests/FakeSchedulerTest.cs
```diff
@@ -15,26 +15,15 @@ namespace Google.Api.Gax.Tests
...
         [Fact]
         public async Task SimpleDelay()
         {
             var scheduler = new FakeScheduler();
             await scheduler.RunAsync(() => scheduler.Delay(TimeSpan.FromTicks(1000)));
             Assert.Equal(1000, scheduler.Clock.GetCurrentDateTimeUtc().Ticks);
         }
-
-        [Fact]
-        public async Task Cancellation()
```
chrisdunelm: Perhaps keep the test, but mark it Ignored for the moment?

jskeet: Would rather not, because the API for scheduling the cancellation is likely to change too. (My ScheduleCancellation didn't do exactly what I wanted.)

chrisdunelm: OK, LGTM then.

PR #117, opened 2017-01-16
Buffering writer for gRPC streams

jskeet:
gRPC's native streams only allow a single message to be written to the stream at a time. This can
be very inconvenient for clients (although it's necessary in other cases). This class can buffer messages
to be written (and the completion signal), while still allowing clients to see when individual writes have succeeded.

// cc @jtattermusch

### DIFF
--- a/src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
+++ b/src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
@@ -1,0 +1,286 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc
+{
+    // TODO: Check whether cancellation token associated with original RPC is honoured throughout
+    // TODO: Check whether CallOptions are honoured throughout
+    // TODO: Check that the "further calls to Write after failure" handling is appropriate. (And document it!)
+
+    /// <summary>
+    /// A wrapper around <see cref="IClientStreamWriter{T}"/> which removes the "one write at a time"
+    /// restriction by buffering messages (and the completion signal) up to a given capacity.
+    /// </summary>
+    /// <typeparam name="T">The type of message in the stream.</typeparam>
+    public class BufferedClientStreamWriter<T> where T : class
+    {
+        private readonly IClientStreamWriter<T> _writer;
+        private readonly object _lock = new object();
+        /// <summary>
+        /// Queue of requests. If this is non-empty, there's at least one request in-flight, which
+        /// is always the head of the queue.
+        /// </summary>
+        private readonly LinkedList<Request> _queue;
+
+        // TODO: Make these internal properties public?
+
+        /// <summary>
+        /// The capacity of the writer.
+        /// </summary>
+        internal int Capacity { get; }
+
+        /// <summary>
+        /// The number of write calls that have been buffered.
+        /// </summary>
+        /// <remarks>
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc
+{
+    // TODO: Check whether cancellation token associated with original RPC is honoured throughout
+    // TODO: Check whether CallOptions are honoured throughout
+    // TODO: Check that the "further calls to Write after failure" handling is appropriate. (And document it!)
```
chrisdunelm: What happens if one of the tasks returned by WriteAsync (et al) is cancelled?
It looks like that's ignored at the moment - is this the desired behaviour?

chrisdunelm: Err, ignore this, daft question due to me temporarily forgetting how cancellation works...

src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
...
+                    return null;
+                }
+                // Not part of state validation, as it's slightly different.
+                if (_failedTask != null)
+                {
+                    return _failedTask;
+                }
+                var lastRequest = _queue.Last?.Value;
+                var effectiveOptions = modifyOptions ? options
+                    : lastRequest != null ? lastRequest.WriteOptions
```
jskeet: Nope - because if `lastRequest?.WriteOptions` is null, that's what we want to propagate. null is a valid value.

src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
...
+                else
+                {
+                    return false;
+                }
+            }
+            if (_queue.Count == Capacity)
+            {
+                if (throwOnError)
+                {
+                    throw new InvalidOperationException($"Queue is full");
```
jskeet: Done.

src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// </summary>
+        /// <exception cref="InvalidOperationException">This method has already been called.</exception>
+        /// <returns>A <see cref="Task"/> which will complete when the stream has finished being completed.</returns>
+        public Task CompleteAsync()
+        {
+            lock (_lock)
+            {
+                if (_completed)
+                {
+                    throw new InvalidOperationException($"Cannot call {nameof(CompleteAsync)} twice");
```
jskeet: I don't think so - I think I'd typically just put the name of the method.

src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
```
jskeet: Done.

src/Google.Api.Gax.Grpc/BufferedClientWriter.cs
```diff
@@ -0,0 +1,285 @@
...
+
+        private static void PropagateTaskResult(Task task, TaskCompletionSource<int> completionSource)
+        {
+            switch (task.Status)
+            {
+                case TaskStatus.Canceled:
+                    completionSource.TrySetCanceled();
+                    break;
+                case TaskStatus.Faulted:
+                    completionSource.SetException(task.Exception.InnerExceptions);
```
jskeet: Yes, it can be SetCanceled, which is better in terms of throwing if we've got a bug. Done.

PR #119, opened 2017-01-17
GAX additions to support bidirectional streaming.

chrisdunelm:


src/Google.Api.Gax.Grpc/BidirectionalStreamingBase.cs
```diff
@@ -0,0 +1,108 @@
...
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Completes the stream when all buffered messages have been sent. This method can only be called
+        /// once, and further messages cannot be written after it has been called.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">This method has already been called.</exception>
+        /// <returns>A <see cref="Task"/> which will complete when the stream has finished being completed.</returns>
+        public virtual Task WriteCompleteAsync()
```
jskeet: Not a full review, but two points before I forget:

- I suspect we also want TryWriteCompleteAsync, having now used this. (Could manage without, but can be handy sometimes.)
- We should rename CompleteAsync to WriteCompleteAsync in the BufferedClientStreamWriter.

chrisdunelm: Done

src/Google.Api.Gax.Grpc/BidirectionalStreamingSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// </summary>
+        /// <param name="bufferCapacity">The capacity of the write buffer.</param>
+        /// <returns>A <see cref="BidirectionalStreamingSettings"/> configured with a write buffer capacity.</returns>
+        public static BidirectionalStreamingSettings FromBufferCapacity(int bufferCapacity) =>
+            new BidirectionalStreamingSettings(bufferCapacity);
+
+        /// <summary>
+        /// Configure settings for bidirection streaming.
+        /// </summary>
+        /// <param name="bufferedClientWriterCapacity">The capacity of the write buffer.</param>
```
jskeet: Do we need a public constructor and a factory method?

(bidirection => bidirectional too, btw)

src/Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -69,5 +70,27 @@ public ClientHelper(ServiceSettingsBase settings)
...
+        /// <param name="streamingSettings">The default streaming settings.</param>
+        /// <returns>An API call to proxy to the RPC calls</returns>
+        public ApiBidirectionalStreamingCall<TRequest, TResponse> BuildApiCall<TRequest, TResponse>(
+            Func<CallOptions, AsyncDuplexStreamingCall<TRequest, TResponse>> grpcCall,
+            CallSettings perMethodCallSettings,
+            BidirectionalStreamingSettings streamingSettings)
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            // TODO: What about retry?
```
jskeet: Looks like retry is basically not an option - so maybe we should check that there *is* no retry?

src/Google.Api.Gax.Grpc/BidirectionalStreamingBase.cs
```diff
@@ -88,9 +88,9 @@ public virtual Task WriteAsync(TRequest message, WriteOptions options)
         /// Completes the stream when all buffered messages have been sent. This method can only be called
         /// once, and further messages cannot be written after it has been called.
```
jskeet: TryWriteComplete can be called multiple times, but only the first time will have any effect (at the moment - that will change when we put it in the buffer, of course).

src/Google.Api.Gax.Grpc/BidirectionalStreamingBase.cs
```diff
@@ -85,8 +85,9 @@ public virtual Task WriteAsync(TRequest message, WriteOptions options)
         }
 
         /// <summary>
-        /// Completes the stream when all buffered messages have been sent. This method can only be called
-        /// once, and further messages cannot be written after it has been called.
+        /// Completes the stream when all buffered messages have been sent.
+        /// Only the first call to this method on any instance will have any affect;
```
jskeet: Nit: affect => effect :)

PR #127, opened 2017-01-27
Method to create a new set of settings with an earlier deadline.

jskeet:
This will be used from polling code which needs to take into account
an overall deadline from PollSettings along with an existing
CallSettings.

Currently no tests - I'll amend this commit to add them when we've settled on the design. I want at least one test hitting each case :)

### DIFF
--- a/src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
+++ b/src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
@@ -5,6 +5,7 @@
  * https://developers.google.com/open-source/licenses/bsd
  */
 using Grpc.Core;
+using System;
 using System.Threading;
 
 namespace Google.Api.Gax.Grpc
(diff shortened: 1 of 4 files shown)


src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -96,6 +97,55 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
...
+            GaxPreconditions.CheckNotNull(clock, nameof(clock));
+            // No deadline? Return what we already have.
+            if (deadline == null)
+            {
+                return settings;
+            }
+            // No settings, or no timing in the settings? Create a simple expiration.
+            if (settings == null || settings.Timing == null)
+            {
+                return settings.WithCallTiming(CallTiming.FromDeadline(deadline.Value));
```
chrisdunelm: settings may be null here

jskeet: Yup, and that's fine - WithCallTiming is an extension method, and it handles the first arg being null :)

chrisdunelm: Can you add a comment to that effect? This *looks* broken ;)

jskeet:
Okay, we now have tests. The only tests I know I haven't explicitly written are for a retry with an overall timeout rather than deadline - but I don't think that would really be proving anything else, and I've already got 3x as much test code as production :)

test/Google.Api.Gax.Grpc.Tests/CallSettingsExtensionsTest.cs
```diff
@@ -158,5 +159,128 @@ public void WithHeader_NonNullSettings()
...
+            CallSettings settings = null;
+            DateTime? deadline = null;
+
+            Assert.Null(settings.WithEarlierDeadline(deadline, new FakeClock()));
+        }
+
+        [Fact]
+        public void WithEarlierDeadline_SimpleSettingsNoDeadline()
+        {
+            CallSettings settings = CallSettings.FromHeader("name", "value"); ;
```
chrisdunelm: Superfluous semicolon.

jskeet: Doh, thanks - will fix then merge.

PR #128, opened 2017-01-27
GAX platform detection

chrisdunelm:
Detection and details of GAE and GCE.

### DIFF
--- a/src/Google.Api.Gax/Platform.cs
+++ b/src/Google.Api.Gax/Platform.cs
@@ -1,4 +1,11 @@
-﻿using System;
+﻿/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -13,6 +20,7 @@ namespace Google.Api.Gax
         Unknown = 0,
         Gce = 1,
         Gae = 2,
+        // TODO: Add GKE if/when possible.
     }
 
     public sealed class GcePlatformDetails
@@ -34,14 +42,17 @@ namespace Google.Api.Gax
 
     public sealed class GaePlatformDetails
     {
-        public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeVersion)
+        public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeService, string gaeVersion)
         {
             ProjectId = gcloudProject;
-
+            InstanceId = gaeInstance;
+            ServiceId = gaeService;
+            VersionId = gaeVersion;
         }
 
         public string ProjectId { get; }
-        public string ModuleId { get; }
+        public string InstanceId { get; }
+        public string ServiceId { get; }
         public string VersionId { get; }
     }
 
@@ -67,31 +78,37 @@ namespace Google.Api.Gax
 
         private static async Task<GcePlatformDetails> LoadGceDetails()
         {
+            // TODO: Respect METADATA_EMULATOR_HOST environment variable. Where is this documented?
             const string metadataUrl = "http://metadata.google.internal/computeMetadata/v1?recursive=true";
             const string metadataFlavorKey = "Metadata-Flavor";
             const string metadataFlavorValue = "Google";
             const long maxContentLength = 512 * 1024; // Maximum allowed metadata size.
-            var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataUrl);
-            httpRequest.Headers.Add(metadataFlavorKey, metadataFlavorValue); // Required for any query.
-            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(1000)); // 1000 found to be reasonable.
-            var httpClient = new HttpClient();
-            var response = await httpClient.SendAsync(httpRequest, cts.Token);
-            IEnumerable<string> metadataValues;
-            if (response.StatusCode == HttpStatusCode.OK
-                && response.Content.Headers.TryGetValues(metadataFlavorKey, out metadataValues)
-                && metadataValues.Contains(metadataFlavorValue)
-                && response.Content.Headers.ContentLength < maxContentLength)
+            try
             {
-                // Valid response from metadata server.
-                string metadataJson = await response.Content.ReadAsStringAsync();
-                try
+                var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataUrl);
+                httpRequest.Headers.Add(metadataFlavorKey, metadataFlavorValue); // Required for any query.
+                var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(1000)); // 1000 found to be reasonable.
+                var httpClient = new HttpClient();
+                var response = await httpClient.SendAsync(httpRequest, cts.Token); // TODO: Consider retrying on IO Exception.
+                IEnumerable<string> metadataValues;
+                if (response.StatusCode == HttpStatusCode.OK
+                    && response.Content.Headers.TryGetValues(metadataFlavorKey, out metadataValues)
+                    && metadataValues.Contains(metadataFlavorValue)
+                    && response.Content.Headers.ContentLength < maxContentLength)
                 {
+                    // Valid response from metadata server.
+                    string metadataJson = await response.Content.ReadAsStringAsync();
                     return new GcePlatformDetails(metadataJson);
                 }
-                catch
-                {
-                    // TODO: How to log problems?
-                }
+            }
+            catch
+            {
+                // Possible exceptions, in all cases there's nothing we can do except
+                // assume we're not running on GCE, and hence return null.
+                // OperationCanceledException: on timeout
+                // HttpRequestException: On general request failure
+                // WebException: DNS problem on Mono
+                // TODO: Decide what to do here.
             }
             return null;
         }
@@ -102,10 +119,11 @@ namespace Google.Api.Gax
             // for details on environment variables.
             var gcloudProject = Environment.GetEnvironmentVariable("GCLOUD_PROJECT");
             var gaeInstance = Environment.GetEnvironmentVariable("GAE_INSTANCE");
+            var gaeService = Environment.GetEnvironmentVariable("GAE_SERVICE");
             var gaeVersion = Environment.GetEnvironmentVariable("GAE_VERSION");
             if (gcloudProject != null && gaeInstance != null && gaeVersion != null)
             {
-                return new GaePlatformDetails(gcloudProject, gaeInstance, gaeVersion);
+                return new GaePlatformDetails(gcloudProject, gaeInstance, gaeService, gaeVersion);
             }
             return null;
         }
@@ -145,6 +163,5 @@ namespace Google.Api.Gax
             GaeDetails != null ? PlatformType.Gae :
             GceDetails != null ? PlatformType.Gce :
             PlatformType.Unknown;
-
     }
 }
--- a/src/Google.Api.Gax/Platform.cs
+++ b/src/Google.Api.Gax/Platform.cs
@@ -1,0 +1,150 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.Threading;
+using System.Net;
+
+namespace Google.Api.Gax
+{
+    public enum PlatformType
+    {
+        Unknown = 0,
+        Gce = 1,
+        Gae = 2,
+    }
+
+    public sealed class GcePlatformDetails
+    {
+        public GcePlatformDetails(string metadataJson)
+        {
+            MetadataJson = metadataJson;
+            var json = Newtonsoft.Json.Linq.JObject.Parse(metadataJson);
+            ProjectId = json["project"]["projectId"].ToString();
+            InstanceId = json["instance"]["id"].ToString();
+            ZoneName = json["instance"]["zone"].ToString();
+        }
+
+        public string MetadataJson { get; }
+        public string ProjectId { get; }
+        public string InstanceId { get; }
+        public string ZoneName { get; }
+    }
+
+    public sealed class GaePlatformDetails
+    {
+        public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeVersion)
+        {
+            ProjectId = gcloudProject;
+
+        }
+
+        public string ProjectId { get; }
+        public string ModuleId { get; }
+        public string VersionId { get; }
+    }
+
+    /// <summary>
+    /// Information about the current execution platform.
+    /// Supported execption platforms are Google App Engine (GAE) and Google Compute Engine (GCE).
+    /// </summary>
+    public sealed class Platform
+    {
+        private readonly static Lazy<Task<Platform>> s_instance = new Lazy<Task<Platform>>(LoadInstance);
+
+        /// <summary>
+        /// Asyncrhonously get execution platform information.
+        /// </summary>
+        /// <returns>A task containing the execution platform information.</returns>
+        public static Task<Platform> InstanceAsync() => s_instance.Value;
+
+        /// <summary>
+        /// Get execution platform information. This may block for up to one second.
+        /// </summary>
+        /// <returns>Execution platform information.</returns>
+        public static Platform Instance() => InstanceAsync().Result;
+
+        private static async Task<GcePlatformDetails> LoadGceDetails()
+        {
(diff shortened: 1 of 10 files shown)


src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Get execution platform information. This may block for up to one second.
+        /// </summary>
+        /// <returns>Execution platform information.</returns>
+        public static Platform Instance() => InstanceAsync().Result;
+
+        private static async Task<GcePlatformDetails> LoadGceDetails()
+        {
+            const string metadataUrl = "http://metadata.google.internal/computeMetadata/v1?recursive=true";
```
evildour: You may want to look for the METADATA_EMULATOR_HOST environment variable and use that value if present instead of "metadata.google.internal"

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+        public string InstanceId { get; }
+        public string ZoneName { get; }
+    }
+
+    public sealed class GaePlatformDetails
+    {
+        public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeVersion)
+        {
+            ProjectId = gcloudProject;
+
```
evildour: Are the other two properties supposed to be set here from the unused parameters?

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
```
Deren-Liao: Do you need to add the licensing disclaimer ?  "Licensed under the Apache License ... "

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Net.Http;
+using System.Threading;
+using System.Net;
+
+namespace Google.Api.Gax
+{
+    public enum PlatformType
```
Deren-Liao: Can we add a TODO here to indicate we'll add GKE later?

src/Google.Api.Gax/Platform.cs
Deren-Liao: Same question.  
Just wanted to be sure, Ivan said Module ID is now renamed as Service.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <returns>A task containing the execution platform information.</returns>
+        public static Task<Platform> InstanceAsync() => s_instance.Value;
+
+        /// <summary>
+        /// Get execution platform information. This may block for up to one second.
+        /// </summary>
+        /// <returns>Execution platform information.</returns>
+        public static Platform Instance() => InstanceAsync().Result;
+
+        private static async Task<GcePlatformDetails> LoadGceDetails()
```
Deren-Liao: Add comment.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+            const string metadataFlavorKey = "Metadata-Flavor";
+            const string metadataFlavorValue = "Google";
+            const long maxContentLength = 512 * 1024; // Maximum allowed metadata size.
+            var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataUrl);
+            httpRequest.Headers.Add(metadataFlavorKey, metadataFlavorValue); // Required for any query.
+            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(1000)); // 1000 found to be reasonable.
+            var httpClient = new HttpClient();
+            var response = await httpClient.SendAsync(httpRequest, cts.Token);
+            IEnumerable<string> metadataValues;
+            if (response.StatusCode == HttpStatusCode.OK
```
Deren-Liao: If timed out, can response be null?

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+                && metadataValues.Contains(metadataFlavorValue)
+                && response.Content.Headers.ContentLength < maxContentLength)
+            {
+                // Valid response from metadata server.
+                string metadataJson = await response.Content.ReadAsStringAsync();
+                try
+                {
+                    return new GcePlatformDetails(metadataJson);
+                }
+                catch
```
Deren-Liao: In general, we do not catch all exceptions.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            const string metadataUrl = "http://metadata.google.internal/computeMetadata/v1?recursive=true";
+            const string metadataFlavorKey = "Metadata-Flavor";
+            const string metadataFlavorValue = "Google";
+            const long maxContentLength = 512 * 1024; // Maximum allowed metadata size.
+            var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataUrl);
+            httpRequest.Headers.Add(metadataFlavorKey, metadataFlavorValue); // Required for any query.
+            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(1000)); // 1000 found to be reasonable.
+            var httpClient = new HttpClient();
+            var response = await httpClient.SendAsync(httpRequest, cts.Token);
```
Deren-Liao: Do you consider to try catch all network I/O exception and do retry ?  Same comment to ReadAsStringAsync.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+        public GcePlatformDetails GceDetails { get; }
+
+        /// <summary>
+        /// The current execution platform.
+        /// </summary>
+        public PlatformType Type =>
+            GaeDetails != null ? PlatformType.Gae :
+            GceDetails != null ? PlatformType.Gce :
+            PlatformType.Unknown;
+
```
Deren-Liao: Is this an extra space?

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+            return null;
+        }
+
+        private static GaePlatformDetails LoadGaeDetails()
+        {
+            // See https://cloud.google.com/appengine/docs/flexible/python/runtime#environment_variables
+            // for details on environment variables.
+            var gcloudProject = Environment.GetEnvironmentVariable("GCLOUD_PROJECT");
+            var gaeInstance = Environment.GetEnvironmentVariable("GAE_INSTANCE");
```
Deren-Liao: Is this GAE_INSTANCE same as Service name (or Module Id) ?  The log entries need input of Module ID (Service Name), Version,  region.   Region is not used for Logs Viewer, so it is optional.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,150 @@
...
+        public GcePlatformDetails(string metadataJson)
+        {
+            MetadataJson = metadataJson;
+            var json = Newtonsoft.Json.Linq.JObject.Parse(metadataJson);
+            ProjectId = json["project"]["projectId"].ToString();
+            InstanceId = json["instance"]["id"].ToString();
+            ZoneName = json["instance"]["zone"].ToString();
+        }
+
+        public string MetadataJson { get; }
```
Deren-Liao: Unless you have other use case,  MetadataJson can be keep as private ?

src/Google.Api.Gax/Platform.cs
chrisdunelm: Oops, yes, thanks.
ModuleId is now renamed to ServiceId
And I've added an InstanceId property.

src/Google.Api.Gax/Platform.cs
chrisdunelm: Where is METADATA_EMULATOR_HOST documented?
I've added a TODO for the moment.

src/Google.Api.Gax/Platform.cs
chrisdunelm: I've copied how it's done in the auth library in [ComputeCredential.cs](https://github.com/google/google-api-dotnet-client/blob/a03d6061f3195163fe39449070d7166c757b63fd/Src/Support/GoogleApis.Auth/OAuth2/ComputeCredential.cs#L108).
I've added a TODO to consider retrying on IO Exception

src/Google.Api.Gax/Platform.cs
chrisdunelm: Timeout throws an exception.
I've changed the code to catch it in a similar way to how ComputeCredential handles it.

src/Google.Api.Gax/Platform.cs
chrisdunelm: True, but I'm not sure what to do here.
I've added a TODO to decide what to do.

src/Google.Api.Gax/Platform.cs
chrisdunelm: No, GAE_INSTANCE is different from GAE_SERVICE.
This is now fixed.

src/Google.Api.Gax/Platform.cs
chrisdunelm: Removed.

chrisdunelm:
Hi Deren,
1. Existing code in [ComputeCredential.cs](https://github.com/google/google-api-dotnet-client/blob/a03d6061f3195163fe39449070d7166c757b63fd/Src/Support/GoogleApis.Auth/OAuth2/ComputeCredential.cs#L108) doesn't handle retry, although if we can do better then we should. Ideally I'd like to find out if there's a way to detect GCE (and GKE later) locally, so network use isn't required.
2. Service name has been added.

src/Google.Api.Gax/Platform.cs
evildour: It isn't documented anywhere at the moment, but for the metadata client library, I started using this variable to conform to the pattern used by the other emulators ([here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/common.py#L20) and [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataClientImpl.cs#L47)).

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        /// <summary>
+        /// Get execution platform information. This may block for up to one second.
+        /// </summary>
+        /// <returns>Execution platform information.</returns>
+        public static Platform Instance() => InstanceAsync().Result;
+
+        private static async Task<GcePlatformDetails> LoadGceDetails()
+        {
+            // TODO: Respect METADATA_EMULATOR_HOST environment variable. Where is this documented?
```
Deren-Liao: My 2 cents.  If the emulator pretends this is GCE, I'd expect this LoadGceDetails pretends it gets all GCE metadata too.  Which means, I don't think we need to check if this is emulator.

evildour: I'm not sure I understand. Without checking for the variable, anyone testing locally with the emulator running would get no GCE details. The emulator listens on a localhost port, not on metadata.google.internal. Are you saying that this should intentionally not return info from the emulator?

All that said, the emulator is not a part of the Cloud SDK, so I don't know if this needs to be supported at this time anyway. But without this support, it makes it more difficult for someone to use this (or another) emulator for local testing.

Deren-Liao:
Good to learn. My understanding was opposite (and wrong).  I thought
checking the environment variable would skipping emulator to treat it as
non-GCE .   Yes, sure we want people using emulator to be able to submit
logs as if it is on a real GCE instance.

src/Google.Api.Gax/Platform.cs
chrisdunelm: The environment variables present on GAE-flex C# (custom runtime) don't match the [python docs](https://cloud.google.com/appengine/docs/flexible/python/runtime#environment_variables).
I'm not yet clear why, but this means GAE-detection currently fails.
I'll try to sort out why this is.

chrisdunelm: OK, this is sorted out. I was running on the wrong platform (managed VM, rather than AppEngine-flex).

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,220 @@
...
+            // Check if emulator is in use by looking for an emulator host in environment variable
+            // METADATA_EMULATOR_HOST. This is the undocumented but de-facto mechanism for doing this.
+            var metadataEmulatorHost = Environment.GetEnvironmentVariable("METADATA_EMULATOR_HOST");
+            const string metadataUrl = "http://metadata.google.internal/computeMetadata/v1?recursive=true";
+            const string metadataFlavorKey = "Metadata-Flavor";
+            const string metadataFlavorValue = "Google";
+            const long maxContentLength = 512 * 1024; // Maximum allowed metadata size.
+            try
+            {
+                var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataEmulatorHost ?? metadataUrl);
```
chrisdunelm: Now checks METADATA_EMULATOR_HOST.

src/Google.Api.Gax/Platform.cs
```diff
@@ -78,21 +105,23 @@ public sealed class Platform
...
+            // METADATA_EMULATOR_HOST. This is the undocumented but de-facto mechanism for doing this.
+            var metadataEmulatorHost = Environment.GetEnvironmentVariable("METADATA_EMULATOR_HOST");
             const string metadataUrl = "http://metadata.google.internal/computeMetadata/v1?recursive=true";
             const string metadataFlavorKey = "Metadata-Flavor";
             const string metadataFlavorValue = "Google";
             const long maxContentLength = 512 * 1024; // Maximum allowed metadata size.
             try
             {
-                var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataUrl);
+                var httpRequest = new HttpRequestMessage(HttpMethod.Get, metadataEmulatorHost ?? metadataUrl);
```
evildour: This will not work. The METADATA_EMULATOR_HOST value, if present, will be something like "localhost:12345", so you would want to use this url: `$"http://{metadataEmulatorHost}/computeMetadata/v1?recursive=true"`

chrisdunelm: Oops ;) I'll fix it soon.

evildour: Btw, here's how I use this in the metadata library: [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataClientImpl.cs#L55) and [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataClientImpl.cs#L55). I think at one point we decided that it was safe to just check for the variable once because its unlikely the presence of the emulator is going to change during the test run of the application.

chrisdunelm: Thanks, this is now done.
It does only check it once, as LoadInstance() is only called once, from the Lazy s_instance static field.

chrisdunelm:
PTAL

evildour:
Metadata emulator check looks good to me. So does everything else, but might be best to let Deren re-review as well.

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,267 @@
...
+        /// Asyncrhonously get execution platform information.
+        /// </summary>
+        /// <returns>A task containing the execution platform information.</returns>
+        public static Task<Platform> InstanceAsync() => s_instance.Value;
+
+        /// <summary>
+        /// Get execution platform information. This may block for up to one second.
+        /// </summary>
+        /// <returns>Execution platform information.</returns>
+        public static Platform Instance() => InstanceAsync().Result;
```
Deren-Liao: I am not quite clear if the Platform Instance result cached here.  Can you make sure Platform object itself is cached?

src/Google.Api.Gax/Platform.cs
```diff
@@ -0,0 +1,267 @@
...
+            $"[GAE: ProjectId='{ProjectId}', InstanceId='{InstanceId}', ServiceId='{ServiceId}', VersionId='{VersionId}']";
+    }
+
+    /// <summary>
+    /// Information about the current execution platform.
+    /// Supported execption platforms are Google App Engine (GAE) and Google Compute Engine (GCE).
+    /// </summary>
+    public sealed class Platform
+    {
+        private readonly static Lazy<Task<Platform>> s_instance = new Lazy<Task<Platform>>(LoadInstance);
```
Deren-Liao: Can you consider to add a static constructor so as to

PR #132, opened 2017-02-08
Further tweaks to x-goog-api-client header

jskeet:
Sorry for going back and forth on this...

### DIFF
--- a/src/Google.Api.Gax.Grpc/UserAgentBuilder.cs
+++ b/src/Google.Api.Gax.Grpc/UserAgentBuilder.cs
@@ -48,10 +48,10 @@ namespace Google.Api.Gax.Grpc
         /// </summary>
         internal UserAgentBuilder AppendDotNetEnvironment()
 #if NETSTANDARD1_5
-            // TODO: Improve this if we can.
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax.Grpc/UserAgentBuilder.cs
```diff
@@ -48,10 +48,10 @@ internal UserAgentBuilder AppendAssemblyVersion(string name, System.Type type)
         /// </summary>
         internal UserAgentBuilder AppendDotNetEnvironment()
 #if NETSTANDARD1_5
-            // TODO: Improve this if we can.
-            => AppendVersion("dotnet", "unknown");
+            // TODO: Improve this if we can. (At least 0.0.0 is valid semver, so should parse.)
+            => AppendVersion("gl-dotnet", "0.0.0");
```
bjwatson: @jskeet @chrisdunelm This unknown value should be an empty string (`""`).

jskeet: It was changed in https://github.com/googleapis/gax-dotnet/pull/134 to fetch the version for .NET Core.

PR #133, opened 2017-02-09
Move gl-dotnet to start of custom header

jskeet:
(Also, update a test to check for gl-dotnet rather than just dotnet.)

### DIFF
--- a/src/Google.Api.Gax.Grpc/ServiceSettingsBase.cs
+++ b/src/Google.Api.Gax.Grpc/ServiceSettingsBase.cs
@@ -22,10 +22,10 @@ namespace Google.Api.Gax.Grpc
         protected ServiceSettingsBase()
         {
             UserAgent = new UserAgentBuilder()
+                .AppendDotNetEnvironment()
                 .AppendAssemblyVersion("gccl", GetType())
                 .AppendAssemblyVersion("gax", typeof(CallSettings))
                 .AppendAssemblyVersion("grpc", typeof(Channel))
-                .AppendDotNetEnvironment()
                 .ToString();
         }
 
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax.Grpc/ServiceSettingsBase.cs
```diff
@@ -22,10 +22,10 @@ public abstract class ServiceSettingsBase
         protected ServiceSettingsBase()
         {
             UserAgent = new UserAgentBuilder()
+                .AppendDotNetEnvironment()
                 .AppendAssemblyVersion("gccl", GetType())
```
bjwatson: Can you add this line:

```C#
AppendAssemblyVersion("gapic", GetType())
```

I know it's duplicative, but it will help keep things consistent across all languages.

bjwatson:
I have one comment on the code to capture a recent change to show both `gccl` and `gapic`, because we want to report `gapic` across the board. For C#, it's fine to use the same version for both, since they're so tightly integrated.

Also, do you have PRs to update `toolkit` and `google-cloud-dotnet`?

bjwatson:
One other question is do you use the `"UNKNOWN"` string anywhere? If so, it needs to be changed to the empty string (e.g. `grpc/` rather than `grpc/UNKNOWN`).

jskeet:
@bjwatson:  I'll add in the gapic header, although it seems a bit pointless to me. We don't use "unknown" anywhere, and we don't need to update any other repos. This piece of code is the only place that sets this header.

bjwatson:
Thanks @jskeet. I know it seems pointless in the context of .NET, but I'm thinking in terms of having a high-level of consistency across all languages in our metrics. When constructing a query, we want to avoid special cases that we have to deal with, like C# really has a GAPIC layer even though the `gapic` field is NULL.

PR #134, opened 2017-02-09
NetCore DotNet Environment

chrisdunelm:


chrisdunelm:
Sadly, [System.Version](https://msdn.microsoft.com/en-us/library/system.version) is specified as returning -1 for any component that isn't set :(

PR #142, opened 2017-02-28
Add empty dictionary class

chrisdunelm:


src/Google.Api.Gax/EmptyDictionary.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    /// <typeparam name="TKey">The type of the keys in the dictionary.</typeparam>
+    /// <typeparam name="TValue">The type of the values in the dictionary.</typeparam>
+    public static class EmptyDictionary<TKey, TValue>
+    {
+        /// <summary>
+        /// Gets a cached empty <see cref="IDictionary{TKey, TValue}"/>. The returned dictionary is read-only.
+        /// </summary>
+        public static IDictionary<TKey, TValue> Instance { get; } =
+            new ReadOnlyDictionary<TKey, TValue>(new Dictionary<TKey, TValue>(0));
```
jskeet: Is ReadOnlyDictionary supported everywhere we need it? That's handy - I thought it wasn't, which is why Google.Protobuf has its own one. (It could be that protobuf supports earlier versions of netcore than we do.)

chrisdunelm: Yes, from .NET4.5, and netstandard1.0 :)

jskeet: Ah, of course - we need it for .NET 3.5, I suspect...

PR #144, opened 2017-03-01
Add(IEnumerable<T>) method to ResourceNameList

chrisdunelm:
To allow adding an enumerable of items using object initializer syntax.

### DIFF
--- a/src/Google.Api.Gax/ResourceNameList.cs
+++ b/src/Google.Api.Gax/ResourceNameList.cs
@@ -52,6 +52,19 @@ namespace Google.Api.Gax
         /// <inheritdoc />
         public void Add(TName item) => _underlyingList.Add(_nameToString(item));
 
+        /// <summary>
+        /// Adds all items to this list.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax/ResourceNameList.cs
```diff
@@ -52,6 +52,18 @@ public ResourceNameList(IList<string> underlyingList, Func<string, TName> string
         /// <inheritdoc />
         public void Add(TName item) => _underlyingList.Add(_nameToString(item));
 
+        /// <summary>
+        /// Adds all items to this list.
+        /// </summary>
+        /// <param name="items">The items to add to this list.</param>
+        public void Add(IEnumerable<TName> items)
+        {
+            foreach (TName item in GaxPreconditions.CheckNotNull(items, nameof(items)))
```
jskeet: I'd probably move the nullity check to a separate line here, but I don't mind much.

chrisdunelm: Done.
Will merge on green

PR #146, opened 2017-03-06
Include 'complete' message in streaming buffer

### LINKED ISSUE #122: Include the "complete" message in buffer size when streaming
That will make a BufferedClientStreamWriter with a buffer size of 1 equivalent to using the interface directly.

chrisdunelm:
Fixes #122

### DIFF
--- a/src/Google.Api.Gax.Grpc/BufferedClientStreamWriter.cs
+++ b/src/Google.Api.Gax.Grpc/BufferedClientStreamWriter.cs
@@ -117,7 +117,7 @@ namespace Google.Api.Gax.Grpc
             GaxPreconditions.CheckNotNull(message, nameof(message));
             lock (_lock)
             {
-                if (!ValidateStateForWrite(throwOnError))
+                if (!ValidateStateForWrite(throwOnError, false))
                 {
                     return null;
                 }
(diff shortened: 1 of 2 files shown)


jskeet:
Will take a look at this later today, with any luck. I need to look at it carefully, as we need to consider how the TryComplete is going to work... it was previously just "Complete if you can, but it's okay if it's already completed" but now we may need to make it obvious to the caller whether or not it couldn't complete because there wasn't buffer space.

chrisdunelm:
OK, thanks.
This PR doesn't address that issue at the moment. It simply returns null if the queue is full OR completion has already occurred.
(Also just noticed I haven't updated the xmldoc, I'll fix that now)

chrisdunelm:
Instead of returning Task from the [Try]Write and [Try]Complete methods, we could return a BufferedWriteResult, which encodes the various results:
* Queued successfully - provides the Task
* Queue full
* Stream completed

This would be a change of API, so needs doing soon if we'd like it.
What do you think?

jskeet:
I think it's probably simpler to expose a `Completed` property - possibly named in a slightly more specific way (`WriteCompleteCalled`?). That only ever goes from `false` to `true`, which helps in terms of not worrying about the value being out of date (or only being out of date in one direection). We may still want to have a way of finding the current buffer count, more for logging / diagnostics than anything else. All of that can come in another PR.

chrisdunelm:
SGTM :)

PR #147, opened 2017-03-10
Support GKE platform detection

chrisdunelm:


chrisdunelm:
All integration tests pass:
* run_tests.sh - GKE integration test
* run_tests.bat - GAE and GCE integration tests

(Yes, these should probably be either combined, or split into three separate tests)

src/Google.Api.Gax/Platform.cs
```diff
@@ -129,6 +163,85 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+        /// <param name="location">The location.</param>
+        public GkePlatformDetails(string metadataJson, string projectId, string clusterName, string location)
+        {
+            MetadataJson = metadataJson;
+            ProjectId = projectId;
+            ClusterName = clusterName;
+            Location = location;
+        }
+
+        /// <summary>
```
jskeet: Do we get the pod name? That would be the most useful part, IMO.

jskeet:
(It might be worth looking at environment variables for GKE. That might have pod names etc. I can't find docs around it at the moment...)

chrisdunelm:
There aren't many environment variables defined. This is the complete list from the instance running the integrationtest:
[DEBIAN_FRONTEND]: 'noninteractive'
[HOME]: '/root'
[HOSTNAME]: 'platformintegrationtest-1096079241-lq41w'
[KUBERNETES_PORT]: 'tcp://10.171.240.1:443'
[KUBERNETES_PORT_443_TCP]: 'tcp://10.171.240.1:443'
[KUBERNETES_PORT_443_TCP_ADDR]: '10.171.240.1'
[KUBERNETES_PORT_443_TCP_PORT]: '443'
[KUBERNETES_PORT_443_TCP_PROTO]: 'tcp'
[KUBERNETES_SERVICE_HOST]: '10.171.240.1'
[KUBERNETES_SERVICE_PORT]: '443'
[KUBERNETES_SERVICE_PORT_HTTPS]: '443'
[PATH]: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
[PLATFORMINTEGRATIONTEST_PORT]: 'tcp://10.171.243.115:8080'
[PLATFORMINTEGRATIONTEST_PORT_8080_TCP]: 'tcp://10.171.243.115:8080'
[PLATFORMINTEGRATIONTEST_PORT_8080_TCP_ADDR]: '10.171.243.115'
[PLATFORMINTEGRATIONTEST_PORT_8080_TCP_PORT]: '8080'
[PLATFORMINTEGRATIONTEST_PORT_8080_TCP_PROTO]: 'tcp'
[PLATFORMINTEGRATIONTEST_SERVICE_HOST]: '10.171.243.115'
[PLATFORMINTEGRATIONTEST_SERVICE_PORT]: '8080'
[PORT]: '8080'

jskeet:
Hmm... that hostname looks like it could be the pod name.

That's supported by https://kubernetes.io/docs/admin/dns/ - but it looks like it can be overridden. Maybe it would be worth including the hostname in the details anyway, and let users do what they like with it?

chrisdunelm:
Added hostname from metadata (rather than environment variable).

PR #151, opened 2017-03-15
Prepare PollSettings for backoff

chrisdunelm:


chrisdunelm:
Actually, not sure even this is necessary.
The initial delay property in [BackoffSettings](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax.Grpc/BackoffSettings.cs#L21) is simply called `Delay`.
So this PR can probably just be abandoned.

jskeet:
Yup, SGTM. I'm still hopeful for constant delay, but we'll see :)

PR #153, opened 2017-03-21
Reorganize version header generation code

jskeet:
This is now desirable for both REST and gRPC APIs, so it's in the
Google.Api.Gax package, and is renamed to VersionHeaderBuilder as
it's not building a User-Agent header.

(Most of the changes are just renames.)


src/Google.Api.Gax.Grpc/ApiServerStreamingCall.cs
```diff
@@ -66,7 +66,7 @@ public AsyncServerStreamingCall<TResponse> Call(TRequest request, CallSettings p
         internal ApiServerStreamingCall<TRequest, TResponse> WithUserAgent(string userAgent) =>
```
chrisdunelm: userAgent -> versionHeader

src/Google.Api.Gax.Grpc/MetadataExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static void Add(this Metadata metadata, IEnumerable<Metadata.Entry> items
         public static Metadata WithUserAgent(this Metadata metadata, string userAgent)
```
chrisdunelm: WithVersionHeader

src/Google.Api.Gax.Rest/UserAgentHelper.cs
```diff
@@ -6,8 +6,7 @@
  */
```
chrisdunelm: Should this class name now be changed to VersionHeaderHelper?

src/Google.Api.Gax/AssemblyInfo.cs
```diff
@@ -10,3 +10,5 @@
...
 [assembly:InternalsVisibleTo("Google.Api.Gax.Tests,PublicKey=0024000004800000940000000602000000240000525341310004000001000100bfa2d20e82583e5f3a47efa518d8a7cda58d6e1b61f51ed3de3900f238fb01e0ca95b3ed1c68147e4865a336b252f070d5207e653c9014d271fc829c18981d1c80d418d300545c3ad0e1ff4d77c596177dbfe0a588fff57306f1d6ba655de3cfa520ad777ea73bb7a92fd6027071c94bb12d5871fcd076299de0502bca12e3aa")]
 [assembly:InternalsVisibleTo("Google.Api.Gax.Grpc,PublicKey=0024000004800000940000000602000000240000525341310004000001000100bfa2d20e82583e5f3a47efa518d8a7cda58d6e1b61f51ed3de3900f238fb01e0ca95b3ed1c68147e4865a336b252f070d5207e653c9014d271fc829c18981d1c80d418d300545c3ad0e1ff4d77c596177dbfe0a588fff57306f1d6ba655de3cfa520ad777ea73bb7a92fd6027071c94bb12d5871fcd076299de0502bca12e3aa")]
+[assembly: InternalsVisibleTo("Google.Api.Gax.Rest,PublicKey=0024000004800000940000000602000000240000525341310004000001000100bfa2d20e82583e5f3a47efa518d8a7cda58d6e1b61f51ed3de3900f238fb01e0ca95b3ed1c68147e4865a336b252f070d5207e653c9014d271fc829c18981d1c80d418d300545c3ad0e1ff4d77c596177dbfe0a588fff57306f1d6ba655de3cfa520ad777ea73bb7a92fd6027071c94bb12d5871fcd076299de0502bca12e3aa")]
```
chrisdunelm: Slightly confused as to why internals are visible to a non-test assembly?

test/Google.Api.Gax.Rest.Tests/UserAgentHelperTest.cs
```diff
@@ -5,6 +5,8 @@
  * https://developers.google.com/open-source/licenses/bsd
  */
```
chrisdunelm: (And this test might want to change name to VersionHeaderHelperTest)

src/Google.Api.Gax.Rest/UserAgentHelper.cs
jskeet: Not sure. It still has code to use with the User-Agent as well... unlike the gRPC one.

src/Google.Api.Gax.Grpc/MetadataExtensions.cs
jskeet: Gah, thanks.

src/Google.Api.Gax/AssemblyInfo.cs
jskeet: We want to be able to get at this from Google.Api.Gax.Rest, but not from other code. Note how the previous line is similar for Grpc, although I'm not sure which internal type that's related to.

src/Google.Api.Gax.Grpc/ApiServerStreamingCall.cs
jskeet: Done. There was another one too...

src/Google.Api.Gax.Rest/UserAgentHelper.cs
jskeet: Have updated docs to include "and related headers" which I think is okay :) (And it means we don't need to make breaking changes outside GAX)

jskeet:
Right - I've now done a search for "gent" to make sure it doesn't occur anywhere we don't want it :)

PR #156, opened 2017-03-23
Common protos

jskeet:
Suggested reviewing strategy:

- Review first commit (new script)
- Skim second commit (entirely autogenerated)
- Review tiny third commit


jskeet:
Looking into the travis failure...

jskeet:
Okay, fixed and pushed change - Travis should go green in a bit. Please keep reviewing :) (The only change was to the third commit.)

jskeet:
Have just noticed that a couple of files that should have been in the second commit are in the first. I can try to fix if you like, but it shouldn't make any difference.

chrisdunelm:
It's fine as is. I'm reviewing now.

generateprotos.sh
```diff
@@ -0,0 +1,82 @@
...
+$PROTOC \
+  -I googleapis \
+  -I $CORE_PROTOS_ROOT \
+  --csharp_out=$OUTDIR \
+  --csharp_opt=base_namespace=Google \
+  googleapis/google/api/*.proto \
+  $(find googleapis/google/rpc -name '*.proto') \
+  $(find googleapis/google/type -name '*.proto')
+  
+rm -rf TARGETDIR/{Api,Rpc,Type}
```
chrisdunelm: $TARGETDIR

jskeet: Shouldn't be there at all, in fact :) Gone.

PR #159, opened 2017-03-27
Add source context support.

Deren-Liao:
A proto generated SourceContext.cs.
And helper methods SourceRevision.cs.


jskeet:
I'd really prefer not to have this in GAX unless there's some reason it's really useful across multiple APIs. It feels like something that belongs either in a separate library in google-cloud-dotnet, or in one of the existing ones.

If it *must* be in a GAX library, it should be in the Grpc one - there's no reason why (say) Google.Cloud.Storage.V1 should end up pulling in Google.Protobuf (which it would) just to have these classes available.

What's the use case here? The better we can all understand it, the better we'll be able to decide where it goes. (As an aside, we may well want to fix the namespace away from Google.Devtools, as we have done for other APIs where the proto is in that package.)

Deren-Liao:
The design doc is at go/*netsourcelocation .  *The plan is to share it with
both Log4Net appender and Google.Cloud.Diagnostics.AspNet.
In short, this PR reads the SourceContext file if it exists.  Google
Log4Net appender or the cloud diagnostics will append the git SHA to log
entry.

jskeet:
Okay - will have a look and comment on the doc. @chrisdunelm you'll probably want to do that too :)

Deren-Liao:
I'll move the files to a separate dll at cloud dot net repo.

PR #164, opened 2017-03-30
Update to latest version of Google.Apis.*

jskeet:

PR #165, opened 2017-03-30
Update GAX to v1.0.0

jskeet:
(This will be one commit when #164 is merged.)


src/Google.Api.Gax.Rest/project.json
```diff
@@ -18,8 +18,8 @@
 
   "dependencies": {
     "Google.Api.Gax": { "target": "project" },
-    "Google.Apis.Auth": "1.22.0",
```
chrisdunelm: Why is this dep here? Doesn't Google.Api.Gax bring it in?

PR #183, opened 2017-05-26
Update GAX to 2.0.0-beta02

jskeet:
Anything else we should do before releasing beta02?

It's mainly just got new dependencies...


jskeet:
Actually, let's add the configure-await checker here too...

PR #192, opened 2017-06-29
Move to v2.1.0

chrisdunelm:


jskeet:
When we do want to do this, we should do the testing projects too.
Maybe we should have a single root XML file that has the version in one place...

PR #194, opened 2017-06-30
Make it easier to specify a FieldMask for CallSettings

### LINKED ISSUE #189: Add FieldMask to CallSettings
It should be possible to set a field mask when making a call.

This should be propagated in the X-Goog-FieldMask header.

jskeet:
(The user doesn't need to know the right header.)

Fixes #189 (although we may want more later.)

### DIFF
--- a/src/Google.Api.Gax.Grpc/CallSettings.cs
+++ b/src/Google.Api.Gax.Grpc/CallSettings.cs
@@ -16,6 +16,8 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class CallSettings
     {
+        internal const string FieldMaskHeader = "x-goog-fieldmask";
+
         /// <summary>
         /// Constructs an instance with the specified settings.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Google.Api.Gax.Grpc/CallSettings.cs
```diff
@@ -154,5 +156,24 @@ public static CallSettings FromHeader(string name, string value)
...
+        /// </summary>
+        /// <remarks>
+        /// The precise effect on the request is not guaranteed: it may be through a header or a side-channel,
+        /// for example. Likewise the effect of combining multiple settings containing field masks is not specified.
+        /// </remarks>
+        /// <param name="fieldMask">The field mask for the request. Must not be null.</param>
+        /// <returns>A new instance.</returns>
+        public static CallSettings FromFieldMask(string fieldMask)
+        {
+            GaxPreconditions.CheckNotNull(fieldMask, nameof(fieldMask));
```
chrisdunelm: Could delegate to `FromHeader()`, so this `CheckNotNull` isn't required here.

jskeet: Except then the parameter name would be wrong...

jskeet:
Whoops - looks like the tests fail because I didn't know about lower-casing of headers. Will fix that.

src/Google.Api.Gax.Grpc/CallSettings.cs
chrisdunelm: OK, fair enough.

PR #195, opened 2017-06-30
Add more GaxPreconditions

chrisdunelm:


src/Google.Api.Gax/GaxPreconditions.cs
```diff
@@ -237,13 +338,7 @@ public static void CheckArgument<T>(bool condition, string paramName, string for
...
         /// <param name="paramName">The name of the parameter whose value is being tested.</param>
-        internal static TimeSpan CheckNonNegativeDelay(TimeSpan value, string paramName)
-        {
-            if (value < TimeSpan.Zero)
-            {
-                throw new ArgumentOutOfRangeException(nameof(value), value, "Delay must not be negative");
-            }
-            return value;
-        }
+        internal static TimeSpan CheckNonNegativeDelay(TimeSpan value, string paramName) =>
```
jskeet: Possibly rename to "TimeSpan" or "Duration" rather than "Delay"? This might not be a delay. (Ditto the comment)

chrisdunelm: This method was already present, I only re-factored it slightly.
I can certainly rename it if you'd like...

jskeet: Best to leave it now... internals being visible to other versions make this tricky. (We should think about that carefully at some point.)

PR #196, opened 2017-06-30
Generic Range Precondition checks

chrisdunelm:


chrisdunelm:
Comment added.
Agreed about the performance. In fact, having the int case be non-generic may be beneficial as this will probably be the most commonly used.

PR #199, opened 2017-07-13
Pre-release changes for GAX

jskeet:
- Adds SourceLink
- Adds a release build script
- Updates the version to 2.1.0-beta01


jskeet:
Ah, got it. We're still trying to pack the net45 DLL even on Linux. Oops.

PR #201, opened 2017-08-14
Make it easier to create CallSettings and RetrySettings

jskeet:
- RetrySettings.WithExpiration already existed, this just makes it public
- CallSettings.WithCallExpiration is new

This removes the need for the code in https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/CallSettingsMutations.cs
It doesn't replace the CallTiming extension method, but that was
only used by CallSettings.WithCallExpiration. We can consider adding
that later.

### DIFF
--- a/src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
+++ b/src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
@@ -147,6 +147,24 @@ namespace Google.Api.Gax.Grpc
         }
 
         /// <summary>
+        /// Returns a new <see cref="CallSettings"/> with the given the expiration, merged with the (optional) original settings specified by <paramref name="settings"/>.
+        /// </summary>
+        /// <remarks>
+        /// If <paramref name="settings"/> is non-null and has retry-based timing, the returned settings will have the same timing, but with the specified
+        /// expiration. Otherwise, the returned settings will have a simple expiration.
+        /// </remarks>
+        /// <param name="settings">Existing settings. May be null, meaning there are currently no settings.</param>
+        /// <param name="expiration">Expiration to use in the returned settings, possibly as part of a retry. Must not be null.</param>
+        /// <returns></returns>
+        public static CallSettings WithExpiration(this CallSettings settings, Expiration expiration)
+        {
+            GaxPreconditions.CheckNotNull(expiration, nameof(expiration));
+            var retry = settings?.Timing?.Retry?.WithTotalExpiration(expiration);
+            var timing = retry == null ? CallTiming.FromExpiration(expiration) : CallTiming.FromRetry(retry);
+            return settings.WithCallTiming(timing);
+        }
+
+        /// <summary>
         /// Transfers settings contained in this into a <see cref="CallOptions"/>.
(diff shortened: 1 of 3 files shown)


src/Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -147,6 +147,24 @@ public static CallSettings WithEarlierDeadline(this CallSettings settings, DateT
...
+        /// Returns a new <see cref="CallSettings"/> with the given the expiration, merged with the (optional) original settings specified by <paramref name="settings"/>.
+        /// </summary>
+        /// <remarks>
+        /// If <paramref name="settings"/> is non-null and has retry-based timing, the returned settings will have the same timing, but with the specified
+        /// expiration. Otherwise, the returned settings will have a simple expiration.
+        /// </remarks>
+        /// <param name="settings">Existing settings. May be null, meaning there are currently no settings.</param>
+        /// <param name="expiration">Expiration to use in the returned settings, possibly as part of a retry. Must not be null.
+        /// <returns></returns>
+        public static CallSettings WithCallExpiration(this CallSettings settings, Expiration expiration)
```
chrisdunelm: Why not put this in CallSettings directly, rather than use an extension method?

jskeet: As `settings` can be null - if there weren't any specific settings before, it'll just create a new one.

chrisdunelm: Oh yes, and I should have read the xmldoc more thoroughly!

chrisdunelm:
Looks good to me.

test/Google.Api.Gax.Grpc.Tests/CallSettingsExtensionsTest.cs
```diff
@@ -282,5 +282,9 @@ public void WithEarlierDeadline_DeadlineIsEarlierThanExistingRetryTotalExpiratio
             Assert.Equal(token, newSettings.CancellationToken);
         }
 
+        [Fact]
+        public void WithExpiration_FIXME()
```
benwulfe: forgot this?

jskeet: Not forgotten - referred to in the PR comment. The plan was to get you to check that I was implementing the appropriate feature, then I'll implement the test :)

jskeet:
Okay, we now have tests. I'm not entirely sure about the CallSettings extension method name though. Options:

- WithExpiration
- WithCallExpiration (current option)
- WithTotalExpiration

Thoughts?

chrisdunelm:
I don't see there's much in it really. I probably very slightly prefer WithExpiration(), as it's the shortest and I don't see adding "call" or "total" adds anything much.

jskeet:
SGTM - will do that now.

jskeet:
Done, and force-pushed. PTAL.

PR #202, opened 2017-08-16
Update PollSettings to support exponential backoff

chrisdunelm:
If this looks OK, I'll add the backoff code in Polling.cs, and extra tests.

### DIFF
--- a/src/Google.Api.Gax/PollSettings.cs
+++ b/src/Google.Api.Gax/PollSettings.cs
@@ -42,10 +42,10 @@ namespace Google.Api.Gax
         public TimeSpan MaxDelay { get; }
 
         /// <summary>
-        /// Creates poll settings from the given expiration and delay.
+        /// Creates poll settings from the given expiration and constant delay.
         /// </summary>
         /// <param name="expiration">The expiration to use in order to know when to stop polling. Must not be null.</param>
-        /// <param name="delay">The delay between RPC calls. Must be non-negative.</param>
+        /// <param name="delay">The constant delay between RPC calls. Must be non-negative.</param>
         public PollSettings(Expiration expiration, TimeSpan delay) : this (expiration, delay, 1.0, delay) { }
 
         /// <summary>
@@ -66,6 +66,20 @@ namespace Google.Api.Gax
             }
             DelayMultiplier = delayMultiplier;
             MaxDelay = GaxPreconditions.CheckNonNegativeDelay(maxDelay, nameof(maxDelay));
+        }
+
+        /// <summary>
+        /// Works out the next delay from the current one, based on the multiplier and maximum.
+        /// </summary>
+        /// <param name="currentDelay">The current delay.</param>
+        /// <returns>The next delay.</returns>
+        internal TimeSpan NextDelay(TimeSpan currentDelay)
+        {
+            checked
+            {
+                TimeSpan next = new TimeSpan((long)(currentDelay.Ticks * DelayMultiplier));
+                return next < MaxDelay ? next : MaxDelay;
+            }
         }
     }
 }
--- a/src/Google.Api.Gax/PollSettings.cs
+++ b/src/Google.Api.Gax/PollSettings.cs
@@ -29,16 +29,43 @@ namespace Google.Api.Gax
         /// and the next RPC request being sent.
         /// </remarks>
         public TimeSpan Delay { get; }
-        
+
         /// <summary>
-        /// Creates poll settings from the given expiration, delay and call settings.
+        /// The multiplier to apply to the delay on each iteration; must be greater or equal to 1.0.
+        /// </summary>
+        public double DelayMultiplier { get; }
+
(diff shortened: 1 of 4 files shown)


src/Google.Api.Gax/PollSettings.cs
```diff
@@ -29,16 +29,43 @@ public sealed class PollSettings
...
+        public double DelayMultiplier { get; }
+
+        /// <summary>
+        /// The maximum delay to use. If the increasing delay due to the delay multiplier exceeds this,
+        /// this maximum is used instead.
+        /// </summary>
+        public TimeSpan MaxDelay { get; }
+
+        /// <summary>
+        /// Creates poll settings from the given expiration and delay.
```
jskeet: Maybe make it clear that it's then a constant delay.

src/Google.Api.Gax/PollSettings.cs
```diff
@@ -29,16 +29,43 @@ public sealed class PollSettings
...
         /// <param name="expiration">The expiration to use in order to know when to stop polling. Must not be null.</param>
         /// <param name="delay">The delay between RPC calls. Must be non-negative.</param>
-        public PollSettings(Expiration expiration, TimeSpan delay)
+        /// <param name="delayMultiplier">The multiplier to apply to the delay on each iteration; must be greater or equal to 1.0.</param>
+        /// <param name="maxDelay">The maximum delay to use.</param>
+        public PollSettings(Expiration expiration, TimeSpan delay, double delayMultiplier, TimeSpan maxDelay)
         {
             Expiration = GaxPreconditions.CheckNotNull(expiration, nameof(expiration));
             Delay = GaxPreconditions.CheckNonNegativeDelay(delay, nameof(delay));
+            if (delayMultiplier < 1.0 || double.IsNaN(delayMultiplier))
```
jskeet: Does infinity count as NaN? I can't remember.

chrisdunelm: No. But this validation is copied from [BackoffSettings.cs](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax.Grpc/BackoffSettings.cs#L54). I thought it best to make them consistent.

chrisdunelm:
I think the case of an LRO that has a very unpredictable runtime makes this reasonable.
If the RPC may take just a few seconds, or a few hours; then exponential backoff does make sense.

src/Google.Api.Gax/PollSettings.cs
jskeet: Ouch. We should probably fix it in both places :) (Perhaps add a common method somewhere?)

chrisdunelm: Is it worth it? Allowing something like double.MaxValue, but disallowing +infinity seems a little unnecessary.

src/Google.Api.Gax/PollSettings.cs
chrisdunelm: Done

chrisdunelm:
PTAL

src/Google.Api.Gax/PollSettings.cs
jskeet: Fair enough :)

src/Google.Api.Gax/PollSettings.cs
```diff
@@ -29,16 +29,57 @@ public sealed class PollSettings
...
+            DelayMultiplier = delayMultiplier;
+            MaxDelay = GaxPreconditions.CheckNonNegativeDelay(maxDelay, nameof(maxDelay));
+        }
+
+        /// <summary>
+        /// Works out the next delay from the current one, based on the multiplier and maximum.
+        /// </summary>
+        /// <param name="currentDelay">The current delay.</param>
+        /// <returns>The next delay.</returns>
+        internal TimeSpan NextDelay(TimeSpan currentDelay)
```
jskeet: Random alternative implementation: return an `IEnumerable<TimeSpan>`. Not really suggesting it, but would be interesting.

src/Google.Api.Gax/Polling.cs
```diff
@@ -59,15 +60,16 @@ public static class Polling
...
+                currentDelay = pollSettings.NextDelay(currentDelay);
             }
         }
 
         /// <summary>
         /// Asynchronously repeatedly calls the specified polling action, delaying between calls,
         /// until a given condition is met in the response.
         /// </summary>
-        /// <typeparam name="TResponse">The response type. Must not be null.</typeparam>
+        /// <typeparam name="TResponse">The response type.</typeparam>
```
jskeet: Thanks :)

src/Google.Api.Gax/PollSettings.cs
chrisdunelm: It'd certainly be neat ;) But possibly not very obvious!
This again is copied directly from BackoffSettings.

PR #205, opened 2017-08-30
Update dependencies

jskeet:
We probably don't want to merge this PR *exactly* as-is, due to the warnings we get for Grpc.Core 1.6.0 depending on System.Threading.ThreadPool 4.0.0 - but it can be reviewed apart from that, and we can find out what the plan is for Grpc.Core.


chrisdunelm:
What about the Google.Apis. deps in apis.json? For storage, translate, bigquery, compute?

jskeet:
Yes, I'll do those as well, once we've got GAX 2.1.0 out. (Or I can do them in parallel, as it doesn't hurt for them to have a higher version number.)

jskeet:
Now updated to use:

- xUnit 2.3.0-beta5
- Grpc.Core 1.6.1
- Google.Apis.* 1.29.1

PTAL, and this time it's ready to merge if you're happy.

PR #207, opened 2017-09-18
Add FlowControl and Batching settings to GAX

chrisdunelm:


src/Google.Api.Gax/FlowControlSettings.cs
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Settings used to control data flow.
```
jskeet: Possibly give more details about the scenario? Is this always for incoming data whereas BatchingSettings is always outgoing data, for example?

chrisdunelm: Hmm, I wrote the descriptions of these types deliberately to avoid mentioning usage scenarios, as they're not tied to any use. It happens that at the moment each type is only used in one particular way, but in theory that could change. The ctor/method/property where they are used should explain the usage.

jskeet: Okay - sounds reasonable then.

PR #210, opened 2017-10-10
Retry for server streaming

chrisdunelm:


src/Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -86,7 +86,10 @@ public ClientHelper(ServiceSettingsBase settings)
             where TResponse : class, IMessage<TResponse>
         {
             CallSettings baseCallSettings = _clientCallSettings.MergedWith(perMethodCallSettings);
+            // These operations are applied in reverse order.
+            // I.e. Version header is added first, then retry is performed.
```
jskeet: It's not really clear to me what this means. I would expect the two things to be basically orthogonal - we're only setting up what can happen when we execute the call, aren't we? If the methods were called in the opposite order, what difference would there be?

testing/Google.Api.Gax.Testing/FakeScheduler.cs
```diff
@@ -225,6 +225,11 @@ private Task StartLoopAsync(DateTime simulatedTimeout)
                                 return;
                             }
                         }
+                        // Give some time for tasks to execute.
```
jskeet: So is this the fix that we've wanted for ages but haven't been able to work out, that basically makes FakeScheduler more reliable in general?

I must admit I can't remember enough about it to understand why that loop would work, but I'm sure you're right :)

test/Google.Api.Gax.Grpc.Tests/RetryTest.cs
```diff
@@ -65,27 +60,34 @@ public class RetryTest
             await scheduler.RunAsync(async () =>
             {
                 var callSettings = CallSettings.FromCallTiming(CallTiming.FromRetry(retrySettings));
-                var retryingCallable = callable.WithRetry(scheduler.Clock, scheduler);
-                var result = await Call(async, retryingCallable, new SimpleRequest { Name = name }, callSettings);
+                SimpleResponse result;
+                if (serverStreaming)
+                {
+                    var retryingCallable = server.ServerStreamingCallable.WithRetry(scheduler.Clock, scheduler);
+                    result = await Call(async, retryingCallable, new SimpleRequest { Name = name }, callSettings);
```
jskeet: Presumably the different types involved here mean we can't factor out the common code?

test/Google.Api.Gax.Grpc.Tests/RetryTest.cs
```diff
@@ -96,30 +98,38 @@ public class RetryTest
             await scheduler.RunAsync(async () =>
             {
                 var callSettings = CallSettings.FromCallTiming(CallTiming.FromRetry(retrySettings));
-                var retryingCallable = callable.WithRetry(scheduler.Clock, scheduler);
-                var result = await Call(async, retryingCallable, new SimpleRequest { Name = name }, callSettings);
+                SimpleResponse result;
```
jskeet: Given that we've got this if/else several times, perhaps move it to a common method?

src/Google.Api.Gax.Grpc/ApiCallRetryExtensions.cs
```diff
@@ -57,7 +62,7 @@ internal static class ApiCallRetryExtensions
         // Sync retry
         internal static Func<TRequest, CallSettings, TResponse> WithRetry<TRequest, TResponse>(
             this Func<TRequest, CallSettings, TResponse> fn,
-            IClock clock, IScheduler scheduler) =>
+            IClock clock, IScheduler scheduler, Action<TResponse> postResponse) =>
```
jskeet: For added flexibility, we could make this a `Func<TResponse, TResponse>` - but if everything's internal, we could do that later if we need to?

chrisdunelm: I also considered making this a Func, but as you say, it is all internal, so can be changed later if required.

src/Google.Api.Gax.Grpc/ClientHelper.cs
chrisdunelm: You're right, these are orthagonal, and changing the order would make no difference.
The comment is there just to be consistent with the `BuildApiCall` method above.

testing/Google.Api.Gax.Testing/FakeScheduler.cs
chrisdunelm: I'm not sure if this is *the* one true fix; but it certainly was required to make some of the new tests pass :)

test/Google.Api.Gax.Grpc.Tests/RetryTest.cs
chrisdunelm: OK, yes, I'll see about doing a small refactor.

test/Google.Api.Gax.Grpc.Tests/RetryTest.cs
chrisdunelm: They are indeed different types; but I'll see about doing a small refactor.

chrisdunelm:
Refactor done. Will merge on green if that's ok

PR #211, opened 2017-10-11
Add CallSettings.Empty singleton

chrisdunelm:


test/Google.Api.Gax.Grpc.Tests/CallSettingsTest.cs
```diff
@@ -95,6 +95,28 @@ public void ToCallOptions_All()
...
+            Assert.Null(empty.PropagationToken);
+            Assert.Null(empty.HeaderMutation);
+            Assert.Null(empty.Timing);
+            Assert.Null(empty.WriteOptions);
+        }
+
+        [Fact]
+        public void Empty_Merge()
+        {
+            var overlay = CallSettings.FromCancellationToken(CancellationToken.None);
```
jskeet: I'd hoped that this *would* return `CallSettings.Empty`, but of course specifying a cancellation token of none isn't the same as omitting it, as it can override one. Urgh.

Presumably we can have a second singleton which is for CancellationToken.None?

chrisdunelm:
Was this what you had in mind?

jskeet:
Yup, that's much more like it. That won't help as much on the merging side as I'd hoped, but it should avoid creating a bunch of call settings for no reason.

With that one present, is Empty still useful do you think?

PR #213, opened 2017-10-17
Correct platform support for GKE

### LINKED ISSUE #209: Monitored resource type incorrect for GKE containers
Looking at [the documentation](https://cloud.google.com/logging/docs/api/v2/resource-list) we should be logging with a resource type of "container" rather than "gke_cluster" if we detect we're running in Container Engine.

[This](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax.Grpc/MonitoredResourceBuilder.cs) code should be fixed - but we should check all the other resource types at the same time.

cc @iantalarico

chrisdunelm:
Discovers extra information about GKE, and creates the correct logging type of 'container', rather than 'gke_cluster'

Fixes #209


chrisdunelm:
The GKE platform integration test passes.

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,37 +170,163 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+                return null;
+            }
+            var baseUrl = $"https://{kubernetesServiceHost}:{kubernetesServicePort}/api/v1";
+            string kubernetesNamespace;
+            string kubernetesToken;
+            string kubernetesCa;
+            string[] mountInfo;
+            try
+            {
+                kubernetesNamespace = File.ReadAllText("/var/run/secrets/kubernetes.io/serviceaccount/namespace");
```
jskeet: Presumably these will all fail if GKE starts running Kubernetes on Windows containers?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,37 +170,163 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+            if (string.IsNullOrEmpty(kubernetesNamespace) || string.IsNullOrEmpty(kubernetesToken) ||
+                string.IsNullOrEmpty(kubernetesCa) || mountInfo == null || mountInfo.Length == 0)
+            {
+                // These files should contain useful data on GKE
+                return null;
+            }
+
+            var handler = new HttpClientHandler
+            {
+                // TODO: Properly check that the certificate is valid, using kubernetesCa
```
jskeet: This sounds like an important TODO - are we happy to ship a release without it?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,37 +170,163 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+                {
+                    return null;
+                }
+            }
+        }
+#elif NET45
+        internal static Task<KubernetesData> LoadKubernetesDataAsync()
+        {
+            // .NET45 code cannot run on Kubernetes.
+            // And .NET45 does not cert functionality that we require.
```
jskeet: "does not cert functionality"?

jskeet: (I do wonder what would happen to code running on Mono in GKE...)

src/Google.Api.Gax/Platform.cs
```diff
@@ -210,13 +339,38 @@ public static GkePlatformDetails TryLoad(string metadataJson)
         /// <param name="clusterName">The cluster name.</param>
         /// <param name="location">The location.</param>
         /// <param name="hostName">The instance host name.</param>
+        [Obsolete("Only partial fills instance with data; use alternative constructor.")]
```
jskeet: Partially?

src/Google.Api.Gax/Platform.cs
```diff
@@ -338,7 +518,8 @@ private static GaePlatformDetails LoadGaeDetails()
             var metadataJson = await LoadMetadataAsync().ConfigureAwait(false);
             if (metadataJson != null)
             {
-                GkePlatformDetails gkeDetails = GkePlatformDetails.TryLoad(metadataJson);
+                var kubernetesData = await GkePlatformDetails.LoadKubernetesDataAsync().ConfigureAwait(false);
+                GkePlatformDetails gkeDetails = GkePlatformDetails.TryLoad(metadataJson, kubernetesData);
```
jskeet: Check for kubernetesData being null before calling TryLoad?

test/Google.Api.Gax.PlatformIntegrationTests/Startup.cs
```diff
@@ -36,14 +36,26 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
-                string output = envs.Cast<DictionaryEntry>()
-                    .Select(x => new { key = x.Key.ToString(), value = x.Value.ToString() })
-                    .OrderBy(x => x.key)
-                    .Aggregate($"{platform.Type}\n{platform}\n\n", (acc, x) => acc + $"[{x.key}]: '{x.value}'\n");
-                await context.Response.WriteAsync(output);
+                try
+                {
+                    var platform = await Google.Api.Gax.Platform.InstanceAsync();
+                    var envs = Environment.GetEnvironmentVariables();
+                    // Output a the detected platform, the all environment variables.
```
jskeet: Now would be a good time to make the comment make more sense :)

test/Google.Api.Gax.Tests/Resources/Metadata.json
```diff
@@ -0,0 +1,80 @@
+{
```
jskeet: Anything sensitive in here?

jskeet: (The ssh keys maybe?)

src/Google.Api.Gax/Platform.cs
chrisdunelm: done

test/Google.Api.Gax.Tests/Resources/Metadata.json
chrisdunelm: done

chrisdunelm:
PTAL

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
```diff
@@ -73,17 +73,45 @@ public void GkePlatform()
...
+                NamespaceJson = namespaceJson,
+                PodJson = podJson,
+                MountInfo = new[] { "/var/lib/kubelet/pods/podid/containers/containername/ /dev/termination-log" }
+            };
+            var platform = new Platform(GkePlatformDetails.TryLoad(metadataJson, kubernetesData));
             var resource = MonitoredResourceBuilder.FromPlatform(platform);
-            Assert.Equal("gke_cluster", resource.Type);
-            Assert.Equal(3, resource.Labels.Count);
+            Assert.Equal("container", resource.Type);
+            Assert.Equal(7, resource.Labels.Count);
```
jskeet: Possibly build an expected dictionary and just `Assert.Equal(expectedLabels, resource.Labels)`?

src/Google.Api.Gax.Grpc/MonitoredResourceBuilder.cs
```diff
@@ -86,12 +86,16 @@ public static MonitoredResource FromPlatform(Platform platform)
...
                     return new MonitoredResource
                     {
-                        Type = "gke_cluster",
+                        Type = "container",
                         Labels =
                         {
                             { "project_id", gke.ProjectId },
                             { "cluster_name", gke.ClusterName },
-                            { "location", gke.Location }
+                            { "namespace_id", gke.NamespaceId },
```
jskeet: Are we okay to produce empty labels rather than only populating the known ones?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
 
+#if NETSTANDARD1_3
+        internal static async Task<KubernetesData> LoadKubernetesDataAsync()
+        {
+            var kubernetesServiceHost = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST");
+            int.TryParse(Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_PORT"), out var kubernetesServicePort);
+            var podName = Environment.GetEnvironmentVariable("HOSTNAME");
+            if (string.IsNullOrEmpty(kubernetesServiceHost) || kubernetesServicePort == 0 || string.IsNullOrEmpty(podName))
+            {
+                // Not running on GKE
```
jskeet: Not running on Kubernetes?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+                return null;
+            }
+            // If anything following fails, then the return value will still show that we're running on GKE.
```
jskeet: Ditto - unless that really is GKE-specific.

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+            try
+            {
+                kubernetesNamespace = File.ReadAllText("/var/run/secrets/kubernetes.io/serviceaccount/namespace");
+                kubernetesToken = File.ReadAllText("/var/run/secrets/kubernetes.io/serviceaccount/token");
+                kubernetesCaCert = new X509Certificate2("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt");
+                mountInfo = File.ReadAllLines("/proc/self/mountinfo");
+            }
+            catch
+            {
+                // Ignore all errors
```
jskeet: Maybe expand this to say "Because we'll return with partial data below anyway"?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
+            {
+                ServerCertificateCustomValidationCallback = (msg, cert, chain, errs) =>
+                {
+                    // Add the kubernetes-provided ca
+                    chain.ChainPolicy.ExtraStore.Add(kubernetesCaCert);
+                    chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+                    // Allow ca that is not installed on the current machine
+                    // This allows *any* ca to be used, hence the following check for the kubernetes-provided ca
+                    chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
+                    if (chain.Build(cert))
```
jskeet: How about

```csharp
// Useful comment here
return chain.Build(cert) && chain.ChainElements[chain.ChainElements.Count - 1].Certificate.Thumbprint == kubernetesCaCert.Thumbprint;
```

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
             {
                 metadata = JObject.Parse(metadataJson);
+                namespaceData = JObject.Parse(kubernetesData.NamespaceJson);
+                podData = JObject.Parse(kubernetesData.PodJson);
             }
             catch
             {
+                // Ignore all errors
+            }
+            if (metadata == null ||
```
jskeet: What happens if this gets JSON without a "kind" value? Maybe use `?` liberally in the code below?

src/Google.Api.Gax/Platform.cs
```diff
@@ -167,56 +173,238 @@ public GaePlatformDetails(string gcloudProject, string gaeInstance, string gaeSe
...
-            JObject metadata;
+            GaxPreconditions.CheckNotNull(metadataJson, nameof(metadataJson));
+            GaxPreconditions.CheckNotNull(kubernetesData, nameof(kubernetesData));
+            JObject metadata = null;
+            JObject namespaceData = null;
+            JObject podData = null;
             try
             {
                 metadata = JObject.Parse(metadataJson);
+                namespaceData = JObject.Parse(kubernetesData.NamespaceJson);
```
jskeet: What happens if PodJson is null? Does JObject.Parse just return null? (I'm trying to think about all the possible partial-information scenarios.)

src/Google.Api.Gax/Platform.cs
chrisdunelm: done

src/Google.Api.Gax/Platform.cs
chrisdunelm: Refactored slightly. It now doesn't matter if Parse returns null or throws, the behaviour will be the same.

src/Google.Api.Gax/Platform.cs
chrisdunelm: Done. All cases of missing/incomplete data are now handled.

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
chrisdunelm: done

chrisdunelm:
PTAL

src/Google.Api.Gax/Platform.cs
```diff
@@ -320,23 +316,26 @@ public static GkePlatformDetails TryLoad(string metadataJson, KubernetesData kub
...
+            if (metadata == null)
             {
-                metadata = JObject.Parse(metadataJson);
-                namespaceData = JObject.Parse(kubernetesData.NamespaceJson);
-                podData = JObject.Parse(kubernetesData.PodJson);
+                // Metadata is required. If it's not present, or the JSON cannot be parsed, return null.
+                return null;
             }
-            catch
+            if (namespaceData != null && namespaceData["kind"].Value<string>() != "Namespace")
```
jskeet: `?.Value`? Ditto for `podData` check below?

test/Google.Api.Gax.Tests/PlatformTest.cs
```diff
@@ -104,5 +104,65 @@ public void Gke_WithData(bool withPodName)
...
         }
+
+        const string metadataValid = "{'project':{'projectId':'FakeProjectId'},'instance':{'attributes':{'cluster-name':'FakeClusterName'},'id':'123','zone':'projects/FakeProject/zones/FakeLocation'}}";
+        const string metadataIncomplete = "{'project':{'projectId':'FakeProjectId'},'instance':{'attr";
+        const string namespaceValid = "{'kind':'Namespace','metadata':{'uid':'namespaceid'}}";
+        const string namespaceWrongKind = "{'kind':'NotNamespace','metadata':{'uid':'namespaceid'}}";
+        const string namespaceIncomplete = "{'kind':'Namespace','m";
+        const string podValid = "{'kind':'Pod','metadata':{'name':'podname','uid':'podid'}}";
+        const string podWrongKind = "{'kind':'NotPod','metadata':{'name':'podname','uid':'podid'}}";
+        const string podIncomplete = "{'kind':'Pod','metadata':{";
```
jskeet:
```csharp
const string podNoKind = "{'unkind': 'Pod'}";
```
(and ditto for `namespace`) to test my earlier objection.

src/Google.Api.Gax/Platform.cs
chrisdunelm: Oops, yes done.

test/Google.Api.Gax.Tests/PlatformTest.cs
chrisdunelm: Done

PR #222, opened 2017-11-17
Update release version

jskeet:
There isn't anything else that needs doing, is there?


jskeet:
(I've rebased to master and checked that we haven't got that rogue auth 1.7.1 dependency...)

chrisdunelm:
Fairly sure that's all that needs doing.

PR #229, opened 2017-12-14
Primitive JSON builder

jskeet:
This provides very crude functionality, but is expected to be
sufficient for our use cases.


jskeet:
(It's actually pretty pleasant to use for simple objects...)

PR #237, opened 2018-02-23
Check for null on all ApiCall* ctor params

chrisdunelm:


jskeet:
Is this definitely appropriate? Normally we expect a null CallSettings to be just "don't care" - is the idea that there have to be call settings *somewhere* for timing purposes?

chrisdunelm:
You're right. I was too enthusiastic in my preconditioning. Fixed now :)

PR #238, opened 2018-03-07
Update common protos

jskeet:


jskeet:
Adding the version bump to 1.2.0 - might as well do it all in one go...

PR #241, opened 2018-03-08
Expose a mutable version header builder in ServiceSettingsBase

jskeet:
This allows clients to add their own library versions.

Additionally, VersionHeaderBuilder now ensures that names are unique,
and that neither names nor versions include " " or "/", both of
which are used as separators.


jskeet:
Ping for review? I think we probably do want this.

PR #214, opened 2017-10-23
Exception customization support

jskeet:
(Prototype updated.)

PR #242, opened 2018-03-15
Make response metadata accessible via a callback.

jskeet:
First commit is from #214 - skip that. I won't merge this PR as-is, but I'd like feedback on the second commit.

### DIFF
--- a/src/Google.Api.Gax.Grpc/ApiCall.cs
+++ b/src/Google.Api.Gax.Grpc/ApiCall.cs
@@ -22,10 +22,75 @@ namespace Google.Api.Gax.Grpc
             where TRequest : class, IMessage<TRequest>
             where TResponse : class, IMessage<TResponse>
         {
-            return new ApiCall<TRequest, TResponse>(
-                (req, cs) => asyncGrpcCall(req, cs.ToCallOptions(clock)).ResponseAsync,
-                (req, cs) => syncGrpcCall(req, cs.ToCallOptions(clock)),
-                baseCallSettings);
+            var adapter = new GrpcCallAdapter<TRequest, TResponse>(asyncGrpcCall, syncGrpcCall, clock);
+            return new ApiCall<TRequest, TResponse>(adapter.CallAsync, adapter.CallSync, baseCallSettings);
+        }
+
+        /// <summary>
+        /// Adapter used to mask the fact that when we need response/trailing metadata, a sync call may need
+        /// to use the async gRPC code.
+        /// </summary>
+        private class GrpcCallAdapter<TRequest, TResponse>
+            where TRequest : class, IMessage<TRequest>
+            where TResponse : class, IMessage<TResponse>
+        {
+            private readonly Func<TRequest, CallOptions, AsyncUnaryCall<TResponse>> _asyncGrpcCall;
+            private readonly Func<TRequest, CallOptions, TResponse> _syncGrpcCall;
+            private readonly IClock _clock;
+
+            internal GrpcCallAdapter(
+                Func<TRequest, CallOptions, AsyncUnaryCall<TResponse>> asyncGrpcCall,
+                Func<TRequest, CallOptions, TResponse> syncGrpcCall,
+                IClock clock)
+            {
+                _asyncGrpcCall = asyncGrpcCall;
+                _syncGrpcCall = syncGrpcCall;
+                _clock = clock;
+            }
+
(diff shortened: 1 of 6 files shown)


src/Google.Api.Gax.Grpc/ApiCall.cs
```diff
@@ -33,10 +33,65 @@ internal static class ApiCall
...
+                // If we don't have any response/trailing metadata handlers, it's simplest just to
+                // return the gRPC task.
+                if (responseMetadataHandler == null && trailingMetadataHandler == null)
+                {
+                    return unaryCall.ResponseAsync;
+                }
+
+                // Otherwise, we want to pass the returned metadata to the handlers before our returned task completes,
+                // so we create a proxy task.
+                return unaryCall.ResponseAsync.ContinueWith(task =>
```
chrisdunelm: I think I'd slightly prefer this done as an inner async function, rather than using `ContinueWith`.

src/Google.Api.Gax.Grpc/CallSettings.cs
```diff
@@ -35,14 +35,39 @@ public sealed class CallSettings
...
+        /// <param name="trailingMetadataHandler">Action to invoke when trailing metadata is received.</param>
+        public CallSettings(
+            CancellationToken? cancellationToken,
+            CallCredentials credentials,
+            CallTiming timing,
+            Action<Metadata> headerMutation,
+            WriteOptions writeOptions,
+            ContextPropagationToken propagationToken,
+            Action<Metadata> responseMetadataHandler,
+            Action<Metadata> trailingMetadataHandler)
```
chrisdunelm: We should consider allowing async handlers: `Func<Metadata, Task> ...Handler`

src/Google.Api.Gax.Grpc/ApiCall.cs
```diff
@@ -33,10 +33,65 @@ internal static class ApiCall
...
+                {
+                    var responseHeadersTask = unaryCall.ResponseHeadersAsync;
+                    if (responseHeadersTask.Status == TaskStatus.RanToCompletion)
+                    {
+                        responseMetadataHandler?.Invoke(responseHeadersTask.Result);
+                    }
+                    trailingMetadataHandler?.Invoke(unaryCall.GetTrailers());
```
chrisdunelm: Handlers are called synchronously, so they have to complete before the RPC is complete.
I agree this is the simplest thing to do.

src/Google.Api.Gax.Grpc/ApiCall.cs
jskeet: Do you mean by calling an async method that awaits the task first? Yes, that's doable.

src/Google.Api.Gax.Grpc/CallSettings.cs
jskeet: Hmm, possibly. That does make it all more complex - and I'd argue that you almost certainly don't want to be doing significant work in the handler anyway.

jskeet:
This PR is now a single commit which *just* addresses the metadata feature, so we can merge it independently of #214.

jskeet:
Just noticed I don't have tests for call settings - adding those now; probably worth holding off on the review for now.

jskeet:
Okay, tests for everything now - PTAL.

test/Google.Api.Gax.Grpc.Tests/CallSettingsExtensionsTest.cs
```diff
@@ -337,5 +337,53 @@ public void WithExpiration_SettingsWithRetry()
...
         }
+
+        [Fact]
+        public void WithResponseMetadataHandler_OnlyHandler()
+        {
+            Metadata setByHandler = null;
+            var settings = ((CallSettings) null).WithResponseMetadataHandler(m => setByHandler = m);
+            var metadata = new Metadata();
+            settings.ResponseMetadataHandler(metadata);
+            Assert.Same(setByHandler, metadata);
```
chrisdunelm: I think the args should be swapped. `metadata` is the expected value? ([Assert.Same()](https://github.com/xunit/assert.xunit/blob/bbece92b18661cc18ede9d72b28648466f9a1709/IdentityAsserts.cs#L32))

jskeet:
Whoops - thanks for the careful review on the tests. Will fix.

PR #243, opened 2018-03-20
Propagate handlers in CallSettings extension methods

jskeet:
I've validated manually that these are the last calls which misuse
the shorter constructor. Thoughts about how to avoid similar
problems in the future would be welcome.


jskeet:
Odd. The test is passing locally - I've restarted the Travis build, and we'll see what happens. The fake scheduler is complex so maybe there was some oddity there...

jskeet:
And it passed... merging this PR as unrelated, but let's keep an eye to see if it happens again.

PR #247, opened 2018-04-12
Add common resource names

jskeet:
These are resource names used in multiple Google Cloud APIs. Having
a single type for each resource name makes it simpler to write use
the output of one API as the input to another.

These classes are the existing generated ones, with a modified comment (and namespace aliasing removed).

### DIFF
--- a/src/Google.Api.Gax/ResourceNames/ProjectName.cs
+++ b/src/Google.Api.Gax/ResourceNames/ProjectName.cs
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2018 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+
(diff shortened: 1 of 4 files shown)


src/Google.Api.Gax/ResourceNames/ProjectName.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <inheritdoc />
+        public bool Equals(ProjectName other) => ToString() == other?.ToString();
+
+        /// <inheritdoc />
+        public static bool operator ==(ProjectName a, ProjectName b) => ReferenceEquals(a, b) || (a?.Equals(b) ?? false);
+
+        /// <inheritdoc />
+        public static bool operator !=(ProjectName a, ProjectName b) => !(a == b);
+    }
+}
```
chrisdunelm: super-nitty-nit: Add blank line at EOF

jskeet: Done.

PR #255, opened 2018-05-31
Introduce a gRPC interceptor into settings

jskeet:


jskeet:
I've removed a commit which would have created 2.5.0-beta01; it's easy to test this without it.

PR #260, opened 2018-07-10
Xamarin.Android: add support

ahouben:
This PR allows the nuget package Google.Cloud.Speech.V1 to be used on the Xamarin.Android platform.

It relates to PR https://github.com/grpc/grpc/pull/15969

### DIFF
--- a/src/Google.Api.Gax/Google.Api.Gax.csproj
+++ b/src/Google.Api.Gax/Google.Api.Gax.csproj
@@ -10,6 +10,7 @@
   <PropertyGroup>
     <Title>Google API Extensions</Title>
     <Description>Support classes for Google API client libraries</Description>
+    <Version>2.4.1</Version>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Google.Api.Gax/VersionHeaderBuilder.cs
+++ b/src/Google.Api.Gax/VersionHeaderBuilder.cs
@@ -9,6 +9,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.InteropServices;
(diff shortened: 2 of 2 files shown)


jskeet:
I was going to merge this with another commit to remove the project file change, but now I can't see why it's necessary.

`FormatVersion` explicitly handles null input, returning "" - the problem is just `AppendVersion` requiring a non-empty string. I'll look into how we actually want to handle unknown versions, and fix it either by returning a non-empty string from `FormatVersion` if it's passed null input, or by removing the constraint in `AppendVersion`.

Note that even when all of this works, we won't be ready to *support* Xamarin for Google Cloud Client Libraries. They may work in many cases, but they won't be supported until we've done more testing, documented an authentication strategy etc.

jskeet:
Closing this as #262 has now been merged.

ahouben:
Your alternative looks good too, thanks for following up.

ahouben:
Please note that in addition to the changes in #262 you still need to set the following to a valid path string in your main application code on Xamarin.Android as a workaround before using Google.Api.Gax to prevent an exception at https://github.com/aspnet/PlatformAbstractions/blob/rel/1.1.0/src/Microsoft.Extensions.PlatformAbstractions/ApplicationEnvironment.cs#L46 during initialization of field ApplicationEnvironment.ApplicationBasePath (since Path.GetFullPath(basePath), where basePath is null or "", throws):

```
AppDomain.CurrentDomain.SetData("APP_CONTEXT_BASE_DIRECTORY", Path.DirectorySeparatorChar.ToString());
```

See here https://github.com/Greenliff/grpc-xamarin-samples/blob/master/GrpcXamarinAndroidDemo/GrpcXamarinAndroidDemo/MainActivity.cs#L80.

PR #261, opened 2018-07-10
Fix the default for unknown versions to "0.0.0"

jskeet:
Previously when creating a version header for a type/framework that
didn't have a version, we'd default to a value of ""... which would
then be rejected by AppendVersion. 0.0.0 is a reasonable semantic
version for "unknown".

This is an alternative to #260.


jskeet:
(Unfortunately this is hard to test without creating an assembly with no version...)

jskeet:
Abandoning this PR as the better fix is to make empty string valid.

PR #266, opened 2018-07-13
Update to xunit v2.3.1 and Microsoft.NET.Test.Sdk v15.8.0

GR3G0R:
This is a follow up to #259 we will be submitting the Google.type.money PR once this is merged.

Issue #259, opened 2018-06-08
Google.Type.Money <=> decimal


nycdotnet:
Greetings,

I am putting together a PR to add an extension method on `decimal` to convert to `Google.Type.Money` with an optional currency code, and to add a `ToDecimal()` extension on `Google.Type.Money` that omits the currency type.  The pattern used will be similar to #251 and #256.

I have a few questions that I'd love to resolve ahead of time.

1.  Are you ok with updating xunit to the latest stable 2.3.1?  This solution currently uses a beta of 2.3.0 which we find is a bit buggy, particularly with regards to the VS runner.
1.  I am currently using `public static Money ToMoney(this decimal value, string currencyCode = "")` as the signature of the `ToMoney` extension method.  Is this acceptable to you?  Would it be better to use `default` with `currencyCode` which would require upgrading the project from C# 7 to C# 7.1?
1.  What do you think the behavior of ToMoney() should be in cases where the decimal value does not fit?  (see below).

I am thinking that if the integral part of the decimal doesn't fit in the `Units` field (long), the `OverflowException` should be caught and used as the `InnerException` of an `ArgumentOutOfRangeException`.  Would love your thoughts on if there is value in doing this, or if we should just not catch the `OverflowException`.

What should happen when the Nanos don't fit is less clear, however.  For example, should `0.555_555_555_5M` be truncated to `555_555_555` nanos?  rounded to `555_555_556` nanos?  Should this throw?  Should the user be able to specify using some sort of enum with the default being to truncate since this sort of thing is only relevant to a very small number of people who will likely want to specify the behavior?

Would love your thoughts on the above.  Thanks!

jskeet:
Hi Steve,

I'm currently flying back from a conference, and will be at more conferences next week. I'll try to give this some thought, but it may be a few days. Hope that's okay!

jskeet:
1. I'm fine with the xunit update

2. Using the default literal wouldn't work, as the default for string is null, which would immediately fail validation. I'm not sure that it's a good idea to have a default at all, to be honest. If you're creating a Money value, you *should* provide the currency. I'd expect anything receiving a value to want there to be a currency code. (Validating that it's a suitable ISO 4217 might be a bit far, admittedly.)

[...]

PR #267, opened 2018-07-17
Add Google.Type.Money to gax-dotnet

GR3G0R:


GR3G0R:
@jskeet, as indicated in #266 we are opening a PR for Google.Type.Money. Please let us know if you have need of additional support from our end. Thank you!

src/Google.Api.CommonProtos/Type/MoneyPartial.cs
```diff
@@ -0,0 +1,23 @@
...
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+
+namespace Google.Type
+{
+    public partial class Money
+    {
+        /// <summary>
```
jskeet: I wonder whether a `DecimalValue` property would be useful here.

You could then move the `ToMoney` extension method into the setter, and avoid the extension at all. Creating a `Money` from a `decimal` would just be:

```csharp
Money money = new Money { DecimalValue = value, CurrencyCode = currency };
```

That doesn't chain as well, but I think I prefer it over the extension method. Thoughts?

src/Google.Api.CommonProtos/Type/MoneyExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        /// Converts a <see cref="decimal"/> to <see cref="Money"/>.
+        /// </summary>
+        /// <param name="value">The amount of money.</param>
+        /// <param name="currencyCode">The 3-letter currency code defined in ISO 4217.</param>
+        /// <returns>A <see cref="Money"/> containing the decimal value in the specified currency</returns>
+        /// <exception cref="ArgumentOutOfRangeException">The integral part of the decimal must be a valid <see cref="Int64"/>, and the fractional part must have a maximum of 9 digits of precision.</exception>
+        public static Money ToMoney(this decimal value, string currencyCode)
+        {
+            decimal integralPart = Math.Truncate(value);
```
jskeet: We should validate that `currencyCode` isn't null.
We *could* also validate that it's not empty.
(None of this is relevant if we move to a property instead.)

src/Google.Api.CommonProtos/Type/MoneyPartial.cs
nycdotnet: Thanks for reviewing this.  I am somewhat partial to the extension method as it means significantly less code for the consumer of this library in each of the call sites and I think it's a bit more discoverable/obvious, but I do like the idea of the `DecimalValue` property.  Would you want to do both?  I will discuss this with @GR3G0R by tomorrow.

If we do not go the property route, a null `currencyCode` will already throw via the auto-generated code (though it will do some arithmetic first that will be wasted).  I prefer to allow `string.Empty` currency codes in anything that is implemented for this as this empty strings are permitted by the proto.

Thank you again!!

src/Google.Api.CommonProtos/Type/MoneyExtensions.cs
GR3G0R: @jskeet, I will work on implementing your suggested modifications and try to push them next week.

GR3G0R:
@jskeet, after having consulted with @nycdotnet, I went ahead and implemented your suggested modifications. Please let us know if these changes are sufficient or if you find that further amendments are warranted. Thank you!

src/Google.Api.CommonProtos/Type/MoneyPartial.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// The amount of money in <see cref="decimal"/> format.
+        /// </summary>
+        /// <exception cref="ArgumentOutOfRangeException">The integral part of the decimal must be a valid <see cref="Int64"/>, and the fractional part must have a maximum of 9 digits of precision.</exception>
+        public decimal DecimalValue {
```
jskeet: Nit: please use "brace at start of line" style for manual code.
(I know protoc uses "brace at end of line" but that's more historical than anything else.)

src/Google.Api.CommonProtos/Type/MoneyPartial.cs
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+using System;
+
+namespace Google.Type
+{
+    public partial class Money
+    {
+        /// <summary>
+        /// The amount of money in <see cref="decimal"/> format.
```
jskeet: It would be worth explaining that this is a view over `Units` and `Nanos`: getting this property combines those property values and setting this property will set both of those properties.

test/Google.Api.CommonProtos.Tests/Type/MoneyTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            { 0.000_000_000_1M, "digits after the decimal" },
+            { -0.000_000_000_1M, "digits after the decimal" },
+            { 9_223_372_036_854_775_808.000_000_000_1M, "integral part" },
+            { -9_223_372_036_854_775_809.000_000_000_1M, "integral part" },
+        };
+
+        [Theory, MemberData(nameof(InvalidMoneyValues))]
+        public void OutOfRangeValueThrowsArgumentOutOfRangeException(decimal valueToConvert, string expectedMessage)
+        {
+            bool caughtArgumentOutOfRangeException = false;
```
jskeet: It would be simpler to write this as:

```csharp
var exception = Assert.Throws<ArgumentOutOfRangeException>(() => new Money { DecimalValue = valueToConvert });
Assert.Contains(expectedMessage, ex.Message);
```

GR3G0R:
@jskeet , please see c4d6ac9.

src/Google.Api.CommonProtos/Type/MoneyPartial.cs
```diff
@@ -12,14 +12,17 @@ namespace Google.Type
     public partial class Money
     {
         /// <summary>
-        /// The amount of money in <see cref="decimal"/> format.
+        /// The amount of money in <see cref="decimal"/> format. This is an abstraction of the Units and Nanos properties.
```
jskeet: Typo: abastraction => abstraction. (I can fix it later if you'd prefer.) Otherwise, feel free to commit --amend and force push.

nycdotnet: Should be fixed in the latest commit.  Thanks Jon!

jskeet:
Great, thanks - merging now.

GR3G0R:
Thanks @jskeet!

nycdotnet:
Hi @jskeet we're updating our dependencies and it seems this hasn't been released yet ( `DecimalValue` feature on `Google.Type.Money`).  The latest version of Google.Api.CommonProtos on NuGet is still 1.3 from June 4.  I know I explicitly said there is no urgency in #259, but it's been a few months.  Any chance we can get this new version released?

Our sincere thanks for all you do!!

jskeet:
Will try to do this next week - thanks for the reminder :)

nycdotnet:
Amazing - thanks.

GR3G0R:
Thanks @jskeet!

jskeet:
Now released in Google.Api.CommonProtos 1.4.0 :)

nycdotnet:
Thanks so much.

PR #269, opened 2018-07-19
Changin zone, podid and namespaceid for MonitoredResources

amanda-tarafa:
Changing zone value for GKE and GCE, and podid and namespaceid values for GKE MonitoredResources build from Platform information.

I've added some comments in the code to be reviewed with questions I have.

### DIFF
--- a/src/Google.Api.Gax/Platform.cs
+++ b/src/Google.Api.Gax/Platform.cs
@@ -7,7 +7,6 @@
 
 using Newtonsoft.Json.Linq;
 using System;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net;
@@ -87,7 +86,12 @@ namespace Google.Api.Gax
         /// <param name="metadataJson">The full JSON string retrieved from the metadata server. Must not be <c>null</c>.</param>
         /// <param name="projectId">The project ID. Must not be <c>null</c>.</param>
         /// <param name="instanceId">The instance ID. Must not be <c>null</c>.</param>
-        /// <param name="zoneName">The zone name. Must not be <c>null</c>.</param>
+        /// <param name="zoneName">The zone name. Must not be <c>null</c>.
+        /// If this value is in the format <code>projects/&lt;project-number&gt;/zones/&lt;zone-name&gt;</code>
+        /// then <see cref="Location"/> will return the <code>&lt;zone-name&gt;</code> part of this value.
+        /// If not, <see cref="Location"/> will throw <see cref="InvalidOperationException"/>.
+        /// If this value has been retrived from Google Compute Engine, the it's format will be the one
+        /// described above.</param>
         public GcePlatformDetails(string metadataJson, string projectId, string instanceId, string zoneName)
         {
             MetadataJson = GaxPreconditions.CheckNotNull(metadataJson, nameof(metadataJson));
@@ -112,9 +116,31 @@ namespace Google.Api.Gax
         public string InstanceId { get; }
 
         /// <summary>
-        /// The zone name where this GCE instance is running.
+        /// The zone where this GCE instance is running.
+        /// This will be in the format <code>projects/&lt;project-number&gt;/zones/&lt;zone-name&gt;</code>
+        /// id the value has been retrieved from Google Compute Engine.
         /// </summary>
         public string ZoneName { get; }
+
+        /// <summary>
+        /// The zone name where this GCE instance is running.
+        /// If <see cref="ZoneName"/> is in the format <code>projects/&lt;project-number&gt;/zones/&lt;zone-name&gt;</code>
+        /// this value will be the <code>&lt;zone-name&gt;</code> part in <see cref="ZoneName"/>.
+        /// If <see cref="ZoneName"/> is in a different format then this getting the value of this property will
+        /// throw <see cref="InvalidOperationException"/>.
+        /// </summary>
+        public string Location
+        {
+            get
+            {
+                if (Platform.s_zoneTemplate.TryParseName(ZoneName, out var zoneResourceName))
+                {
+                    return zoneResourceName[1];
+                }
+                throw new InvalidOperationException($@"For {nameof(Location)} to have a value the format of {nameof(ZoneName)} 
+should be projects/<project_number>/zones/<zone_name>. {nameof(ZoneName)} current value is {ZoneName}.");
+            }
+        }
 
         /// <inheritdoc/>
         public override string ToString() =>
@@ -198,8 +224,6 @@ namespace Google.Api.Gax
             /// </summary>
             public string[] MountInfo { get; set; }
         }
-
-        private static readonly PathTemplate s_zoneTemplate = new PathTemplate("projects/*/zones/*");
 
 #if NETSTANDARD1_3
         internal static async Task<KubernetesData> LoadKubernetesDataAsync()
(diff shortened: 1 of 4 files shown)


src/Google.Api.Gax/Platform.cs
```diff
@@ -112,10 +111,28 @@ public GcePlatformDetails(string metadataJson, string projectId, string instance
         public string InstanceId { get; }
 
         /// <summary>
-        /// The zone name where this GCE instance is running.
+        /// The zone where this GCE instance is running.
+        /// This is in the format <code>projects/&lt;project-number&gt;/zones/&lt;zone-name&gt;</code>.
```
amanda-tarafa: When we build an instance of this class we use the TryLoad method which receives the metadata server info, so, as long as the metadata server returns project/121231231/zones/<zone_name> this comment will hold true.
The problem is that the class has a public constructor that clients might be using to pass whatever zone value they want and in that case we cannot obtain a location. Should we:

- Enforce that the value passed is project/234324/zones/<zone_name> (Overkill and more breaking still)
- If we cannot parse zone, then return whatever zone value is in Location?

src/Google.Api.Gax/Platform.cs
```diff
@@ -112,10 +111,28 @@ public GcePlatformDetails(string metadataJson, string projectId, string instance
...
+        /// </summary>
+        public string Location
+        {
+            get
+            {
+                if (Platform.s_zoneTemplate.TryParseName(ZoneName, out var zoneResourceName))
+                {
+                    return zoneResourceName[1];
+                }
+                return null;
```
amanda-tarafa: `return ZoneName` instead of `null` here?

src/Google.Api.Gax/Platform.cs
```diff
@@ -112,10 +111,28 @@ public GcePlatformDetails(string metadataJson, string projectId, string instance
...
         public string ZoneName { get; }
 
+        /// <summary>
+        /// The zone name where this GCE instance is running.
+        /// This is equivalent to the value of the <code>&lt;zone-name&gt;</code> part in
+        /// <see cref="ZoneName"/>
+        /// </summary>
+        public string Location
+        {
+            get
```
amanda-tarafa: I didn't want to receive a value for Location in the constructor because it is public and don't want to change it. If we overload the constructor to receive Location, we still need to do something when the current constructor is used, like maybe `return ZoneName`?
I think overloading the constructor might be overkill, so maybe accept any ZoneName value, if we can parse it we return the <zone_name> part else we return ZoneName as is.

src/Google.Api.Gax/Platform.cs
```diff
@@ -426,6 +442,8 @@ public GkePlatformDetails(string metadataJson, string projectId, string clusterN
 
         /// <summary>
         /// The cluster location, which is visible in the Google Cloud Platform Console.
+        /// This is equivalent to the value of the <code>&lt;zone-name&gt;</code> part in
+        /// <see cref="Zone"/>
```
amanda-tarafa: We don't have the same problem here than we do with `GcePlatformDetails` because `Location` already existed and was being filled with the `<zone_name>` part of `Zone`.

src/Google.Api.Gax/Platform.cs
jskeet: I think I'll have to take a closer look at the existing code.

src/Google.Api.Gax/Platform.cs
jskeet: You could always rewrite the comment as "When the details have been retrieved from Google Compute Engine, the value is in the format..."

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
```diff
@@ -36,7 +36,7 @@ public void GcePlatform()
   },
   'instance': {
     'id': 'FakeInstanceId',
-    'zone': 'FakeZone'
+    'zone': 'projects/1234567/zones/FakeZone'
```
jskeet: So these changes represent what we *actually* get back from the relevant metadata servers, yes?

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
```diff
@@ -112,11 +113,11 @@ public void GkePlatform()
             {
                 { "project_id", "FakeProjectId" },
                 { "cluster_name", "FakeClusterName" },
-                { "namespace_id", "namespaceid" },
+                { "namespace_id", "namespacename" },
                 { "instance_id", "123" },
-                { "pod_id", "podid" },
+                { "pod_id", "podname" },
```
jskeet: It might be worth adding a comment for why this isn't what one might naively expect it to be.

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
amanda-tarafa: Yes

test/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
amanda-tarafa: OK

src/Google.Api.Gax/Platform.cs
amanda-tarafa: Yes, the comment is not the problem, it was added by me so of course we can have it say whatever. The situation described is independent of the comment.

src/Google.Api.Gax/Platform.cs
jskeet: I think null or throwing an exception is the right approach here. I would be tempted to throw `InvalidOperationException`, personally.

src/Google.Api.Gax/Platform.cs
jskeet: I think this is fine, but the comment for the `zoneName` constructor parameter should describe the now-expected format.

src/Google.Api.Gax/Platform.cs
jskeet: (And we should have a test for it :)

amanda-tarafa: But then the only way a client has of not getting an Exception here is to pass the zone in the format projects/<project_number>/zones/<zone_name> in the constructor. And then, shouldn't we enforce that in the constructor instead of "here" (which would be breaking)?
After all, the Location property is being introduced in this PR, so we can define it: if we can extract the <zone_name> from ZoneName (oh the names) then Location = <zone_name> else Location = ZoneName. I think this is the least breaking change, and admittedly not the cleanest but it is not that bad either.
Anyway, happy to go the exception way if you still prefer it.

amanda-tarafa: (agreed with the test, I'll add them once we have decided)

src/Google.Api.Gax/Platform.cs
amanda-tarafa: OK

src/Google.Api.Gax/Platform.cs
jskeet: I think it's reasonable to define this property as only being available if the zone is in the right format. I don't think we should validate in the constructor - mostly to avoid it being a breaking change, as you say.

My second choice would be to return null to indicate that `ZoneName` is not in a format that we can extract the zone ID from... I don't mind that *much*, although I'd argue that an `InvalidOperationException` with a good message is more useful than a `NullReferenceException` for developers who don't read the docs with enough vigour.

@chrisdunelm thoughts?

amanda-tarafa:
Well, having different values in PodId and NamespaceId in GkePlatformDetails (the names instead of the ids) might also be breaking for some client code. I though of adding to new properties, PodName and NamespaceName to hold the new values and used those for the monitored resource, but then I get into breaking the public constructor issue, which I really think is worse (more breaking) than having different values in these same two fields.

src/Google.Api.Gax/Platform.cs
chrisdunelm: I think I prefer throwing; which should only occur if the user is constructing this class with metadata from a source that isn't GCE. Throwing with a good error message will better help the user to understand what's going on.

chrisdunelm:
Agreed that it's better to not change the ctor, and fixing the existing Id properties.

chrisdunelm:
This PR LGTM :)

src/Google.Api.Gax/Platform.cs
amanda-tarafa: Ok, will throw.

amanda-tarafa:
PTAL

amanda-tarafa:
Thanks!

PR #272, opened 2018-08-02
Remove leftover alias from copy/paste

jskeet:


jskeet:
Those were from google-apis-dotnet-client - a different repo. (I'm on the case.)

PR #274, opened 2018-09-03
Introduce Google.Api.Gax.Grpc.Gcp package

jskeet:
(It's a good job that no-one is ever going to want to type that in
by hand; it's horrible name otherwise! It's likely to be a
dependency from more packages with friendlier names.)

Notes:

- This is the code from #273 just moved around a bit
- We may well want to move some code (EqualityHelpers and
  DefaultChannelCredentialsCache) up the dependency chain later,
  but we can do that later
- I've given it an initial version which reflects a desire for this
  to be "part of the GAX packages" with a common version number; it
  *could* be independent, but it's probably simpler to keep them
  together if we can
- There's a dodgy test-to-test dependency at the moment, but I'd
  rather address that in a subsequent PR

cc @evildour

### DIFF
--- a/src/Google.Api.Gax.Grpc.Gcp/Google.Api.Gax.Grpc.Gcp.csproj
+++ b/src/Google.Api.Gax.Grpc.Gcp/Google.Api.Gax.Grpc.Gcp.csproj
@@ -1,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\CommonProjectProperties.xml" />
+
+  <!-- For the moment, this package is versioned independently. -->
+  <PropertyGroup>
+    <Version>2.5.0-beta01</Version>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard1.5;net45</TargetFrameworks>
+  </PropertyGroup>
+
+  <!-- Packaging information -->
+  <PropertyGroup>
+    <Title>Google Cloud Platform gRPC API Extensions</Title>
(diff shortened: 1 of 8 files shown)


src/Google.Api.Gax.Grpc.Gcp/Google.Api.Gax.Grpc.Gcp.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\CommonProjectProperties.xml" />
+
+  <!-- For the moment, this package is versioned independently. -->
+  <PropertyGroup>
+    <Version>2.5.0-beta01</Version>
```
chrisdunelm: Just checking that you want to start this at v2.5.0? If it's independently versioned, why not start at 1.0.0?

chrisdunelm: beta00?
I assume this isn't going to be released right now?

jskeet: It was deliberate - while it's beta it'll be independent, but I *think* we want to get it in sync with all the rest.

jskeet: I was going to probably do so, unless we have a good reason not to? (We want to release a beta *very soon* anyway, so that we can use it in Bigtable.)

PR #275, opened 2018-09-03
Change the source layout

jskeet:
The src/ test/ and testing/ directories weren't adding any value,
and just made things a bit more complex.

build.sh and generateprotos.sh both work fine.


jskeet:
(We can always fix breakage as we go :)

Issue #276, opened 2018-09-04
Metadata information for platform/monitored resources not populated by default for GKE container projects


tillig:
.NET applications running inside containers in Google Kubernetes Engine pods don't correctly populate the data for "pod ID" and "namespace ID" by default due to permissions on the Kubernetes API.

This is a challenge when using the ASP.NET Core integration with Stackdriver because the log entries don't get attributed to the right namespace and pod - queries don't work.

The ASP.NET Core integration [uses `MonitoredResourceBuilder.FromPlatform()` to populate the metadata for log entries](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs#L87).

The `MonitoredResourceBuilder.FromPlatform()` uses the current environment to get information about the platform. This happens in `Google.Api.Gax.Platform.LoadInstanceAsync()` and there's a [specific section for handling GKE info](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L603).

When the GKE info is loaded, it [gets the pod name from the environment variable `HOSTNAME`](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L233) and the [namespace name from a mounted file in the filesystem](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L247). The pod and namespace are needed to read additional information [from the Kubernetes API](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L285). The important part here is **at this point the values are known**.

When creating the master set of `GkePlatformDetails`, these values get ignored [in favor of parsing the data from JSON retrieved from the Kubernetes API](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L368-L369). This is a challenge because [if there's a problem reading from the Kubernetes API then no JSON is retrieved and, thus, can't be parsed](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax/Platform.cs#L293).

[...]

PR #277, opened 2018-09-11
Surfacing environment values for podId and namespace when obtaining platform metadata.

amanda-tarafa:
Addresses #276.
If the metadata values cannot be obtained through the k8s API, we still surface the ones we obtained from the container environment.

### DIFF
--- a/Google.Api.Gax/Platform.cs
+++ b/Google.Api.Gax/Platform.cs
@@ -205,9 +205,14 @@ should be projects/<project_number>/zones/<zone_name>. {nameof(ZoneName)} curren
         public class KubernetesData
         {
             /// <summary>
-            /// The kubernetes pod name
+            /// The Kubernetes pod name
             /// </summary>
             public string PodName { get; set; }
+
+            /// <summary>
+            /// The Kubernetes namespace name
+            /// </summary>
+            public string NamespaceName { get; set; }
 
             /// <summary>
             /// JSON from https://kubernetes/api/v1/namespaces/{namespace}
@@ -257,7 +262,12 @@ should be projects/<project_number>/zones/<zone_name>. {nameof(ZoneName)} curren
                 kubernetesCaCert == null || mountInfo == null || mountInfo.Length == 0)
             {
                 // These files should contain useful data on GKE. If anything is missing then return what we already know.
-                return new KubernetesData { PodName = podName, MountInfo = mountInfo };
+                return new KubernetesData
+                {
+                    PodName = podName,
+                    NamespaceName = kubernetesNamespace,
+                    MountInfo = mountInfo
+                };
             }
 
             var handler = new HttpClientHandler
@@ -296,6 +306,7 @@ should be projects/<project_number>/zones/<zone_name>. {nameof(ZoneName)} curren
             return new KubernetesData
             {
                 PodName = podName,
+                NamespaceName = kubernetesNamespace,
                 NamespaceJson = namespaceJson,
                 PodJson = podJson,
                 MountInfo = mountInfo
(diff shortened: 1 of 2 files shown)


Google.Api.Gax/Platform.cs
```diff
@@ -209,6 +209,11 @@ public class KubernetesData
             /// </summary>
             public string PodName { get; set; }
 
+            /// <summary>
+            /// The kubernetes namespace name
```
jskeet: Given that this documentation and that for PodName are public, could we capitalize the K in Kubernetes?

Google.Api.Gax/Platform.cs
```diff
@@ -257,7 +262,12 @@ internal static async Task<KubernetesData> LoadKubernetesDataAsync()
                 kubernetesCaCert == null || mountInfo == null || mountInfo.Length == 0)
             {
                 // These files should contain useful data on GKE. If anything is missing then return what we already know.
-                return new KubernetesData { PodName = podName, MountInfo = mountInfo };
+                return new KubernetesData
+                {
+                    PodName = podName,
+                    NamespaceName = string.IsNullOrEmpty(kubernetesNamespace) ? null : kubernetesNamespace,
```
jskeet: Any reason to normalize this to null, when we later normalize to "" in `var namespaceId = ...`?

amanda-tarafa: Just that in the if above we are considering string.IsNullOrEmpty(kubernetesNamespace) as a reason not to continue forward, because we need a not empty value for kubernetesNamespace to load more data, and that is better expressed with a null than with a "" regardless of what we do afterwards with that? This is an internal method that could be used somewhere else in the future. But I do agree that it's a stretch, so I'm more than happy to remove this here, LMK.

jskeet: Okay, let's stick with it - but it would be good to document that it might be null but will never be empty (if that's the case). Likewise PodName, I'd say.

(Even though this method is internal, KubernetesData is public and is presumably exposed somewhere, so it's good to know what to expect.)

amanda-tarafa: Hmm, for podName we have a similar check (at the beginning of this method) but we are not normalizing the value here. Maybe since that's been like that for long it's better not to change it, and then it might be better as well to be consistent and not normalize kubernetesNamespace? Sorry for the back and forth.

jskeet: It happens - not a problem :)

I agree we probably shouldn't change the behavior, and we should be consistent across the two properties. I would start by adding documentation indicating possible values returned from whatever method returns it, and we can go from there.

amanda-tarafa: Agreed and done. On the bright side, although KubernetesData is public, we don't have a public method that returns it or a public property of KubernetesData type. It is public so that client code can populate an instance an use it to to create a GkePlatformDetails instance using GkePlatformDetails.TryLoad which is the one that normalizes to "", that is, it is prepared to handle all values of both PodName and NamespaceName.

amanda-tarafa:
Thanks! I'm OK with this PR. We can chat on Friday about the [possibility](https://github.com/googleapis/gax-dotnet/issues/276#issuecomment-420034770) of using the downward API as a fallback, but that would be an entirely new feature request anyway.
Will commit the review changes and merge on green.

Google.Api.Gax/Platform.cs
jskeet: Aha. In that case, the documentation is needed on TryLoad instead, which should say how it handles both null and empty values - and we can do whatever's most convenient in the code creating the instance :)

amanda-tarafa: Yep, that's what I did.

PR #281, opened 2018-11-06
Additional Task-related extension methods:

jskeet:
- WaitWithUnwrappedExceptions overloads to match Wait
- TaskCompletionSource.WithCancellationToken

### DIFF
--- a/Google.Api.Gax.Tests/TaskCompletionSourceExtensionsTests.cs
+++ b/Google.Api.Gax.Tests/TaskCompletionSourceExtensionsTests.cs
@@ -1,0 +1,62 @@
+﻿/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Api.Gax.Tests
+{
+    public class TaskCompletionSourceExtensionsTests
+    {
+        [Fact]
+        public async Task WithCancellationToken_None()
+        {
+            var tcs = new TaskCompletionSource<int>();
+            var task = tcs.WithCancellationToken(CancellationToken.None);
+
+            tcs.SetResult(5);
+            Assert.Equal(5, await task);
+        }
+
+        [Fact]
+        public async Task WithCancellationToken_Uncancelled()
+        {
+            var tcs = new TaskCompletionSource<int>();
+            var source = new CancellationTokenSource();
+            var task = tcs.WithCancellationToken(source.Token);
+
+            tcs.SetResult(5);
+            Assert.Equal(5, await task);
+        }
+
+        [Fact]
+        public async Task WithCancellationToken_CancelledBeforeCreation()
+        {
+            var source = new CancellationTokenSource();
+            source.Cancel();
+            var tcs = new TaskCompletionSource<int>();
+            var task = tcs.WithCancellationToken(source.Token);
+
+            await Assert.ThrowsAsync<TaskCanceledException>(() => task);
+        }
+
+        [Fact]
+        public async Task WithCancellationToken_CancelledAfterCreation()
+        {
+            var tcs = new TaskCompletionSource<int>();
+            var source = new CancellationTokenSource();
+            var task = tcs.WithCancellationToken(source.Token);
+            source.Cancel();
+
+            await Assert.ThrowsAsync<TaskCanceledException>(() => task);
+        }
+    }
+}
--- a/Google.Api.Gax/TaskCompletionSourceExtensions.cs
+++ b/Google.Api.Gax/TaskCompletionSourceExtensions.cs
@@ -1,0 +1,44 @@
+﻿/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax
+{
+    /// <summary>
+    /// Extension methods for <see cref="TaskCompletionSource{TResult}"/>.
+    /// </summary>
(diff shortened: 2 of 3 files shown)


Google.Api.Gax/TaskCompletionSourceExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+            GaxPreconditions.CheckNotNull(source, nameof(source));
+            if (!cancellationToken.CanBeCanceled)
+            {
+                return source.Task;
+            }
+            return Wait();
+
+            async Task<TResult> Wait()
+            {
+                using (var registration = cancellationToken.Register(() => source.TrySetCanceled()))
```
chrisdunelm: nit: `registration` local not required.

Google.Api.Gax/TaskExtensions.cs
```diff
@@ -49,7 +50,98 @@ public static void WaitWithUnwrappedExceptions(this Task task)
...
+        }
 
+        /// <summary>
+        /// Synchronously waits for the given task to complete.
+        /// Any <see cref="AggregateException"/> thrown is unwrapped to the first inner exception.
+        /// </summary>
+        /// <param name="task">The task to wait for.</param>
+        /// <param name="millisecondsTimeout">The number of milliseconds to wait, or
+        /// -1 to wait indefinitely.</param>
+        public static void WaitWithUnwrappedExceptions(this Task task, int millisecondsTimeout)
```
chrisdunelm: Do we use these overloads with `int` timeout parameters? If not, I'd rather we didn't include them here, as I don't really like `int`s being used as durations.

Google.Api.Gax.Tests/TaskCompletionSourceExtensionsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            var tcs = new TaskCompletionSource<int>();
+            var source = new CancellationTokenSource();
+            var task = tcs.WithCancellationToken(source.Token);
+
+            tcs.SetResult(5);
+            Assert.Equal(5, await task);
+        }
+
+        [Fact]
+        public async Task WithCancellationToken_Cancelled()
```
chrisdunelm: Consider adding a test with an already-cancelled cancellation-token?

Google.Api.Gax/TaskCompletionSourceExtensions.cs
jskeet: Good point. Will remove.

Google.Api.Gax.Tests/TaskCompletionSourceExtensionsTests.cs
jskeet: Sure.

Google.Api.Gax/TaskExtensions.cs
jskeet: Hmm. The point would be to make it trivial to go from a `Wait(anything)` to `WaitWithUnwrappedExtensions(anything)`, but I do see your point. Let's see what Amanda thinks :)

jskeet: Note that if we do remove the milliseconds-based ones, we'd want to add a "TimeSpan and CancellationToken" one - which doesn't exist on Wait itself.

chrisdunelm:
Agreed about adding "TimeSpan and CancellationToken" method.

Google.Api.Gax/TaskExtensions.cs
```diff
@@ -49,7 +50,98 @@ public static void WaitWithUnwrappedExceptions(this Task task)
...
+        public static void WaitWithUnwrappedExceptions(this Task task, TimeSpan timeout)
+        {
+            try
+            {
+                task.Wait(timeout);
+            }
+            catch (AggregateException e)
+            {
+                // Unwrap the first exception, a bit like await would.
+                // It's very unlikely that we'd ever see an AggregateException without an inner exceptions,
```
amanda-tarafa: super nit: without an inner exception (singular)

Google.Api.Gax/TaskExtensions.cs
```diff
@@ -49,7 +50,98 @@ public static void WaitWithUnwrappedExceptions(this Task task)
...
+        public static void WaitWithUnwrappedExceptions(this Task task, int millisecondsTimeout)
+        {
+            try
+            {
+                task.Wait(millisecondsTimeout);
+            }
+            catch (AggregateException e)
+            {
+                // Unwrap the first exception, a bit like await would.
+                // It's very unlikely that we'd ever see an AggregateException without an inner exceptions,
```
amanda-tarafa: nit: same here, inner exception (singular) and everywhere else.

Google.Api.Gax/TaskExtensions.cs
amanda-tarafa: Agree with `int` as duration being ugly, but it's already in `Wait`, and we are just wrapping `Wait` so, might make sense here. These are just helper methods after all. I'm more incline to leave it as is.

Google.Api.Gax/TaskExtensions.cs
```diff
@@ -49,7 +50,98 @@ public static void WaitWithUnwrappedExceptions(this Task task)
...
+        }
+
+        /// <summary>
+        /// Synchronously waits for the given task to complete.
+        /// Any <see cref="AggregateException"/> thrown is unwrapped to the first inner exception.
+        /// </summary>
+        /// <param name="task">The task to wait for.</param>
+        /// <param name="timeout">A TimeSpan that represents the number of milliseconds to wait, or
+        /// -1 milliseconds to wait indefinitely.</param>
+        public static void WaitWithUnwrappedExceptions(this Task task, TimeSpan timeout)
```
amanda-tarafa: name wise, I would prefer `WaitAndUnwrapExceptions`, but not set on it.

Google.Api.Gax/TaskExtensions.cs
jskeet: Will fix this everywhere.

Google.Api.Gax/TaskExtensions.cs
jskeet: We've already got the original WaitWithUnwrappedExceptions method, so we can't really change that now.

jskeet:
Will chat with Chris about the millisecond part...

jskeet:
Discussed with Chris - repushed with the comment fixes; will merge on green.

PR #282, opened 2018-11-06
Enhancements to FakeScheduler

jskeet:
These have been prototyped in Spanner already, so we're pretty
confident in them.


Google.Api.Gax.Testing/FakeScheduler.cs
```diff
@@ -180,78 +195,166 @@ public void Run<T>(Func<Task<T>> taskProvider)
...
 
-        private Task StartLoopAsync(DateTime simulatedTimeout)
+        /// <summary>
+        /// Runs the scheduler for the given number of seconds. If the task completes normally,
+        /// the clock should have advanced by the given number of seconds.
+        /// </summary>
+        /// <param name="seconds">How many seconds (in simulated time) to run for</param>
+        /// <returns>A task which will complete when scheduler has processed tasks up until
+        /// the given time.</returns>
+        public Task RunAndPauseForSeconds(int seconds) => RunAndPause(TimeSpan.FromSeconds(seconds));
```
chrisdunelm: `RunAndPauseForSecondsAsync` ?

Google.Api.Gax.Testing/FakeScheduler.cs
```diff
@@ -180,78 +195,166 @@ public void Run<T>(Func<Task<T>> taskProvider)
...
+
+        /// <summary>
+        /// Runs the scheduler for the given amount of time. If the task completes normally,
+        /// the clock should have advanced by the given TimeSpan.
+        /// </summary>
+        /// <param name="time">How long (in simulated time) to run for</param>
+        /// <returns>A task which will complete when scheduler has processed tasks up until
+        /// the given time.</returns>
+        public async Task RunAndPause(TimeSpan time)
```
chrisdunelm: `...Async` ?

Google.Api.Gax.Testing/FakeScheduler.cs
jskeet: I think I originally had that, then made it a void method, then made it async again but forgot to change the name back. Admittedly it's a pretty unwieldy name - if you have any other suggestions, I'd love to hear them...

jskeet: Hmm. The parameter name isn't as clear when reading code using it - especially as often unspecified time units are in milliseconds. Maybe it would be okay though...

jskeet: Yes, that's what I want to convey - it's "run for this duration" rather than "run until a delegate completes". I'm very, very much up for an entirely different name if that would be useful. (I do want some method with "seconds" though, as having to write TimeSpan.FromSeconds everywhere was really annoying.)

Let's keep thinking about names.

amanda-tarafa: What about Advance?

- `AdvanceAsync(Timespan)`
- `AdvanceSecondsAsync(int)`
- `AdvanceUntilAsync(Task)`

That said, I don't think it's bad as it is, except maybe for the missing Aysnc.

jskeet: - Advance feels like it might be skipping over rather than running along the period
- I've been avoiding Until before that suggests it's to a time rather than for a duration (although we could definitely add a method that did that)

Heck, maybe we could just overload RunAsync? Or do you think they're sufficiently different to merit different names?

chrisdunelm: Overloading `RunAsync` sounds good to me.

amanda-tarafa: I think you haven't pushed.

jskeet: Whoops. Probably forgot to force push...

jskeet: Done now :)

jskeet:
Hmm... the Travis failure *might* mean I've messed something up. Restarted, and will merge on green, but investigate further if it fails again.

PR #285, opened 2018-11-07
Update Google.Api.Gax.Grpc.Gcp

jskeet:
- Use the latest Grpc.Gcp package
- Use the latest Grpc.Core package
- Remove our own ChannelOptions equality comparer as Grpc.Core now does this for us

### DIFF
--- a/Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
+++ b/Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
@@ -112,46 +112,6 @@ namespace Google.Api.Gax.Grpc.Gcp
             }
         }
 
-        // Note: this class will disappear if/when https://github.com/grpc/grpc/issues/16533 is fixed.
-        private class ChannelOptionComparer : IEqualityComparer<ChannelOption>
-        {
-            public static readonly ChannelOptionComparer Instance = new ChannelOptionComparer();
-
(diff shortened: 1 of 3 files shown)


Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
```diff
@@ -166,12 +126,12 @@ public Key(ServiceEndpoint endpoint, List<ChannelOption> options)
             public override int GetHashCode() =>
                 EqualityHelpers.CombineHashCodes(
                     Endpoint.GetHashCode(),
-                    EqualityHelpers.GetListHashCode(Options, ChannelOptionComparer.Instance));
+                    EqualityHelpers.GetListHashCode(Options, EqualityComparer<ChannelOption>.Default));
```
chrisdunelm: Isn't `EqualityComparer<ChannelOption>.Default` the default? So specifying it here (and a few lines down) isn't necessary?

jskeet: EqualityListHelpers.GetListHashCode doesn't have any default :)

We could remove it from SequenceEqual though.

jskeet: (And I'll do so now)

PR #289, opened 2018-11-12
Update to Grpc.Gcp version 1.1.1

jskeet:
No version bump in Google.Api.Gax.Grpc.Gcp, as we can just give
direct dependencies for now. This makes sure that in GAX 2.6.0 we'll
use the right version though.


amanda-tarafa:
But several unit tests are failing.

jskeet:
Hmm. We've seen that BufferedWriterClientTest fail very occasionally before. Interesting if it's becoming more common. (Completely unrelated to this PR though - different project entirely.)

PR #291, opened 2019-02-05
Make platform version detection more robust

### LINKED ISSUE #231: Microsoft.Extensions.PlatformAbstractions TypeInitializationException exception
Platform: Xamarin .Net project for mobile

The use of Microsoft.Extensions.PlatformAbstractions causes an exception to be thrown as per the description in this Xamarin bug report. Also includes a reproduction project.

https://bugzilla.xamarin.com/show_bug.cgi?id=60698

The PlatformAbstractions is reported as being dead. So can the replacements be substituted instead. https://github.com/aspnet/PlatformAbstractions

jskeet:
It's unfortunate that we still need to use reflection in order to support .NET Standard 1.3, but we do so in a more robust way than the previous dependency did.

Additionally, we now fail gracefully: it's better to use an unknown version number for the environment than failing completely.

Fixes #231 

cc @amanda-tarafa

### DIFF
--- a/Google.Api.Gax/VersionHeaderBuilder.cs
+++ b/Google.Api.Gax/VersionHeaderBuilder.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Versioning;
 
 namespace Google.Api.Gax
 {
@@ -19,6 +20,8 @@ namespace Google.Api.Gax
     /// </summary>
     public sealed class VersionHeaderBuilder
     {
+        private static readonly Lazy<string> s_environmentVersion = new Lazy<string>(GetEnvironmentVersion);
+
         /// <summary>
         /// The name of the header to set.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -54,12 +57,33 @@ public VersionHeaderBuilder AppendAssemblyVersion(string name, System.Type type)
...
-#endif
+        public VersionHeaderBuilder AppendDotNetEnvironment() => AppendVersion("gl-dotnet", s_environmentVersion.Value);
+
+        private static string GetEnvironmentVersion()
+        {
+            try
+            {
+                // Assembly.GetEntryAssembly() isn't available in netstandard1.3. Attempt to fetch it with reflection, which is ugly but should work.
+                // This is a slightly more robust version of the code we previously used in Microsoft.Extensions.PlatformAbstractions.
+                var getEntryAssemblyMethod = typeof(Assembly)
```
chrisdunelm: Slightly amusing that you don't get this by name ;)

PR #299, opened 2019-02-27
Potentially fixing Kokoro not finding Gax.sln

amanda-tarafa:
This fix works in local.


amanda-tarafa:
Thanks. I'll create a testing VM if this doesn't work so as not be trying back and forth.

PR #303, opened 2019-03-05
Creating and adapting release scripts for Kokoro.

amanda-tarafa:
This ran from Kokoro almost to completion, and it wasn't successful because I was explicitly using a "fake_key" value for the nuget.org key, so as not to push any package to nuget. The failure was when pushing packages to nuget with the following expected error.

```
info : Pushing Google.Api.CommonProtos.1.5.0.nupkg to 'https://www.nuget.org/api/v2/package'...
info :   PUT https://www.nuget.org/api/v2/package/
info :   Forbidden https://www.nuget.org/api/v2/package/ 560ms
error: Response status code does not indicate success: 403 (The specified API key is invalid, has expired, or does not have permission to access the specified package.).
```

The two scripts in `.kokoro` are new. The `buildrelease.sh` script in the root directory is a modified version of the one that existed already. The main changes are as follows:

- The release is now based in a commit (which is what we get from Kokoro) and not a tag. The tags still need to be created beforehand, because they will be used to determined which packages to push to nuget.org. There can be several tags pointing at the same commit and all of those will be taken into account.
- The process still generates all nuget packages, but it deletes the ones that are not "included" in one of the tags pointing to the given commit. This happens so that the scripts in `.kokoro` can blindly push all packages to nuget.org.
  - For instance, if there's only one tag Google.Api.Gax-version then all the packages except the one generated for Google.Api.CommonProtos will be kept and pushed to nuget.
  - If there are two tags pointing to the given commit, say Google.Api.Gax.Grpc-version and Google.Api.CommonProtos-version then only the generated packages for Google.Api.Gax.Grpc.* and Google.Api.CommonProtos will be kept and pushed to nuget.

It's important then to generate the tags following the same format that has been used till now, so that Kokoro pushes the right packages to nuget.

### DIFF
--- a/buildrelease.sh
+++ b/buildrelease.sh
@@ -1,10 +1,16 @@
 #!/bin/bash
 
 # Script to perform a release build for GAX, based on
-# a tag which should already have been created.
-# This builds both GAX and the common protos; you should only push
-# the nuget packages you actually want to update.
-# Sample: buildrelease.sh Google.Api.Gax-2.1.0
+# a commit for which tags have been created.
+# This builds both GAX and the common protos; but will delete
+# all built packages that don't match one of the tags in the commit
+# hash. For instance, if there's only one tag in the commit hash 
+# that starts with Google.Api.Gax.Grpc then only the following nuget
+# packages will remain after the script is done:
+# Google.Api.Gax.Grpc.<version>.nupkg
+# Google.Api.Gax.Grpc.Gcp.<version>.nupkg
+# Google.Api.Gax.Grpc.Testing.<version>.nupkg
+# Sample: buildrelease.sh 847118821d8a8b4ac2ea3bbe692acea29c14c4da
 
 # The repo is cloned into a fresh "releasebuild" directory.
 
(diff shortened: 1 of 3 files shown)


buildrelease.sh
```diff
@@ -1,27 +1,28 @@
...
+# a commit for which tags have been created.
+# This builds both GAX and the common protos; but will delete
+# all built packages that don't match one of the tags in the commit
+# hash. For instance, if there's only one tag in the commit hash 
+# that starts with Google.Api.Gax.Grpc then only the following nuget
+# packages will remain after the script is done:
+# Google.Api.Gax.Grpc.<version>.nupkg
+# Google.Api.Gax.Grpc.Gcp.<version>.nupkg
+# Google.Api.Gax.Grpc.Testing.<version>.nupkg
+# Sample: buildrelease.sh 847118821d8a8b4ac2ea3bbe692acea29c14c4da
```
jskeet: So to check, if we tag "Google.Api.Gax" (which is what we normally do) that will push everything apart from CommonProtos, right?

amanda-tarafa: Exactly, and if you tag Google.Api.Gax.Grpc that will push only the Google.Api.Gax.Grpc.* packages.
And you can have two tags on the same commit, one for Google.Api.Gax.Grcp and one for Google.Api.CommonProtos which will push Google.Api.Gax.Grcp.* and CommomProtos but not the other Google.Api.Gax packages.
I checked and noticed that the tags are "always" like this. I added examples on the PR comments.

PR #317, opened 2019-05-23
Channel options in channel pool and ClientBuilderBase

jskeet:
I've verified that this works with Datastore (in terms of the user agent)


Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -22,7 +22,12 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class ChannelPool
     {
-        private static readonly List<ChannelOption> s_defaultOptions = new List<ChannelOption> { GrpcChannelOptions.OneMinuteKeepalive };
+        private static readonly List<ChannelOption> s_defaultOptions = new List<ChannelOption>
```
chrisdunelm: Type this field as `IReadOnlyList<>`?

jskeet: Done, and likewise in ClientBuilderBase. (It's possible we could have a single list, but I'd need to think through that carefully. Maybe something for another PR.)

jskeet:
Thanks - will merge on green.

PR #328, opened 2019-07-10
Regenerate common protos

jskeet:


jskeet:
Okay, now changed to update the protobuf version and release Google.Api.Gax.* as well.

PR #333, opened 2019-07-24
Provide support for Cloud Run as a platform

jskeet:
This is sufficiently different from GCE that it makes sense to break it out. It's possible that KNative apps running in different environments will have *similar* details, but they may not have a project ID for example... so for now, let's be specific. We canalways add a generic KNative type later.

### DIFF
--- a/Google.Api.Gax/Platform.cs
+++ b/Google.Api.Gax/Platform.cs
@@ -43,6 +43,113 @@ namespace Google.Api.Gax
         /// Execution platform is Google Container Engine (Kubernetes).
         /// </summary>
         Gke = 3,
+
+        /// <summary>
+        /// Execution platform is Google Cloud Run.
+        /// </summary>
+        CloudRun = 4
+    }
+
+    /// <summary>
+    /// Google Cloud Run details.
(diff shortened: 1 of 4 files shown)


Google.Api.Gax/Platform.cs
```diff
@@ -43,6 +43,116 @@ public enum PlatformType
...
+            }
+            
+            var projectId = metadata["project"]?["projectId"]?.ToString();
+            var zoneName = metadata["instance"]?["zone"]?.ToString();
+            if (projectId == null || zoneName == null || !Platform.s_zoneTemplate.TryParseName(zoneName, out var zoneResourceName))
+            {
+                return null;
+            }
+            string location = zoneResourceName[1];
+            // FIXME: Temporary hack, do not commit
```
chrisdunelm: Should this be fixed?

jskeet: That's weird - I thought I'd pushed a version which had that removed. Will check.

jskeet: Doh, hadn't committed the change. Removed.

PR #335, opened 2019-07-24
Give more generous timeouts in tests

### LINKED ISSUE #326: BufferedClientStreamWriterTest is flaky
This seems to be a problem in Travis more than locally, but we've had it enough that we should really do something about it.

See https://travis-ci.org/googleapis/gax-dotnet/builds/545666675 for an example.

jskeet:
Attempts to fix #326.

It'll be hard to know whether it's really worked without merging and seeing what happens over time.


jskeet:
Well I guess the failure here proves there's more work to do. I wonder whether there's actually a race condition here. Will have a look.

jskeet:
Okay, I suspect this still won't fix everything, but will at least *reduce* the flakiness - PTAL.

PR #346, opened 2019-08-12
Use gke_container instead of container for GKE monitored resources

### LINKED ISSUE #233: monitored resource type "container" or "gke_container"?
Hello,

I tried to switch to using  `MonitoredResourceBuilder.FromPlatform()` now that [it's supposed to work for GKE]( https://github.com/googleapis/gax-dotnet/issues/209) but I'm getting this error:

```
....Monitoring.Stackdriver.Publisher Failed to create 1 timeseries status
Status(StatusCode=InvalidArgument, Detail="Field timeSeries[0].resource.type had an invalid value of 
"container": Unrecognized resource name."), dropping these. (108/108)
```
Should [`container` here](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax.Grpc/MonitoredResourceBuilder.cs#L89) perhaps be `gke_container`?

As an experiment, I used:

```
var monitoredResource = MonitoredResourceBuilder.FromPlatform();
monitoredResource.Type = "gke_container";
```

But then I get another error!

```
Kaggle.Monitoring.Stackdriver.Publisher Failed to create 1 timeseries status
Status(StatusCode=InvalidArgument, Detail="One or more TimeSeries could not be written: Unrecognized 
region or location.: timeSeries[0]")
```

Looking at the labels, the value for `zone` does look wrong (`projects/.../zones/us-...-d`), over what we used before (simply `us-...-d`):

https://cloud.google.com/monitoring/api/resources#tag_gke_container isn't too useful in determining what this exactly should be, but I'd bet on the short name.

=>

```
var monitoredResource = MonitoredResourceBuilder.FromPlatform();
monitoredResource.Type = "gke_container";
monitoredResource.Labels["zone"] = "us-...-d";
```

which works (can write time series and labels look ok to me).

My code looks a bit closer to what it should be, but I do feel I'm just trading one TODO for another here...

jskeet:
Fixes #233.

Note that this requires additional changes in Google.Cloud.Diagnostics.AspNetCore and Google.Cloud.Diagnostics.Common; new packages should be pushed out at the same time.

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
@@ -122,7 +122,7 @@ namespace Google.Api.Gax.Grpc.Tests
             };
             var platform = new Platform(GkePlatformDetails.TryLoad(metadataJson, kubernetesData));
             var resource = MonitoredResourceBuilder.FromPlatform(platform);
-            Assert.Equal("container", resource.Type);
+            Assert.Equal("gke_container", resource.Type);
             Assert.Equal(new Dictionary<string, string>
             {
                 { "project_id", "FakeProjectId" },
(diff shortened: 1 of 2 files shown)


jskeet:
https://github.com/googleapis/google-cloud-dotnet/pull/3358 is the corresponding google-cloud-dotnet change.

PR #354, opened 2019-10-22
Stability header handling

jskeet:
Draft as this is for discussion at the moment. We could potentially merge the first commit, but not the second - let me know if that would be preferable.

### DIFF
--- a/Google.Api.Gax.Grpc/ServiceSettingsBase.cs
+++ b/Google.Api.Gax.Grpc/ServiceSettingsBase.cs
@@ -17,10 +17,18 @@ namespace Google.Api.Gax.Grpc
     public abstract class ServiceSettingsBase
     {
         /// <summary>
-        /// Constructs a new service settings base object with a default version header, unset call settings and
+        /// Constructs a new service settings base object for a stable API with a default version header, unset call settings and
         /// unset clock.
         /// </summary>
-        protected ServiceSettingsBase()
+        protected ServiceSettingsBase() : this(LaunchStage.Ga)
+        {
+        }
+
+        /// <summary>
+        /// Constructs a new service settings base object for the given API stability level with a default version header, unset call settings and
+        /// unset clock.
+        /// </summary>
+        protected ServiceSettingsBase(LaunchStage stabilityLevel)
         {
             VersionHeaderBuilder = new VersionHeaderBuilder()
                 .AppendDotNetEnvironment()
@@ -28,6 +36,7 @@ namespace Google.Api.Gax.Grpc
                 .AppendAssemblyVersion("gapic", GetType())
                 .AppendAssemblyVersion("gax", typeof(CallSettings))
                 .AppendAssemblyVersion("grpc", typeof(Channel));
+            StabilityLevel = stabilityLevel;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


jskeet:
I think we should probably decide what level of control we want to give users first. Options I can see are:

- No control: you have to use the stability associated with the package.
- Individual call control: you can pass in a CallSettings to modify the stability, but it's awkward. (It's probably hard *not* to allow this, to be honest.)
- ServiceSettingsBase-level control: you can tell a beta package to act as a stable package or vice versa, and you can also do it on a per-call basis.

For the second and third options, we probably want to work out how easy to make that. (With no help, users would need to hunt out the right header etc.)

I can definitely see cases where a user wants to import the beta package because in one place they use beta functionality (and they don't want both packages as dependencies, with the pain of extern aliases) but they want to only use stable functionality in all other places.

chrisdunelm:
In my opinion we should provide no (obvious) control to the user - the stability should be as associated with the package - because this introduces another layer of complication/confusion about versioning with almost no advantage.
If a user requires a beta server feature, then they use a beta client library.

I don't think we should stop a user from manually adding the suitable header if they really want to, but we should provide nothing to support this.

jskeet:
Okay. I'll have a think about what that would mean in terms of putting it in ServiceSettingsBase. We could potentially have it as a read-only property and then add a setter later if we had enough demand and good justification.

amanda-tarafa:
I think as well we shouldn't give the user control over the stability, it should be associated with the package.
Having the same package behave as beta or stable for one or two features might be sort of straightforward for exclusively generated libraries, but for the handwritten ones that can become a nightmare. Take Spanner.V1 for instance, and imagine both BatchCreateSession and FeatureA are beta. On our beta package we have already changed the session pool to use BatchCreateSession but a client "activates" beta only for FeatureA... I don't know how to make that work unless we ourselves start baking feature toggling into our handwritten client libraries, and I personally woudl rather we don't. (Unless I'm missing something here).
Also, why add the complexity, which users would that really benefit. I would argue that users that include both stable and beta packages of our libraries either know very well what they are doing, in which case they can handle the aliasing, or really have no idea, and probably have included both packages just by accident, in which case, no amount of options we offer will help them.
And last but not least, if we start restricting the user control over the stability header we can always make it less restrictive later, if we see that there is a huge demand for something like that. If we start allowing user control over it, then it's far more difficult to disallow it in the future.
Again, take my comments with a pinch of salt, I'm not overly familiar with all this.

jskeet:
Added another commit, which I *think* does what we're all happy with.

Google.Api.Gax.Grpc/ServiceSettingsBase.cs
```diff
@@ -17,17 +17,26 @@ namespace Google.Api.Gax.Grpc
     public abstract class ServiceSettingsBase
     {
         /// <summary>
-        /// Constructs a new service settings base object with a default version header, unset call settings and
+        /// Constructs a new service settings base object a stable API with a default version header, unset call settings and
```
amanda-tarafa: ... **for** a stable

jskeet:
Rebased this into a single commit. We could merge now, or leave this pending - I don't much mind.

amanda-tarafa:
Fine by me either way as well.

jskeet:
Closing this PR now as we won't need it.

PR #355, opened 2019-10-28
Replace the very specific WithVersionHeader method with WithMergedBaseCallSettings

jskeet:
WithVersionHeader was only used in ClientHelper, which can easily migrate to the new call.


jskeet:
(This is the first part of the stability-oriented version headers, currently in #354.)

jskeet:
Merging now as Amanda was happy in the other PR :)

Issue #363, opened 2019-11-13
Consider returning IAsyncEnumerable<T> in streaming calls


jskeet:
`IAsyncEnumerator<T>` doesn't support "await foreach". While it's reasonable for our `GetResponseStream()` method returns an `IAsyncEnumerator<T>` in that it's single-shot, it's harder to use than if we returned an `IAsyncEnumerable<T>` that failed if it's enumerated more than once.

We need to decide this before GAX v3 goes out.

cc @chrisdunelm @amanda-tarafa

jskeet:
Note that just changing to `IAsyncEnumerable<T>` would make it harder for non-C# 8 users to use.

Potentially we could return a `SingleShotAsyncEnumerable<T>` that implements both `IAsyncEnumerable<T>` and `IAsyncEnumerator<T>`. (Name pending...)

jskeet:
Fixed by #365.

PR #364, opened 2019-11-13
Modify GetResponseStream to return IAsyncEnumerable<T>

jskeet:
First option for fixing #363


chrisdunelm:
I take it this PR isn't required if #365 is merged?

jskeet:
Yes. These are two different options for fixing the same problem.

jskeet:
Abandoned in favour of #365.

PR #365, opened 2019-11-13
Make GetResponseStream return a combination IAsyncEnumerable/IAsyncEnumerator

jskeet:
This is a second option for #363.


Google.Api.Gax.Grpc/AsyncResponseStream.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// Moves to the next item, using the specified cancellation token.
+        /// </summary>
+        /// <remarks></remarks>
+        /// <param name="cancellationToken">The cancellation token to use for this step.</param>
+        /// <returns>A task that will complete with a result of true if the enumerator was successfully advanced to the next element, or false if the enumerator has passed the end of the collection.</returns>
+        public async ValueTask<bool> MoveNextAsync(CancellationToken cancellationToken) =>
+            await _reader.MoveNext(_cancellationToken).ConfigureAwait(false);
```
chrisdunelm: `cancellationToken` here I think, not `_cancellationToken`?

Google.Api.Gax.Grpc/AsyncResponseStream.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This type implements both of the standard asynchronous sequence interfaces for simplicity of use:
+    /// <list type="bullet">
+    ///   <item>C# 8 users can use  <c>await foreach</c> because it implements <see cref="IAsyncEnumerable{T}"/></item>
+    ///   <item>It's compatible with the System.Linq.Async package for query transformations.</item>
+    ///   <item>Pre-C# 8 users who will be calling <see cref="MoveNextAsync()" /> and <see cref="Current"/> directly don't need
+    ///   to call <see cref="GetAsyncEnumerator(CancellationToken)"/>.</item>
+    /// </list>
+    /// Note that <see cref="GetAsyncEnumerator(CancellationToken)"/> can only be called once per instance due to the "only iterate once"
```
chrisdunelm: This warning might be better in the summary.

Google.Api.Gax.Grpc/AsyncResponseStream.cs
jskeet: Humbug, yes! Will fix now.

Google.Api.Gax.Grpc/AsyncResponseStream.cs
jskeet: I'm in two minds about that, but happy to go that way.

jskeet:
Updated the PR. Let's see what Amanda thinks. If we merge this, I'll make the corresponding changes on the generator and google-cloud-dotnet.

jskeet:
Cool. Will add a test before merging, and then deal with downstream changes.

jskeet:
Okay, added some tests now (and fixed a bug that they found!) - PTAL.

jskeet:
Added tests, and fixed a bug found in those tests. PTAL.

Google.Api.Gax.Grpc.Tests/AsyncResponseStreamTest.cs
```diff
@@ -0,0 +1,109 @@
...
+            mock.Setup(x => x.MoveNext(CancellationToken.None)).Returns(Task.FromResult(true)).Verifiable();
+
+            var stream = new AsyncResponseStream<int>(mock.Object);
+            Assert.True(stream.MoveNextAsync().Result);
+
+            mock.Verify();
+        }
+
+        [Fact]
+        public void MoveNextAsync_Parameterless_AfterGetAsyncEnumerator_RespectsToken()
```
amanda-tarafa: nit: the RespectsToken part of the name threw me off, I would associate that more with the token being canceled and that cancelation respected. Consider changing to PropagatesToken, this one and the following.

jskeet:
Fixing Travis now.

Google.Api.Gax.Grpc/BidirectionalStreamingBase.cs
```diff
@@ -113,14 +113,15 @@ public virtual Task WriteCompleteAsync()
...
         /// <remarks>
         /// If this method is called more than once, all the returned enumerators will be enumerating over the
-        /// same underlying response stream, which may cause confusion.
+        /// same underlying response stream, which may cause confusion. Additionally, the sequence returned by
+        /// this method can only be iterated over a single time. Attempting to iterate more than once will cause
+        /// an <see cref="InvalidOperationException"/>.
         /// </remarks>
-        public virtual IAsyncEnumerator<TResponse> GetResponseStream(CancellationToken cancellationToken) =>
-            new ResponseStreamAdapter<TResponse>(GrpcCall.ResponseStream, cancellationToken);
+        public virtual AsyncResponseStream<TResponse> GetResponseStream() =>
```
amanda-tarafa: Maybe add an optional parameter CancellationToken and pass it on to the constructor, so the call to GetAsyncEnumerator can be skipped, I think it makes sense given that AsyncResponseStream in an IAsyncEnumerator itself?

Google.Api.Gax.Grpc.Tests/AsyncResponseStreamTest.cs
jskeet: Works for me.

Google.Api.Gax.Grpc/BidirectionalStreamingBase.cs
jskeet: That means there'd be *three* places you could specify a cancellation token - I'd rather not do that. At the moment it's reasonably simple: if you want to specify the same cancellation token for everything, use it as an `IAsyncEnumerable` and pass the cancellation token into GetAsyncEnumerator. If you want to specify a potentially-different cancellation token on each call, just do that on MoveNextAsync.

PR #366, opened 2019-11-26
New CallSettings shape

jskeet:
The tests don't even build yet - I wanted to get early feedback on the shape of the API.

With the [gRPC-based retry config](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) we always specify the overall deadline as the RPC deadline, so there's only one set of initial/max/multiplier to consider - so I've inlined that into RetrySettings.

The overall deadline is also somewhat separate from retry settings, so I've extracted that to the top level, and removed both BackoffSettings and CallTiming.

I believe this will now be significantly easier to use, due to the much-reduced nesting.


chrisdunelm:
This all looks good to me. I like the simplification of retry.

jskeet:
This is now ready for actual review. We expect that server-side streaming retry may be affected by https://github.com/googleapis/gax-dotnet/issues/367

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -133,55 +170,33 @@ public static CallSettings WithEarlierDeadline(this CallSettings settings, DateT
...
-        public static CallSettings WithExpiration(this CallSettings settings, Expiration expiration)
+        /// <param name="callSettings">The call settings for the call. May be null.</param>
+        /// <returns><paramref name="callSettings"/></returns>
+        internal static CallSettings ValidateNoRetry(this CallSettings callSettings)
         {
-            GaxPreconditions.CheckNotNull(expiration, nameof(expiration));
-            var retry = settings?.Timing?.Retry?.WithTotalExpiration(expiration);
-            var timing = retry == null ? CallTiming.FromExpiration(expiration) : CallTiming.FromRetry(retry);
-            return settings.WithCallTiming(timing);
+            GaxPreconditions.CheckState(callSettings?.Retry == null, "Retry not permitted for this operation type");
```
chrisdunelm: nit: `is null` ?

Google.Api.Gax.Grpc/ApiBidirectionalStreamingCall.cs
```diff
@@ -65,7 +65,7 @@ public sealed class ApiBidirectionalStreamingCall<TRequest, TResponse>
         public AsyncDuplexStreamingCall<TRequest, TResponse> Call(CallSettings perCallCallSettings)
         {
             CallSettings effectiveCallSettings = BaseCallSettings.MergedWith(perCallCallSettings);
-            GaxPreconditions.CheckState(effectiveCallSettings?.Timing?.Type != CallTimingType.Retry,
+            GaxPreconditions.CheckState(effectiveCallSettings.Retry == null,
```
chrisdunelm: Could use `ValidateNoRetry()`?
nit: Or use `is null` ?

Google.Api.Gax.Grpc/CallSettings.cs
```diff
@@ -19,24 +19,26 @@ public sealed class CallSettings
...
-        internal static CallSettings CancellationTokenNone { get; } = new CallSettings(default(CancellationToken), null, null, null, null, null);
+        internal static CallSettings CancellationTokenNone { get; } = new CallSettings(default(CancellationToken), null, null, null, null, null, null);
 
         /// <summary>
         /// Constructs an instance with the specified settings.
         /// </summary>
         /// <param name="cancellationToken">Cancellation token that can be used for cancelling the call.</param>
         /// <param name="credentials">Credentials to use for the call.</param>
-        /// <param name="timing"><see cref="CallTiming"/> to use, or null for default retry/expiration behavior.</param>
+        /// <param name="expiration"><see cref="Expiration"/> to use, or null for default expiration behavior.</param>
```
chrisdunelm: nit: `...or <c>null</c>...`
and below

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -61,28 +61,49 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
...
         /// <summary>
-        /// Returns a new <see cref="CallSettings"/> with the specified call timing,
+        /// Returns a new <see cref="CallSettings"/> with the specified expiration,
         /// merged with the (optional) original settings specified by <paramref name="settings"/>.
         /// </summary>
         /// <param name="settings">Original settings. May be null, in which case the returned settings
         /// will only contain call timing.</param>
-        /// <param name="timing">Call timing for the new call settings.
+        /// <param name="expiration">Expiration to use in the returned settings, possibly as part of a retry. Must not be null.</param>
+        /// <returns>A new set of call settings, or null if both parameters are null.</returns>
```
chrisdunelm: This looks incorrect. Both parameters can never be null.

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -61,28 +61,49 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
...
+        /// </summary>
+        /// <param name="settings">Original settings. May be null, in which case the returned settings
+        /// will only contain call timing.</param>
+        /// <param name="retry">Call timing for the new call settings.
         /// This may be null, in which case any call timing in <paramref name="settings"/> are
         /// not present in the new call settings. If both this and <paramref name="settings"/> are null,
         /// the return value is null.</param>
         /// <returns>A new set of call settings, or null if both parameters are null.</returns>
-        public static CallSettings WithCallTiming(
+        public static CallSettings WithRetry(
```
chrisdunelm: Handling of a null parameter, `retry` in this case, is inconsistent with the `WithExpiration` method above.
I'm not sure which is best, but they should probably be the same!

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -116,6 +137,22 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
         public static CallSettings WithTrailingMetadataHandler(this CallSettings settings, Action<Metadata> handler) =>
             settings.MergedWith(CallSettings.FromTrailingMetadataHandler(handler));
 
+        /// <summary>
+        /// Returns a <see cref="CallSettings"/> which will have the specified timeout.
+        /// </summary>
+        /// <param name="settings">Existing settings. May be null, meaning there are currently no settings.</param>
+        /// <param name="deadline">The deadline for the new settings.</param>
+        public static CallSettings WithDeadline(this CallSettings settings, DateTime deadline) =>
```
chrisdunelm: xmldoc missing a `returns` section.

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -116,6 +137,22 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
...
+            settings.WithExpiration(Expiration.FromDeadline(deadline));
+
+        /// <summary>
+        /// Returns a <see cref="CallSettings"/> which will have the specified deadline.
+        /// </summary>
+        /// <param name="settings">Existing settings. May be null, meaning there are currently no settings.</param>
+        /// <param name="timeout">The timeout for the new settings.</param>
+        public static CallSettings WithTimeout(this CallSettings settings, TimeSpan timeout) =>
```
chrisdunelm: xmldoc missing a `returns` section.

Google.Api.Gax.Grpc/RetrySettings.cs
```diff
@@ -53,64 +59,37 @@ public sealed class RetrySettings
...
-            BackoffSettings timeoutBackoff,
-            Expiration totalExpiration,
-            Predicate<RpcException> retryFilter) : this(retryBackoff, timeoutBackoff, totalExpiration, retryFilter, null)
-        {
-        }
+        public IJitter BackoffJitter { get; }
 
         /// <summary>
-        /// Constructs an instance with the given configuration.
+        /// FIXME
```
chrisdunelm: Fix ;)

jskeet:
Thanks - will check all of these next week :)

Google.Api.Gax.Grpc/RetrySettings.cs
jskeet: Have kept this internal, but added static factory methods.

Google.Api.Gax.Grpc/CallSettingsExtensions.cs
```diff
@@ -70,21 +70,20 @@ public static CallSettings MergedWith(this CallSettings original, CallSettings o
         /// merged with the (optional) original settings specified by <paramref name="settings"/>.
         /// </summary>
         /// <param name="settings">Original settings. May be null, in which case the returned settings
-        /// will only contain call timing.</param>
-        /// <param name="expiration">Expiration to use in the returned settings, possibly as part of a retry. Must not be null.</param>
-        /// <returns>A new set of call settings, or null if both parameters are null.</returns>
+        /// will only contain the expiratoin.</param>
+        /// <param name="expiration">Expiration to use in the returned settings, possibly as part of a retry. May be null,
+        /// in which case any expiration in <paramref name="settings"/> is not present in the new call settings. If
```
chrisdunelm: I'm not sure this is the most suitable semantics for `expiration` being `null`. We have `Expiration.None` to mean no expiration; to my mind, null should mean it passes through the expiration from `settings`?

Google.Api.Gax.Grpc/RetrySettings.cs
```diff
@@ -161,11 +163,21 @@ public static Predicate<RpcException> FilterForStatusCodes(IEnumerable<StatusCod
...
         /// <summary>
-        /// Builds a new RetrySettings which is identical to this one, but with the given expiration.
+        /// Works out the next backoff from the current one, based on the multiplier and maximum.
         /// </summary>
-        /// <param name="expiration">New expiration</param>
-        public RetrySettings WithTotalExpiration(Expiration expiration) =>
-            new RetrySettings(RetryBackoff, TimeoutBackoff, expiration, RetryFilter, DelayJitter);
+        /// <param name="currentBackoff">The current backoff to use as a basis for the next one.</param>
+        /// <returns>The next backoff to use, which is always at least <see cref="InitialBackoff"/> and at most <see cref="MaxBackoff"/>.</returns>
+        public TimeSpan NextBackoff(TimeSpan currentBackoff)
```
amanda-tarafa: A little late to the party, but I would expect this method to use whatever jitter is set.
I know the jitter is being applied when retrying, but now that the backoff configuration and the jitter are defined on the same class, it's just unexpected that this doesn't use the jitter.
I would at least add a comment here.
And I would probably actually add a new property `InitialJitteredBackoff` and a new method `NextJitteredBackoff`. The retry code would be slightly simpler with these new members.

jskeet: I don't think it would make sense to have extra properties (what would `InitialJitteredBackoff` mean?) - but a few options:

- Just add documentation
- Add an `out` parameter for the "jittered backoff"
- Write a method that encapsulates all of this by returning an `IEnumerable<TimeSpan>` that is "actual backoffs to apply" and which keeps track of the multiplied version

Thoughts?

amanda-tarafa: I really like option 3, as long as the existing exposed methods and properties are not removed so that client code can use the retry config on other delays that were not generated in the sequence (for instance if the server recommends a delay (I know this doesn't happen now)).
Options 2 is good enough, only problem would be that the initial backoff still needs to be jittered "manually" (that's what the InitialJitteredBackoff property does).

jskeet: How about option 3, but removing the existing method?

For option 2, I'd envisaged it being "the out parameter is how much you should delay right now, and the return value is what should be passed in next time". So there wouldn't be any need for manual jittering.

chrisdunelm: I like option 3 (sequence) with existing method removed. We can always add it back if a use-case for it arises.

PR #369, opened 2019-12-02
DO NOT MERGE - RetrySettings possible change

jskeet:


jskeet:
I've left NextBackoff present but internal for now. (It has a test elsewhere - I haven't added any tests for the new method yet.)
If we can work out the API surface we want to expose, we can then do the rest.

The reason for the parameter (that might be unexpected) is that we use this in Firestore - it feels useful to be able to say "retry immediately before you actually start delaying".

Google.Api.Gax.Grpc/RetrySettings.cs
```diff
@@ -179,6 +181,24 @@ public TimeSpan NextBackoff(TimeSpan currentBackoff)
...
 
+        /// <summary>
+        /// Returns a sequence of already-jittered backoffs. A caller can just iterate over this
+        /// sequence and delay by each returned timespan. This sequence is infinite; callers do not
+        /// need to test the result of MoveNext.
+        /// </summary>
+        /// <param name="initialBackoffOverride">The initial backoff to apply, or null to use <see cref="InitialBackoff"/>;
+        /// this can be used to retry immediately after an initial failure, or to start with a longer-than-normal backoff.</param>
+        /// <returns>A sequence of jittered backoffs, suitable for delays.</returns>
+        public IEnumerator<TimeSpan> GetJitteredBackoffSequence(TimeSpan? initialBackoffOverride = null)
```
chrisdunelm: Why not make this an `IEnumerable<TimeSpan>`, then the code in `WithRetry()` can use a `foreach` loop over it, which would be nice.

jskeet: We *could* do that, I suppose. It would be slightly odd to have "the backoff that I'll try if it fails" in scope the whole time. Will add a second commit to this PR to show what it would look like.

Google.Api.Gax.Grpc/RetrySettings.cs
```diff
@@ -179,6 +181,24 @@ public TimeSpan NextBackoff(TimeSpan currentBackoff)
             }
         }
 
+        /// <summary>
+        /// Returns a sequence of already-jittered backoffs. A caller can just iterate over this
+        /// sequence and delay by each returned timespan. This sequence is infinite; callers do not
```
amanda-tarafa: Couldn't the sequence respect MaxAttempts value and be finite? If done with a foreach as Chris suggested, the calling code would be simpler I think.

Google.Api.Gax.Grpc/RetrySettings.cs
amanda-tarafa: +1

Google.Api.Gax.Grpc/RetrySettings.cs
jskeet: Unfortunately that's pretty hard - because when we fall off the end, we really want to just not catch the exception. (We don't want to catch it and then rethrow it later, losing a bunch of context, IMO.)
I do see the appeal, but the nature of exceptions is working against us :(

amanda-tarafa: True, to keep context we would still have to keep track of the attempts on the calling code. And I do agree that context is more important here.

Google.Api.Gax.Grpc/RetrySettings.cs
jskeet: I've pushed the new commit now - PTAL. If we don't mind "simple rethrow of an exception" rather than "rethrow preserving stack trace", we probably *could* bake MaxAttempts in there...

Google.Api.Gax.Grpc/RetrySettings.cs
jskeet: I'm now starting to think that there's so little in the loop, we can probably put an "execute with retry" call into RetrySettings after all. The situation is simpler than it was. Hmm.

chrisdunelm: So far, we've had an informal agreement to have no extra public methods in "settings" classes - they are pure data. This would break that. Not that we can't do it, but we should consciously do it if we choose to.

To continue the foreach/max-attempts idea:
If the foreach loop  looped over `retrySettings.GetJitteredBackoffSequence().Append(TimeSpan.Zero)`, then the `catch when` condition can include checking that, and it's all nice and elegant :) (In my opinion anyway!)

jskeet: Will keep noodling on this. I suspect there's stuff we can do here :)

amanda-tarafa: What if the the sequence is not of TimeSpan but of RetryAttempt (or whatever other name that's better)? And
```
IEnumerable<RetryAttempt> GetJitteredBackoffSequence(Expiration overallDeadline, IScheduller, IClock);
class RetryAttempt
{    
    bool CanRetry(RpcException e) {// MaxAttempts, Filter, OverallDeadline all taken into account} // 
    Task Delay(); // Throws if above was false when fisrt called (because of OverallDeadline)
}

```
The last element of the sequence is the first one that returns false on `CanRetry`.

If we could `catch ... when await ...` that'd be great, then we could have only one method Task<bool> TryDelay(RpcException) :)

jskeet: Will probably come back to all of this next week. I like that idea, but want to get other stuff done first :)

jskeet:
Closing in favour of #391.

PR #370, opened 2019-12-11
Build both net461 and netstandard2.0 on all platforms

jskeet:
This uses Microsoft.NETFramework.ReferenceAssemblies when on non-Windows platforms.

Note that we still need conditions in test projects - we can't run .NET Framework tests other than on Windows.

### DIFF
--- a/Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
+++ b/Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
@@ -5,7 +5,6 @@
   <PropertyGroup>
     <Version>2.0.0-alpha00</Version>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard2.0</TargetFrameworks>
   </PropertyGroup>
 
   <!-- Packaging information -->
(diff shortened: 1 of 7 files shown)


jskeet:
I'm going to run dotnet pack on my Linux box and check that it does what I expect.

jskeet:
Seems to work! We'll need to do a bit more testing than that, but this could be great. We'll want to roll it out everywhere else too, of course.

Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
```diff
@@ -18,5 +17,6 @@
     <PackageReference Include="Google.Protobuf" Version="3.8.0" />
     
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.3" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers" />
```
chrisdunelm: I was expecting this to be conditional, to only be included when building on Linux?

jskeet: That was my initial expectation too, but this is basically the XML from the NuGet package page. I believe the point is that it's basically a no-op on Windows.

PR #371, opened 2019-12-12
Update common resource names, and IResourceName

chrisdunelm:


chrisdunelm:
@jskeet Do we still need `TemplatedResourceName`?

Google.Api.Gax/IResourceName.cs
```diff
@@ -13,9 +13,12 @@ namespace Google.Api.Gax
...
     {
         /// <summary>
-        /// The kind of resource name.
+        /// Whether this instance contains a known resource name.
         /// </summary>
-        ResourceNameKind Kind { get; }
+        /// <remarks>
+        /// A "known resource name" is one which conforms to one of the pre-defined patterns within this class.
+        /// </remarks>
+        bool IsKnown { get; }
```
jskeet: How about IsKnownPattern? If we have an "unknown" resource name, we still know the resource name - just not the pattern it conforms to.

chrisdunelm: Yes, I agree that's better. I'll change it now.

PR #372, opened 2019-12-12
Updated to use UnparsedResourceName

chrisdunelm:


Google.Api.Gax/UnparsedResourceName.cs
```diff
@@ -12,30 +12,30 @@ namespace Google.Api.Gax
     /// <summary>
     /// A resource name in which nothing is known about the name structure.
     /// </summary>
-    public sealed class UnknownResourceName : IResourceName, IEquatable<UnknownResourceName>
+    public sealed class UnparsedResourceName : IResourceName, IEquatable<UnparsedResourceName>
     {
         /// <summary>
-        /// Parse a resource name into an <see cref="UnknownResourceName"/>.
+        /// Parse a resource name into an <see cref="UnparsedResourceName"/>.
```
jskeet: This sentence is kinda funky - "We're parsing it but we're not really." I think it's okay though.

Google.Api.Gax/UnparsedResourceName.cs
```diff
@@ -12,30 +12,30 @@ namespace Google.Api.Gax
...
         /// Only minimal verification is carried out that <paramref name="name"/> is a value resource name string.
```
jskeet: Nit: value => valid, while we're editing this file?

chrisdunelm: done

Google.Api.Gax/UnparsedResourceName.cs
chrisdunelm: ack

PR #373, opened 2019-12-17
DO NOT MERGE: Prototype of "how to depend just on Grpc.Core.Api"

jskeet:
We'd need to work out how a "default" channel factory would be supplied, and what this would mean for client library packaging.

### DIFF
--- a/Google.Api.Gax.Grpc/ChannelFactory.cs
+++ b/Google.Api.Gax.Grpc/ChannelFactory.cs
@@ -1,0 +1,21 @@
+﻿/*
+ * Copyright 2019 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+using Grpc.Core;
+
+namespace Google.Api.Gax.Grpc
+{
+    // TODO: Make this an abstract class instead of a delegate?
+
+    /// <summary>
+    /// Creates a channel for the given endpoint, with the given credentials and options.
+    /// </summary>
+    /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
+    /// <param name="credentials">Credentials for the channel. Must not be null.</param>
+    /// <param name="options">Options for the channel. Must not be null.</param>
+    /// <returns>A channel for the given endpoint.</returns>
+    public delegate ChannelBase ChannelFactory(ServiceEndpoint endpoint, ChannelCredentials credentials, GrpcChannelOptions options);
+}
--- a/Google.Api.Gax.Grpc/ChannelPool.cs
+++ b/Google.Api.Gax.Grpc/ChannelPool.cs
@@ -22,12 +22,10 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class ChannelPool
     {
-        private static readonly IReadOnlyList<ChannelOption> s_defaultOptions = new List<ChannelOption>
-        {
-            GrpcChannelOptions.OneMinuteKeepalive,
-            GrpcChannelOptions.DisableServiceConfigResolution
-        };
-        
+        private static readonly GrpcChannelOptions s_defaultOptions = GrpcChannelOptions.FromKeepAlive(TimeSpan.FromMinutes(1))
+            .MergedWith(GrpcChannelOptions.FromEnableServiceConfigResolution(false));
+
+        private readonly ChannelFactory _channelFactory;
         private readonly IEnumerable<string> _scopes;
 
         /// <summary>
@@ -41,16 +39,18 @@ namespace Google.Api.Gax.Grpc
 
         // TODO: See if we could use ConcurrentDictionary instead of locking. I suspect the issue would be making an atomic
         // "clear and fetch values" for shutdown.
-        private readonly Dictionary<Key, Channel> _channels = new Dictionary<Key, Channel>();
+        private readonly Dictionary<Key, ChannelBase> _channels = new Dictionary<Key, ChannelBase>();
         private readonly object _lock = new object();
 
         /// <summary>
         /// Creates a channel pool which will apply the specified scopes to the default application credentials
         /// if they require any.
         /// </summary>
+        /// <param name="channelFactory">The channel factory to use, or null for the default channel factory.</param>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        public ChannelPool(ChannelFactory channelFactory, IEnumerable<string> scopes)
         {
+            _channelFactory = channelFactory;
             // Always take a copy of the provided scopes, then check the copy doesn't contain any nulls.
             _scopes = GaxPreconditions.CheckNotNull(scopes, nameof(scopes)).ToList();
             GaxPreconditions.CheckArgument(!_scopes.Any(x => x == null), nameof(scopes), "Scopes must not contain any null references");
@@ -76,14 +76,17 @@ namespace Google.Api.Gax.Grpc
         /// <returns>A task which will complete when all the (current) channels have been shut down.</returns>
         public Task ShutdownChannelsAsync()
         {
-            List<Channel> channelsToShutdown;
+            List<ChannelBase> channelsToShutdown;
             lock (_lock)
             {
                 channelsToShutdown = _channels.Values.ToList();
                 _channels.Clear();
             }
+            // FIXME: Reinstate this
+            return Task.CompletedTask;
+            /*
             var shutdownTasks = channelsToShutdown.Select(c => c.ShutdownAsync());
-            return Task.WhenAll(shutdownTasks);
+            return Task.WhenAll(shutdownTasks);*/
         }
 
         /// <summary>
@@ -92,7 +95,7 @@ namespace Google.Api.Gax.Grpc
         /// </summary>
         /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
         /// <returns>A channel for the specified endpoint.</returns>
-        public Channel GetChannel(ServiceEndpoint endpoint) => GetChannel(endpoint, s_defaultOptions);
+        public ChannelBase GetChannel(ServiceEndpoint endpoint) => GetChannel(endpoint, s_defaultOptions);
 
         /// <summary>
         /// Asynchronously returns a channel from this pool, creating a new one if there is no channel
@@ -101,7 +104,7 @@ namespace Google.Api.Gax.Grpc
         /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
         /// <returns>A task representing the asynchronous operation. The value of the completed
         /// task will be channel for the specified endpoint.</returns>
-        public Task<Channel> GetChannelAsync(ServiceEndpoint endpoint) => GetChannelAsync(endpoint, s_defaultOptions);
+        public Task<ChannelBase> GetChannelAsync(ServiceEndpoint endpoint) => GetChannelAsync(endpoint, s_defaultOptions);
 
         /// <summary>
         /// Returns a channel from this pool, creating a new one if there is no channel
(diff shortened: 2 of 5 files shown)


Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -76,14 +76,17 @@ private async Task<ChannelCredentials> CreateChannelCredentialsUncached()
...
             lock (_lock)
             {
                 channelsToShutdown = _channels.Values.ToList();
                 _channels.Clear();
             }
+            // FIXME: Reinstate this
+            return Task.CompletedTask;
+            /*
             var shutdownTasks = channelsToShutdown.Select(c => c.ShutdownAsync());
-            return Task.WhenAll(shutdownTasks);
```
jtattermusch: a bit more detail on shutting down ChannelBase:

- Channel (from Grpc.Core) has ShutdownAsync() method ( https://github.com/grpc/grpc/blob/528dbb13e400654f9c73d8e26e70be48f27052c4/src/csharp/Grpc.Core/Channel.cs#L224)
- GrpcChannel (from grpc-dotnet) has `Dispose()` method (which disposes the underlying HttpClient if configured so https://github.com/grpc/grpc-dotnet/blob/36e42811e7bf85c8aa765813c82fab3dcfeb8297/src/Grpc.Net.Client/GrpcChannel.cs#L270)
- ChannelBase does not expose either of those. I think we could add either ShutdownAsync() or Dispose() to ChannelBase to provide some way of signaling that the channel is no longer going to be used.

The managed GrpcChannel is a thin wrapper on top of HttpClient. I wasn't able to find much info on how important it is to Dispose() HttpClient when done with it.

I filed https://github.com/grpc/grpc/issues/21612 to track this.

Google.Api.Gax.Grpc/ChannelFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+    // TODO: Make this an abstract class instead of a delegate?
+
+    /// <summary>
+    /// Creates a channel for the given endpoint, with the given credentials and options.
+    /// </summary>
+    /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
+    /// <param name="credentials">Credentials for the channel. Must not be null.</param>
+    /// <param name="options">Options for the channel. Must not be null.</param>
+    /// <returns>A channel for the given endpoint.</returns>
+    public delegate ChannelBase ChannelFactory(ServiceEndpoint endpoint, ChannelCredentials credentials, GrpcChannelOptions options);
```
jtattermusch: I assume both Grpc.Core and grpc-dotnet would provide an implementation of such factory that creates channels of the right kind. If that's so and the channelFactory implementation is the only piece that would depend on Grpc.Core or Grpc.Net.Client, it should at least be trivial to generate two versions of the nuget:
- one version depends on Grpc.Core package
- the other version depends on Grpc.Net.Client
- they are exactly the same except one file that provides the right channelFactory implementation (there could even be an #ifdef that allows one to switch by defining a symbol)

This is still not an ideal soution of course, but at least it would make it reasonably easy to obtain packages that only depend on Grpc.Net.Client

jtattermusch: I think if we were able to generate a set of client libraries that depended on Grpc.Net.Client instead of Grpc.Core (and would be able to upload them to dev nuget channel), it would be huge for making sure that one day, the managed grpc client will work with google APIs.
(having a package to play with makes it much easier to see what features are missing a what bugs need to be fixed).

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -22,12 +22,10 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class ChannelPool
```
hazzik: I was doing exactly the same exercise today (trying to make gax.grpc a managed library). What I found is that ChannelPool and ClientBuilderBase are not used anywhere except tests. My naive implementation was to introduce 2 additional libraries. One will be called ` Google.Api.Gax.Grpc.Api` (by the analogy of how Grpc.Core was split) which will pool all managed code from the `Google.Api.Gax.Grpc`, another one is Google.Api.Gax.Grpc.AspNetCore which will contain managed client code.

Google.Api.Gax.Grpc/ChannelFactory.cs
hazzik: Actually only 2 pieces which depend on the managed/native part are `ChannelPool` and `ClientBuilderBase<T>` if they both moved to a separate package it would solve all your questions and concerns (in particular: `Client.ShutdownAsync` vs `Client.Dispose`, unsupported client options (keepalive), etc),

hazzik: I doubt that majority of the "end" libraries would need to have a ClientBuilder or a ChannelPool. So, I think it is _a good enough_ solution.

Google.Api.Gax.Grpc/ChannelPool.cs
hazzik: What if implement the same solution as for instantiation of the client - pass a delegate do shutdown the unwanted clients?

Google.Api.Gax.Grpc/ChannelFactory.cs
jskeet: The GAX libraries are specifically designed to support the libraries at https://github.com/googleapis/google-cloud-dotnet, and *every* gRPC-based library there uses `ClientBuilder`; almost all of them use `ChannelPool`. So no, it doesn't really solve much.
If you know of other libraries using GAX, I'd be interested to hear about them.

Google.Api.Gax.Grpc/ChannelPool.cs
jskeet: That would definitely be possible. We could potentially isolate everything into one abstract class. It all feels somewhat weird, admittedly.
The bigger issue we need to address is what the final set of dependencies is. Will write more about that later.

Google.Api.Gax.Grpc/ChannelFactory.cs
jskeet: More details about the deployment issue. The absolute (and unachievable as far as I'm aware) ideal that I'd like is to have something which is effectively "default to Grpc.Core, but allow an opt out somehow". Basically, there has to be a dependency on Grpc.Core somewhere, but the question is where.

**Option 1: Bias heavily towards Grpc.Core**

Google.Cloud.Speech.V1 depends on Grpc.Core (as it currently does, both directly and indirectly). It tells GAX that it wants to use the Grpc.Core-based way of creating clients and shutting down channels. That leaves two problems if you want to use Grpc.Net.Client instead:

- Grpc.Core (which is large) will *always* get deployed alongside the application, even if you're not using it.
- There has to be some way of telling Google.Cloud.Speech.V1 to use the different abstraction. It's not obvious where that injection point should be.

**Option 2: Three packages per API**

An alternative would be to make Google.Cloud.Speech.V1 only depend on GAX (and indirectly Grpc.Core.Api), and to produce two *extra* packages: Google.Cloud.Speech.V1.GrpcCore and Google.Cloud.Speech.V1.GrpcNetClient. It's still not obvious to me what the API surface looks like for that, and it means tripling the number of packages we produce and maintain, which isn't ideal either.

**Option 3: Two packages per API**

This is basically Jan's suggestion above. It involves code duplication (there being two different SpeechClient classes, for example - one in each flavour of package) but does reduce the dependency chain somewhat.

If the purpose of using Grpc.Net.Client is primarily to validate that we have compatibility, then we should probably go for option 1 listed above - the unnecessary "baggage" of Grpc.Core probably isn't much of an issue for that, although we still need to work out the injection point for the abstraction. (A static property would be feasible but unpleasant, for example.)

I don't have a good answer to all of this yet, but that's a summary of the problem I'm trying to get my head round.

Google.Api.Gax.Grpc/ChannelPool.cs
jskeet: See the comment on ChannelFactory for deployment thoughts.

Google.Api.Gax.Grpc/ChannelFactory.cs
jtattermusch: I think for now we mostly want to:
- validate that Grpc.Net.Client works for Google API (and see what doesn't work)
- we already know that at least in some cases Grpc.Net.Client wont works because it doesn't support all the necessary features (e.g. some channel options are not implemented).
- there's some potential that Grpc.Net.Client might become the default choice one day, but it's likely a distant future and not clear if it will ever happen at all.

So I'd say we can go with Option 1 for now, assuming we can provide a simple way of using Grpc.Net.Client with the client libraries for those who want to opt-in (for testing, validation or to work on filling in the gaps in Grpc.Net.Client).
In addition to that, we could provide
- a HOWTO doc for using client libraries with Grpc.Net.Client (and list of gotchas)
- a small support nuget package that provides the Grpc.Net.Client binding in a digestible form (to make using it as simple as possible).

WDYT?

CC @JamesNK

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -41,16 +39,18 @@ public sealed class ChannelPool
...
         private readonly object _lock = new object();
 
         /// <summary>
         /// Creates a channel pool which will apply the specified scopes to the default application credentials
         /// if they require any.
         /// </summary>
+        /// <param name="channelFactory">The channel factory to use, or null for the default channel factory.</param>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        public ChannelPool(ChannelFactory channelFactory, IEnumerable<string> scopes)
```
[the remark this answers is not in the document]
jskeet: You don't need to create your own ChannelPool in order to test client libraries - the channel pool is only used by default; you can specify your own channel or CallInvoker via a builder when creating the client. The tricky part about changing the default is that the channel pool is a static field within the client type - intercepting that would be difficult. (We could investigate some options though...)

Google.Api.Gax.Grpc/ChannelFactory.cs
jtattermusch: One (dirty and hacky) idea on eliminating the Grpc.Core download:
- the hypothetical Google.Api.Gax.Grpc.ManagedClientBinding" package (which provides Grpc.Net.Client binding) also has Grpc.Core dependency, but uses some ridiculous fake version of the package so that the dependency is resolved to that version. That way, a fake version of Grpc.Core is downloaded instead of the real heavyweight package (it could contain a gutted version of the Grpc.Core assembly that does nothing). This way the Grpc.Core dependency stops being a problem.

@jskeet  WDYT?

jskeet: I thought I'd suggested that a while ago and it was rejected? It's definitely an option, but it would screw over anyone who also wanted Grpc.Core for some reason.
Let's keep thinking.

jskeet: I'll experiment with it - now is definitely the time to do so. I'm torn between "let this package depend on Grpc.Core to streamline the normal case" and "provide a Google.Api.Gax.Grpc.GrpcCore package that has the Grpc.Core implementation of the abstraction layer". The latter is sort of cleaner, but requires yet another package to maintain. I think I'll need to play a bit more.

jskeet: I'm now thinking of a *possible* option of loading things with reflection, so that we can avoid extra packages. It may make things tricky in terms of package trimming, but it's at least worth experimenting with.

Google.Api.Gax.Grpc/ChannelPool.cs
JamesNK: I think that adding ShutdownAsync to ChannelBase is fine. When I added ChannelBase I considered adding some of Channel's methods to it (state, connect, shutdown) but since we didn't have a reason for them I left ChannelBase bare-bones.

What you can do:
* Move ShutdownAsync from Channel to ChannelBase
* Add virtual ShutdownAsyncCore to ChannelBase. By default it does nothing.
* Override ShutdownAsyncCore on Channel to run its current ShutdownAsync logic.
* Override ShutdownAsyncCore on GrpcChannel to call HttpClient.Dispose.

jskeet:
Closing in favour of #383.

Issue #368, opened 2019-11-27
Update Google.Api.CommonProtos generated code


JamesNK:
Google.Protobuf has a new API for getting custom options and the types in Google.Api.CommonProtos don't support it because the generated code is from an older version of protoc.

https://github.com/protocolbuffers/protobuf/issues/6956#issuecomment-559212920

jskeet:
Will need to check - some of the more recent Google.Protobuf releases have had issues; if that's been fixed, I'm happy to regenerate.

jskeet:
Right, it looks like the latest GA release is still 3.10.1 which I know has some bugs in (as I fixed them :)
It looks like 3.11.0 should be out soon - at that point I'll update the generator we use (and our dependency).

It will be somewhat awkward to do another release before the start of 2020, however - we're in the middle of planning a large new major version release (prompted by the gRPC major version bump). We definitely *can* do another release, via creating a new branch specifically for it, based on the older version, but I'd *prefer* not to if possible. What's the urgency of this?

JamesNK:
It's not urgent. I'm playing around with a prototype.

The current version of Google.Protobuf seems pretty broken around custom options. I could force the old version, or regenerate `http.proto` and ship it with the library.

jskeet:
From what I remember, you should be able to access the custom options using 3.8.0 (which is what Google.Api.CommonProtos targets), but you need to know the option field number first, and it's awkward. If you'd like me to put together an example of doing it (awkwardly but plausibly) for the current library, let me know.

JamesNK:
It's fine. I'm already doing it with 3.8.0.

```cs
const int HttpRuleFieldId = 72295728;
var success = methodDescriptor.CustomOptions.TryGetMessage<HttpRule>(HttpRuleFieldId, out httpRule);
```

I just want to make sure that there is a solution for new Google.Protobuf versions.

A question I do have is is there a way to support `google.api.http` without requiring people to worry about placing `google/api/annotations.proto` and `google/api/http.proto` at the right location in their project? Basically I want importing those files to Just Work like `import "google/protobuf/empty.proto";` does.

This is for a gRPC projects and the compiler is called by Grpc.Tools and msbuild. I'm guessing I could probably do some cleaver tricks with msbuild to make the protoc compiler happy by automatically supplying the extra proto files during build.

Any ideas or suggestions?

jskeet:
[...]

PR #377, opened 2020-01-13
Update protoc and regenerate common protos

jskeet:
Note that this doesn't yet update the version of gRPC tools that we're using, but we can do that later.

Note that a couple of proto messages are deleted as part of this, which is fine as we'll be releasing a 2.0.0-beta01 when we're ready.
This is part of addressing #368.


Google.Api.CommonProtos/Type/DateExtensions.cs
```diff
@@ -15,11 +15,11 @@ namespace Google.Type
     public static class DateExtensions
     {
         /// <summary>
-        /// Converts the <see cref="DateTime.Date"/> part of <see cref="DateTime"/> to <see cref="Date"/>.
+        /// Converts the <see cref="System.DateTime.Date"/> part of <see cref="System.DateTime"/> to <see cref="Date"/>.
         /// </summary>
         /// <param name="dateTime">The <see cref="DateTime"/> instance being converted.</param>
```
chrisdunelm: Does this `see` reference also need to be `System.DateTime`?

Google.Api.CommonProtos/Type/DatePartial.cs
```diff
@@ -14,13 +14,13 @@ public partial class Date
         /// <summary>
         /// Converts <see cref="Date"/> to <see cref="DateTime"/>.
         /// </summary>
-        /// <returns>The converted <see cref="DateTime"/> with time at midnight and <see cref="DateTime.Kind"/> of <see cref="DateTimeKind.Unspecified"/>.</returns>
+        /// <returns>The converted <see cref="DateTime"/> with time at midnight and <see cref="System.DateTime.Kind"/> of <see cref="DateTimeKind.Unspecified"/>.</returns>
```
chrisdunelm: `System.DateTime` in the first `see` here too?

Google.Api.CommonProtos/Type/DatePartial.cs
```diff
@@ -14,13 +14,13 @@ public partial class Date
         /// <summary>
         /// Converts <see cref="Date"/> to <see cref="DateTime"/>.
```
chrisdunelm: `System.DateTime`?

jskeet:
Well spotted - thanks, fixed those three.

PR #381, opened 2020-01-21
Removes ServiceEndpoint from GAX

### LINKED ISSUE #379: Consider removing ServiceEndpoint in 2.x
It's somewhat late in the day to do this, but I want to at least investigate the idea of removing ServiceEndpoint.

This is motivated by #357, and the [gRPC documentation](https://grpc.github.io/grpc/cpp/md_doc_naming.html). We really don't want to have to mimic all the different available schemes in gRPC, within ServiceEndpoint - we'd always be playing catchup.

For now, I'll document what would be required to remove ServiceEndpoint, then we can consider it in more detail.

jskeet:
Fixes #379. We'd then need to change the microgenerator accordingly.


jskeet:
This part is simple - I'm now going to try using this branch from the next-major-version of google-cloud-dotnet, regenerating and testing Spanner and Firestore as "reasonably complex APIs".

jskeet:
Woot. Will make sure it all works end to end before merging this - it could be a few days...

jskeet:
Okay, I've got it working end-to-end for Firestore, which is enough to prove to me that it'll be okay. Merging, then will cascade the changes down...

MaikuMori:
This breaks `Google.Cloud.AutoML.V1 Version="1.0.0"`, any plan to release betas for the other packages which depend on this?

jskeet:
@MaikuMori: We've released new libraries for everything except AutoML, where we're still waiting on some proto annotations. I'm hoping I'll be able to publish an AutoML beta this coming week. Until then, you'll need to stick with GAX 2.10.

PR #383, opened 2020-01-24
Make gRPC library implementation-neutral in API surface

jskeet:
This still depends on Grpc.Core and has a default implementation (embedded in it) that uses Grpc.Core, but the public API surface only depends on Grpc.Core.Api.

Still to come in later commits:

- Doc comments on GrpcChannelOptions
- Tests for GrpcChannelOptions
- Tests for GrpcCoreChannelFactory
- Decisions around how to handle duplication of custom options (either with each other or with standard options)
- A proof-of-concept for a Grpc.Net.Client implementation.

### DIFF
--- a/Google.Api.Gax.Grpc/ChannelPool.cs
+++ b/Google.Api.Gax.Grpc/ChannelPool.cs
@@ -22,12 +22,11 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class ChannelPool
     {
-        private static readonly IReadOnlyList<ChannelOption> s_defaultOptions = new List<ChannelOption>
-        {
-            GrpcChannelOptions.OneMinuteKeepalive,
-            GrpcChannelOptions.DisableServiceConfigResolution
-        };
-        
+        private static readonly GrpcChannelOptions s_defaultOptions = GrpcChannelOptions.Empty
+            .WithKeepAliveTime(TimeSpan.FromMinutes(1))
+            .WithEnableServiceConfigResolution(false);
+
+        private readonly GrpcImplementation _grpcImplementation;
         private readonly IEnumerable<string> _scopes;
 
         /// <summary>
@@ -41,7 +40,7 @@ namespace Google.Api.Gax.Grpc
 
         // TODO: See if we could use ConcurrentDictionary instead of locking. I suspect the issue would be making an atomic
         // "clear and fetch values" for shutdown.
-        private readonly Dictionary<Key, Channel> _channels = new Dictionary<Key, Channel>();
+        private readonly Dictionary<Key, ChannelBase> _channels = new Dictionary<Key, ChannelBase>();
         private readonly object _lock = new object();
 
         /// <summary>
@@ -49,14 +48,25 @@ namespace Google.Api.Gax.Grpc
         /// if they require any.
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        /// <param name="grpcImplementation">The gRPC implementation to use, or null for the default implementation.</param>
+        internal ChannelPool(IEnumerable<string> scopes, GrpcImplementation grpcImplementation = null)
         {
+            _grpcImplementation = grpcImplementation ?? GrpcImplementation.Default;
             // Always take a copy of the provided scopes, then check the copy doesn't contain any nulls.
             _scopes = GaxPreconditions.CheckNotNull(scopes, nameof(scopes)).ToList();
             GaxPreconditions.CheckArgument(!_scopes.Any(x => x == null), nameof(scopes), "Scopes must not contain any null references");
             // In theory, we don't actually need to store the scopes as field in this class. We could capture a local variable here.
             // However, it won't be any more efficient, and having the scopes easily available when debugging could be handy.
             _lazyScopedDefaultChannelCredentials = new Lazy<Task<ChannelCredentials>>(() => Task.Run(CreateChannelCredentialsUncached));
+        }
+
+        /// <summary>
+        /// Creates a channel pool which will apply the specified scopes to the default application credentials
+        /// if they require any.
+        /// </summary>
+        /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
+        public ChannelPool(IEnumerable<string> scopes) : this(scopes, null)
+        {
         }
 
         private async Task<ChannelCredentials> CreateChannelCredentialsUncached()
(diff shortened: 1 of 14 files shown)


jskeet:
cc @jtattermusch

Google.Api.Gax.Grpc/GrpcCoreChannelFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+        internal ChannelBase CreateChannel(string endpoint, ChannelCredentials credentials, GrpcChannelOptions options) =>
+            // TODO: Cache the result of ConvertOptions
+            new Channel(endpoint, credentials, ConvertOptions(options));
+
+        private List<ChannelOption> ConvertOptions(GrpcChannelOptions options)
+        {
+            List<ChannelOption> ret = new List<ChannelOption>();
+            if (options.EnableServiceConfigResolution is bool enableServiceConfigResolution)
+            {
+                ret.Add(new ChannelOption("grpc.service_config_disable_resolution", enableServiceConfigResolution ? 1 : 0));
```
jtattermusch: is the 0 : 1 reversed here? (disable vs enable semantics)

Google.Api.Gax.Grpc/GrpcChannelOptions.cs
```diff
@@ -1,45 +1,282 @@
...
+
+        /// <summary>
+        /// If non-null, explicitly enables or disables service configuration resolution.
+        /// </summary>
+        public bool? EnableServiceConfigResolution { get; }
+
+        /// <summary>
+        /// If non-null, explicitly specifies the keep-alive period for the channel.
+        /// </summary>
+        public TimeSpan? KeepAlive { get; }
```
jtattermusch: nit: note that grpc C core supports mutliple keepalive-related options:
https://github.com/grpc/grpc/blob/5791495f595b30545f3dfd31321f4b5760b11c75/include/grpc/impl/codegen/grpc_types.h#L227
so let's make sure that the name "KeepAlive" isn't too general in case some of the other options are needed later.

Google.Api.Gax.Grpc/GrpcCoreChannelFactory.cs
jskeet: Whoops - that's just me failing. Fixing.

Google.Api.Gax.Grpc/GrpcChannelOptions.cs
jskeet: Good call. Will change to KeepAliveTime for now, to match the C-core name. We'll probably want to revisit all the naming though.

jskeet:
Ping for review, @chrisdunelm and @amanda-tarafa if possible.

Google.Api.Gax.Grpc/GrpcChannelOptions.cs
```diff
@@ -1,45 +1,282 @@
...
+            MergedWith(new GrpcChannelOptions(enableServiceConfigResolution: enableServiceConfigResolution));
+
         /// <summary>
-        /// "After a duration of this time the client/server pings its peer to see if the
-        /// transport is still alive. Int valued, milliseconds."
-        /// Required for any channel using a streaming RPC, to ensure an idle stream doesn't
-        /// allow the TCP connection to be silently dropped by any intermediary network devices.
-        /// 60 second keepalive time is reasonable. This will only add minimal network traffic,
-        /// and only if the channel is idle for more than 60 seconds.
+        ///
```
chrisdunelm: Is `summary` deliberately empty for all these `With` methods?

Google.Api.Gax.Grpc/GrpcChannelOptions.cs
```diff
@@ -1,45 +1,282 @@
...
-        /// allow the TCP connection to be silently dropped by any intermediary network devices.
-        /// 60 second keepalive time is reasonable. This will only add minimal network traffic,
-        /// and only if the channel is idle for more than 60 seconds.
+        /// 
         /// </summary>
-        internal static ChannelOption OneMinuteKeepalive { get; } = new ChannelOption("grpc.keepalive_time_ms", 60_000);
+        /// <param name="keepAliveTime"></param>
+        /// <returns></returns>
+        public GrpcChannelOptions WithKeepAliveTime(TimeSpan keepAliveTime) =>
+            new GrpcChannelOptions(keepAliveTime: keepAliveTime);
```
chrisdunelm: Should this be `MergedWith(...)`?

jskeet: Doh! Yes, thanks :)

Google.Api.Gax.Grpc/GrpcChannelOptions.cs
jskeet: For this commit, yes. I didn't want to write a load of docs and then have to change them.

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -41,16 +40,18 @@ public sealed class ChannelPool
...
         private readonly object _lock = new object();
 
         /// <summary>
         /// Creates a channel pool which will apply the specified scopes to the default application credentials
         /// if they require any.
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        /// <param name="grpcImplementation">The gRPC implementation to use, or null for the default implementation.</param>
+        internal ChannelPool(IEnumerable<string> scopes, GrpcImplementation grpcImplementation = null)
```
amanda-tarafa: Should this one be public now?

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -74,6 +72,11 @@ public abstract class ClientBuilderBase<TClient>
         /// </summary>
         public string UserAgent { get; set; }
 
+        /// <summary>
+        /// The gRPC implementation to use, or null to use the default channel factory.
+        /// </summary>
+        internal GrpcImplementation GrpcImplementation { get; set; }
```
amanda-tarafa: I believe `protected abstract GetChannelPool()` might need to use this if it ends up creating the pool, so this one should probably be `protected internal`?
I see that GrpcImplementation is still internal by design, but still the possible discrepancy between this one here and the one in ChannelPool, I don't like.

I wonder if it would be possible to have something like GrpcImplementation.Current, that falls back to GrpcImplementation.Default, but that can be set on startup (and maybe even just once) so that everything can refer back to that and we don't need to be passing a GrpcImplementation around?

Google.Api.Gax.Grpc/ChannelPool.cs
amanda-tarafa: Ah I see now that everything is still internal by desing, so it's fine.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: The intention is that the public API doesn't need to change right now. I'm not sure what you mean by "the possible discrepancy between this one here and the one in ChannelPool" - the idea is that they both go public at the same time, when we're slightly more confident about the shape. (We *could* make it public right now, but I don't think we need to just yet.)

amanda-tarafa: Yes, I'm (mostly*) fine with them not being public right now. What I meant by "the possible discrepancy..." is that, as they stand, for any given execution, this one could bet set to one implementation and the one in `ChannelPool` to another implementation which I think is not needed/wanted and might be problematic. That's why I was thinking of having a `static` `GrpcImplementation.Current` (as in current execution implementation) that both client builders and `ChannelPool` could use, without the need of passing a `GrpcImplementation` around. `GrpcImplementation.Current` can be set on startup (can be set only once probably) and if not set would fallback to `GrpcImplementation.Default`.

\* Mostly fine because of the `protected abstract GetChannelPool()` method in `ClientBuilderBase` should be able to access `ClientBuilderBase.GrpcImplementation` in case it ends up building a channel. But given that everything is internal this is just a nit.

jskeet:
Right, I've now added tests and docs. This is ready for a proper review.

Issue #347, opened 2019-08-12
Support Stackdriver Kubernetes Engine Monitoring


jskeet:
Currently we create monitored resource types of either "container" or "gke_container" (pre/post #346).
Those are for *legacy* Stackdriver; to support Stackdriver Kubernetes Engine Monitoring we need to use "k8s_container" instead.

Unfortunately it's hard to detect whether a cluster uses legacy or new-style Stackdriver. We can do it if we have access to the GKE API, but that typically won't be the case.

jskeet:
We should look at this again before 3.0, in case we want to change the default.

garethbudden:
Out of interest why is the GKE api not typically available?  For now I've managed to create the correct set of labels for legacy vs new stackdriver logging by hacking around with the metadata json file (happy to share the horror if you're interested) but it would be good to get it supported

jskeet:
@garethbudden: The Kubernetes API is only available with "compute credentials" (the default on GKE machines) if you have "Legacy authorisation" enabled in your cluster, which isn't the default or the recommended setup.

What we've got now definitely isn't where I'd like it to be, but I honestly don't know the best way forward at this stage :(

amanda-tarafa:
I'm assigning this to myself as well so I remember to check the state of things internally before 3.0.

amanda-tarafa:
I've asked internally for more information about legacy/new monitoring in Kubernetes. I will update here when we get the info and have come up with how we want to support this.

amanda-tarafa:
Given our current plan of migrating GAX and Google Cloud .NET client libraries to gRPC C# 2.x (see googleapis/google-cloud-dotnet#3519) our plan regarding this issue is as follows:

- Keep GAX v2.x (current version) creating legacy types only, that is  "container" or "gke_container" (pre/post #346)
- Have GAX v3.x (first beta release probably next week 2020-02-17) create new types only, that is "k8s_container".
- We won't support conditional creation of legacy/new types, mainly because there doesn't seem to be a consistent/documented way to detect whether running on legacy/new. 
- In places in which we create monitored resources whith a given type, client code has the possibility to modify such types after creation, so client code is free to detect legacy/new and modify the resources' types accordingly. If we've missed a case in which this is not possible, please feel free to open a new issue describing the problem.

amanda-tarafa:
[...]

PR #386, opened 2020-02-08
use kube-env vars to check new/legacy stackdriver

garethbudden:
Hi @amanda-tarafa ,
This is essentially what we're currently doing at work to figure out if we're running in legacy or new version of stackdriver; it looks in `instance.attributes.kube-env` of the metadata json for the `HEAPSTER_USE_NEW_STACKDRIVER_RESOURCES` variable and uses the value found.
It does seem to work for us, but I appreciate you may not want to take a dependency like this into the library  as it looks like it's using internal operation details? #347

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
@@ -5,6 +5,7 @@
  * https://developers.google.com/open-source/licenses/bsd
  */
 
+using System;
 using System.Collections.Generic;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


jskeet:
I think we'd only want to rely on documented, stable environment variables. But Amanda can have a look in more detail on Monday - thanks for the PR.

amanda-tarafa:
Hi @garethbudden, I'll check tomorrow how #347 is doing internally and get back to you on that.
I'll also try to find out if there is a more stable way to know whether we are running legacy or new Stackdriver, and maybe is just a matter of doing that. I'll update here as soon as I have more info.
Thanks for the PR.

amanda-tarafa:
Just a heads up, I have asked internally and will update when we know more.
I will keep this PR open until we have come up with a plan. Thanks!

amanda-tarafa:
Closing this PR now as we won't be supporting conditional detection of monitored resource types.
See [comment](https://github.com/googleapis/gax-dotnet/issues/347#issuecomment-586208127) in #347.
Please let us know if you encounter any problems with our approach and thanks very much for the PR.

PR #387, opened 2020-02-10
Proposal: Adds emulator support to Gax.Grpc

amanda-tarafa:
Still in draft because more tests are needed. And it's a proposal anyway.

I've added one test that shows what an actual client builder should do to use the base's emulator suppor code.

Might be a little messy on the generics but clients won't see that anyway, so I think we can deal with it.

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -20,7 +20,11 @@ namespace Google.Api.Gax.Grpc
     /// Base class for API-specific builders.
     /// </summary>
     /// <typeparam name="TClient">The type of client created by this builder.</typeparam>
-    public abstract class ClientBuilderBase<TClient>
+    /// <typeparam name="TEmulatorConfiguration">The type of the emulator configuration used by this builder.</typeparam>
+    /// <typeparam name="TSelf">The type of the class given value to <typeparamref name="TSelf"/></typeparam>
+    public abstract class ClientBuilderBase<TClient, TEmulatorConfiguration, TSelf>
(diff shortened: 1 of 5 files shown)


Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -20,7 +20,11 @@ namespace Google.Api.Gax.Grpc
     /// Base class for API-specific builders.
     /// </summary>
     /// <typeparam name="TClient">The type of client created by this builder.</typeparam>
-    public abstract class ClientBuilderBase<TClient>
+    /// <typeparam name="TEmulatorConfiguration">The type of the emulator configuration used by this builder.</typeparam>
+    /// <typeparam name="TSelf">The type of the class given value to <typeparamref name="TSelf"/></typeparam>
```
amanda-tarafa: nit for myself: given => giving

amanda-tarafa:
We can definetely make it "less" generic at the expense of some casting (of course) in cases where the emulator support in GAX is not enough. I'm thinking that those might be very few (currently Datastore which has emulator projectId) so I'm totally fine with the tradeoff.
I'll add a new test to this commit showing how emulator support can be customized. And then I'll add a new commit removing the generics so we can compare the two.

amanda-tarafa:
Rebased onto the Gax EmulatorDetection enum.
Maybe start by looking at the tests which show how everything comes together. Two tests, one for default support (as is in GAX, which would cover Firestore and Spanner's 3 low level libraries) and another one for custom support (which would cover Datastore and Spanner.Data (I think)).

amanda-tarafa:
Closing in favor of #402. Will keep branch around for a while just in case.

PR #388, opened 2020-02-11
Adds EmulatorDetection enum.

amanda-tarafa:
I've very intentionally replaced mentions of environment variables for emulator configuration. Added a comment saying that "emulator configuration is usually done through environment variables" or similar, but I think we don't want to be more specific than that here.

### DIFF
--- a/Google.Api.Gax/EmulatorDetection.cs
+++ b/Google.Api.Gax/EmulatorDetection.cs
@@ -1,0 +1,52 @@
+﻿// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


Google.Api.Gax/EmulatorDetection.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Always connect to the emulator, throwing an exception if no emulator
+        /// configuration is detected.
+        /// </summary>
+        EmulatorOnly = 2,
+
+        /// <summary>
+        /// Connect to the emulator if an emulator configuration is detected,
+        /// or production otherwise. This is a convenient option, but risks damage to
+        /// production databases or running up unexpected bills if tests are accidentally
+        /// run in production due to emulator configuration unexpectedly missing.
```
jskeet: Nit: this sentence has some missing words. Maybe "due to the emulator configuration being absent unexpectedly"?

PR #389, opened 2020-02-11
Prototype of emulator support in GAX

jskeet:
Steps for a library to support emulators:

- Override the AllEmulatorEnvironmentVariables and RequiredEmulatorEnvironmentVariables properties
- Override BuildEmulatorClientBuilder
- In the Build/BuildAsync, first call MaybeUseEmulator. If that returns a non-null value, just delegate to that to build the client

The MaybeUseEmulator method does all the common validation that it can, and delegates to BuildEmulatorClientBuilder if necessary, passing in the relevant environment variable values as a dictionary.

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -77,6 +77,24 @@ namespace Google.Api.Gax.Grpc
         /// </summary>
         internal GrpcImplementation GrpcImplementation { get; set; }
 
+        private EmulatorDetection _emulatorDetection;
+        /// <summary>
+        /// The emulator detection policy to apply when building a client. Derived classes which support
+        /// emulators should create public properties which delegate to this one.
+        /// </summary>
+        protected EmulatorDetection EmulatorDetection
+        {
+            get => _emulatorDetection;
+            set => _emulatorDetection = GaxPreconditions.CheckEnumValue(value, nameof(value));
+        }
+
+        /// <summary>
+        /// The provider used to retrieve environment variables. This is used to faciliate testing, and defaults
+        /// to using <see cref="Environment.GetEnvironmentVariable(string)"/>.
+        /// </summary>
+        protected Func<string, string> EnvrionmentVariableProvider { get; set; } =
+            variable => Environment.GetEnvironmentVariable(variable);
+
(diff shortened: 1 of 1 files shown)


Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -102,6 +120,8 @@ protected void CopyCommonSettings<TOther>(ClientBuilderBase<TOther> source)
             TokenAccessMethod = source.TokenAccessMethod;
             CallInvoker = source.CallInvoker;
             UserAgent = source.UserAgent;
+            GrpcImplementation = source.GrpcImplementation;
+            EmulatorDetection = source.EmulatorDetection;
```
amanda-tarafa: This might not be copying from a builder of the same client, not sure if it makes sense to copy `EmulatorDetection`. Imagine `TOther` supports emulator and `TClient` doesn't, and we copy `EmulatorOnly` to this builder which is for `TClient`. The emulator for `TClient` will never be set and in `TClientClientBuilder` the `EmulatorDetection` property will be `protected` so it can't even be changed back. Calling `Build` will always fail.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -122,6 +142,112 @@ protected virtual void Validate()
...
+                        GaxPreconditions.CheckState(false,
+                            "Emulator environment variable '{0}' is set, but '{1}' is not set.",
+                            sampleSet, sampleNotSet);
+                    }
+                    break;
+                default:
+                    throw new InvalidOperationException($"Invalid emulator detection value: {EmulatorDetection}");
+            }
+
+            // We know we want to use the emulator. Validate nothing else has been set that shouldn't be.
```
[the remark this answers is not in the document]
jskeet: Hmm. I take your point, but we've always validated in the past that nothing's been set... it's easier to remove that validation later than to add it in, I'd say.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: Yes, that's true. Hmm. Maybe we need to not do that, but document it very clearly. Tricky.

jskeet:
Closing in favour of #402.

PR #391, opened 2020-02-12
Provide a simpler retry API.

jskeet:
This is currently just a proposal. Next steps would be:

- Adding tests
- Checking that we can use this easily in places that use RetrySettings in APIs (on next-major-version)

This is an alternative approach to #369. It avoids putting extra logic in RetrySettings, but hopefully gives a simpler API for other code that needs details of the retry sequence.

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/RetryAttemptTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/RetryAttemptTest.cs
@@ -1,0 +1,122 @@
+﻿/*
+ * Copyright 2020 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Google.Api.Gax.Testing;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Api.Gax.Grpc.Tests
+{
+    public class RetryAttemptTest
+    {
+        private static readonly TimeSpan OneSecond = TimeSpan.FromSeconds(1);
+        private static readonly TimeSpan FiveSeconds = TimeSpan.FromSeconds(5);
+
+        [Fact]
+        public Task DeadlineRespected()
+        {
+            var settings = RetrySettings.FromExponentialBackoff(
+                maxAttempts: 10, initialBackoff: OneSecond, maxBackoff: FiveSeconds,
+                backoffMultiplier: 2, retryFilter: ex => true, backoffJitter: RetrySettings.NoJitter);
+            var scheduler = new FakeScheduler();
+            var deadline = scheduler.Clock.GetCurrentDateTimeUtc() + FiveSeconds;
+
+            var sequence = RetryAttempt.CreateRetrySequence(settings, scheduler, deadline, scheduler.Clock);
+            // Should attempt at T=0, T=1, T=3, then stop because the next attempt would be after the deadline.
+            return AssertAttemptsAsync(sequence, scheduler, () => new Exception(), 0, 1, 3);
+        }
+
+        [Fact]
+        public Task MaxAttemptsRespected()
+        {
+            var settings = RetrySettings.FromConstantBackoff(
+                maxAttempts: 4, backoff: OneSecond, retryFilter: ex => true, backoffJitter: RetrySettings.NoJitter);
+            var scheduler = new FakeScheduler();
+
+            var sequence = RetryAttempt.CreateRetrySequence(settings, scheduler);
+            // Should attempt at T=0, T=1, T=2, T=3, then stop because we're only allowed four attempts.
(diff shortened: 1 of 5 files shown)


jskeet:
Great - I'll find everywhere that we use RetrySettings.NextBackoff in google-cloud-dotnet and see whether this change makes that simpler.

jskeet:
APIs that need checking:

- Firestore
- Bigtable
- Spanner.Data
- Spanner.V1

jskeet:
Trying to use this in a couple of places, we sometimes don't have a clock, and that's fine because we don't have a deadline, and in those cases we don't use ShouldRetry anyway. So I'm tempted to move the IClock to being a parameter of ShouldRetry - any objections?

jskeet:
Significant snag: one place we use RetrySettings is in Spanner.Data, where we have a SpannerException rather than an RpcException. Humbug.
One option would be to still revert NextBackoff to be public again, so that using this new scheme is an *option*, but not the only option. Thoughts welcome.

amanda-tarafa:
> Significant snag: one place we use RetrySettings is in Spanner.Data, where we have a SpannerException rather than an RpcException. Humbug.
> One option would be to still revert NextBackoff to be public again, so that using this new scheme is an _option_, but not the only option. Thoughts welcome.

- Make RetrySettings generic on the Exception type?
- Or simply accept any Exception type on the predicate instead of RpcException and keep the helper methods that return a Predicate<RpcExeption> based on status codes?

chrisdunelm:
I also like this design.

For Spanner-line situations, could `CreateRetrySequence` allow a custom exception type, and custom way of determining if the error is retryable?
I accept this may be over-complex!

jskeet:
> Make RetrySettings generic on the Exception type?

That seems overly complicated for the sake of a small amount of tidying up.

> Or simply accept any Exception type on the predicate instead of RpcException and keep the helper methods that return a Predicate<RpcExeption> based on status codes?

Hmm. Maybe. So those methods would then return `Predicate<Exception>` but basically include `e is RpcException rpcException && ...`?

amanda-tarafa:
> Trying to use this in a couple of places, we sometimes don't have a clock, and that's fine because we don't have a deadline, and in those cases we don't use ShouldRetry anyway. So I'm tempted to move the IClock to being a parameter of ShouldRetry - any objections?

Maybe overload CreateRetrySequence so as to have one that doesnt receive a deadline and clock, and it's probably easy to check for null on ShouldRetry, and it that way, ShouldRetry can also be used to check for the exception and max attempts in cases where there might be no deadline?

jskeet:
> Maybe overload CreateRetrySequence so as to have one that doesnt receive a deadline and clock

Yup, that works for me.

amanda-tarafa:
> Hmm. Maybe. So those methods would then return `Predicate<Exception>` but basically include `e is RpcException rpcException && ...`?

Yes, and this model can then be used for all cases in which we need to retry, not just RPCs but more "complex" operations as well, as in Spanner with the RetriableTransaction.

jskeet:
Note that the foreach loop is still copied in quite a few places. We could probably make that public in GAX at some point - but we can do that later.

jskeet:
Okay, done, and there's now a test. PTAL, including the comment for the commit which goes into a bit more detail.

jskeet:
The API library changes are at https://github.com/googleapis/google-cloud-dotnet/pull/4264

Google.Api.Gax.Grpc/RetryAttempt.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// as <see cref="RetrySettings.InitialBackoff"/>. This is typically to allow an "immediate first retry".</param>
+        /// <returns></returns>
+        public static IEnumerable<RetryAttempt> CreateRetrySequence(RetrySettings settings, IScheduler scheduler, DateTime? deadline, IClock clock, TimeSpan? initialBackoffOverride = null)
+        {
+            GaxPreconditions.CheckNotNull(settings, nameof(settings));
+            GaxPreconditions.CheckNotNull(clock, nameof(clock));
+            GaxPreconditions.CheckNotNull(scheduler, nameof(scheduler));
+
+            // It's simpler not to need nullable logic when computing deadlines.
+            var effectiveDeadline = deadline ?? DateTime.MaxValue;
```
amanda-tarafa: Or maybe, since there is the other overload now, make the deadline non-nullable in this one?

Google.Api.Gax.Grpc.Tests/RetryAttemptTest.cs
```diff
@@ -0,0 +1,122 @@
+﻿/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
jskeet: Thanks, will fix.

Google.Api.Gax.Grpc/RetryAttempt.cs
jskeet: No, as there are cases where we might or might not have a deadline - quite a bit of code uses `DateTime?` for an effective deadline. It would be nasty to have to conditionalise which overload is called.

PR #392, opened 2020-02-13
Changes to the artist formerly known as GrpcImplementation

jskeet:
I thought we'd probably need to do this eventually, but GcpCallInvokerPool forced our hand a bit earlier.
I think this design is nicer though - all control over gRPC implementation is now in the ClientBuilder, basically.

I'd really like to get this merged on Friday if at all possible so I can make the corresponding changes in google-cloud-dotnet and the microgenerator, then start releasing betas on Monday. Apologies for dumping a large PR on you both like this...


jskeet:
(Oh, and definitely review the first commit first. The second one will be a breeze after that.)

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -92,70 +80,75 @@ public Task ShutdownChannelsAsync()
...
         /// </summary>
+        /// <param name="grpcAdapter">The gRPC implementation to use. Must not be null.</param>
         /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
         /// <param name="channelOptions">The channel options to include. May be null.</param>
         /// <returns>A task representing the asynchronous operation. The value of the completed
         /// task will be channel for the specified endpoint.</returns>
-        public async Task<ChannelBase> GetChannelAsync(string endpoint, GrpcChannelOptions channelOptions)
+        public async Task<ChannelBase> GetChannelAsync(GrpcAdapter grpcAdapter, string endpoint, GrpcChannelOptions channelOptions)
         {
             GaxPreconditions.CheckNotNull(endpoint, nameof(endpoint));
```
chrisdunelm: Check grpcAdaptor is not `null`?

jskeet: Yup, will do.

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -92,70 +80,76 @@ public Task ShutdownChannelsAsync()
...
             var credentials = await _lazyScopedDefaultChannelCredentials.Value.ConfigureAwait(false);
-            return GetChannel(endpoint, channelOptions, credentials);
+            return GetChannel(grpcAdapter, endpoint, channelOptions, credentials);
         }
 
-        private ChannelBase GetChannel(string endpoint, GrpcChannelOptions channelOptions, ChannelCredentials credentials)
+        private ChannelBase GetChannel(GrpcAdapter grpcAdapter, string endpoint, GrpcChannelOptions channelOptions, ChannelCredentials credentials)
         {
-            var key = new Key(endpoint, channelOptions);
+            var key = new Key(grpcAdapter, endpoint, channelOptions);
```
amanda-tarafa: +1 to having the grpc implementation as part of the key.

PR #393, opened 2020-02-14
Generated MonitoredResource for GKE now supports new Stackdriver Kubernetes Engine Monitoring instead of Legacy Stackdriver

### LINKED ISSUE #347: Support Stackdriver Kubernetes Engine Monitoring

amanda-tarafa:
This is a breaking change:

  * The `Type` of the monitored resource is now `k8s_container` instead of `gke_container`.
  * The `instance_id` label is not included.
  * The `zone` label is not included.
  * A new `location` label is included whose value is the location of the cluster (may be different from node/pod location).
  * The `namespace_id` label is now called `namespace_name`. The value is the same.
  * The `pod_id` label is now called `pod_name`. The value is the same.

There's more information about Legacy to Stackdriver Kubernetes Engine Monitoring migration in https://cloud.google.com/monitoring/kubernetes-engine/migration.

All the information needed to build a Legacy GKE `Google.Api.MonitoredResource` can be found in `Google.Api.Gax.GkePlatformDetails`. An instance of `Google.Api.Gax.GkePlatformDetails` representing the platform the code is running in can be obtained by calling `Google.Api.Gax.Platfomr.Instance()`.

Closes #347 .

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
@@ -90,6 +90,7 @@ namespace Google.Api.Gax.Grpc.Tests
   'instance': {
     'attributes': {
       'cluster-name': 'FakeClusterName',
+      'cluster-location': 'FakeClusterLocation'
     },
     'id': '123',
     'zone': 'projects/FakeProject/zones/FakeLocation'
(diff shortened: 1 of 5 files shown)


amanda-tarafa:
@garethbudden FYI.

garethbudden:
this excellent thank you @amanda-tarafa 😀

PR #396, opened 2020-02-26
Add more asyncness and more cancellation

jskeet:
(If you have any thoughts about making more methods of ScopedCredentialProvider and ChannelPool internal, let me know. Better to do it now than regret it later :)


jskeet:
Gosh - not sure why VS hadn't shown me that before. Fixed, will merge on green.

PR #397, opened 2020-02-26
Make some methods on ChannelPool and ScopedCredentialProvider internal

jskeet:
These can be made public later, but while they're internal we can
change the signatures easily if necessary.

The client libraries don't currently call these methods themselves,
and it seems unlikely that customer code would need to.


jskeet:
Note: I'm on the fence about this in general. It feels odd to have a type that isn't really useful directly, but this does feel like a sensibly-conservative approach.

jskeet:
Great - merging, thanks for the review. Will try building a local package and make sure it actually *does* work with all the client libraries some time next week :)

PR #398, opened 2020-02-27
Fixing client builder tests.

amanda-tarafa:
Since #396 options are now requiered to obtain a channel from a channel pool. The clien builder's default options are not the empty options but ones modified with a keep alive, on client builder tests, empty options were being used to obtain the expected channel and the actual channel was being obtained with the keep alive options so they weren't the same.


amanda-tarafa:
No worries :).

PR #403, opened 2020-03-06
Update auth dependencies to 1.44.1

jskeet:
This should fix our integration tests, which are failing due the
bug in 1.44.0 which makes loading credentials from a file
asynchronously fail due to a stream being closed while we're trying
to read from it.


jskeet:
(This will, of course, fail until 1.44.1 is published.)

jskeet:
Will rerun Travis when 1.44.1 is out, and merge on green.

PR #404, opened 2020-03-09
Add first attempt at a Grpc.Net.Client adapter.

jskeet:
This is very simplistic, and will fail with custom options (so won't
work with all Google Cloud libraries) but should work in many cases.

### DIFF
--- a/Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
+++ b/Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
@@ -1,0 +1,73 @@
+﻿/*
+ * Copyright 2020 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Grpc.Core;
+using Grpc.Net.Client;
+using System;
+using System.Text.RegularExpressions;
+
+namespace Google.Api.Gax.Grpc.GrpcNetClient
+{
+    /// <summary>
+    /// Implementation of <see cref="GrpcAdapter"/> for Grpc.Net.Client.
+    /// </summary>
+    public sealed class GrpcNetClientAdapter : GrpcAdapter
+    {
+        // Note: this is "Default" rather than "Instance" as we expect to have other factory methods later, e.g. accepting
+        // an HTTP client factory.
+
+        /// <summary>
+        /// Returns the default instance of this class.
+        /// </summary>
+        public static GrpcNetClientAdapter Default { get; } = new GrpcNetClientAdapter();
+
+        private GrpcNetClientAdapter()
+        {
+        }
+
+        /// <inheritdoc />
+        protected override ChannelBase CreateChannelImpl(string endpoint, ChannelCredentials credentials, GrpcChannelOptions options)
+        {
+            var grpcNetClientOptions = ConvertOptions(credentials, options);
+            var address = ConvertEndpoint(endpoint);
+            return GrpcChannel.ForAddress(address, grpcNetClientOptions);
+        }
+
+        // Internal for testing
+        internal static global::Grpc.Net.Client.GrpcChannelOptions ConvertOptions(ChannelCredentials credentials, GrpcChannelOptions options)
+        {
+            // If service config resolution is explicitly enabled, throw - we can't support that,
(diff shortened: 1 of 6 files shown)


Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
```diff
@@ -0,0 +1,84 @@
...
+            // If service config resolution is explicitly enabled, throw - we can't support that,
+            // and users may be depending on it.
+            if (options.EnableServiceConfigResolution == true)
+            {
+                throw new ArgumentException($"{nameof(options.EnableServiceConfigResolution)} is not currently supported in {nameof(GrpcNetClientAdapter)}");
+            }
+
+            if (options.CustomOptions.Count > 0)
+            {
+                throw new ArgumentException($"{nameof(options.EnableServiceConfigResolution)} are not currently supported in {nameof(GrpcNetClientAdapter)}");
```
amanda-tarafa: `nameof(options.CustomOptions)`

Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
```diff
@@ -0,0 +1,84 @@
...
+            if (endpoint.StartsWith("http:", StringComparison.Ordinal) || endpoint.StartsWith("https:", StringComparison.Ordinal))
+            {
+                return endpoint;
+            }
+            if (s_simpleAddress.IsMatch(endpoint))
+            {
+                // Note that we assume HTTPS for any bare address; this feels like a reasonable assumption for now.
+                return $"https://{endpoint}";
+            }
+            throw new ArgumentException($"Unable to use endpoint '{endpoint}' with Grpc.Net.Client");
```
amanda-tarafa: Just a question, if I understand this correctly, something like storage.googleapis.com won't work unless it is prepended with http(s)://?

jskeet: Yikes, you're absolutely right. I need to go back to that, basically.

Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
jskeet: Thanks, will fix.

Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
jskeet: (I've no idea why I thought everything would be numeric.)

jskeet: I'm tempted to just make it add "https://" for anything that doesn't start with "http:" or "https:" and call it done... if that fails (e.g. because we're trying to connect to "ipv6:[::1]:1234") then so be it... Thoughts?

jskeet: That's what I've done for now - PTAL.

Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
```diff
@@ -0,0 +1,73 @@
...
+            return new global::Grpc.Net.Client.GrpcChannelOptions
+            {
+                Credentials = credentials,
+                MaxReceiveMessageSize = options.MaxReceiveMessageSize,
+                MaxSendMessageSize = options.MaxSendMessageSize                 
+            };
+        }
+
+        // Internal for testing
+        internal static string ConvertEndpoint(string endpoint) =>
```
amanda-tarafa: Yep, fine!

Google.Api.Gax.Grpc.GrpcNetClient/GrpcNetClientAdapter.cs
amanda-tarafa: And just a ping on this :)

jskeet: Doh! Done now :) (Actually called it just "Custom options" rather than using the nameof, as it's not a property that users set.)

jskeet:
Thanks - will merge on green.

PR #406, opened 2020-03-16
Release all GAX libraries as v3.0.0.

jskeet:
There are no changes since 3.0.0-beta03.


jskeet:
There are no FIXMEs to implement, and none of the TODOs look like they need doing before this GA. (Some of them should probably be removed due to being ancient and unlikely to have be changed.)

jskeet:
Just realized I need to do CommonProtos at the same time...

jskeet:
(Amended the commit.)

jskeet:
(The reference from GAX to CommonProtos is a project reference, so should be okay.)

PR #408, opened 2020-04-27
Packaging improvements

jskeet:
(Defniitely worth reviewing the two commits separately.)


Google.Api.Gax.Rest/Google.Api.Gax.Rest.csproj
```diff
@@ -16,6 +16,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Google.Api.Gax\Google.Api.Gax.csproj" />
 
-    <PackageReference Include="Google.Apis.Auth" Version="1.44.1" />
+    <PackageReference Include="Google.Apis.Auth" Version="[1.44.1, 2.0.0)" />
```
amanda-tarafa, line 19: Question: If we don't pin Google.Apis.Auth though then its dependencies can go to a different major version while it itself stays under 2, right? And we have only pushed the "problem" one level deeper. And thinking, that actually happens for all our pinned dependencies whose own dependencies are not pinned, right?

jskeet: Yes. It's definitely not a complete solution - just better than nothing. (It would have prevented some issues with GAX 2.x and Grpc.Core 2.x, for example - and by "prevented" I mean "explained in terms of versions rather than causing exceptions")

PR #412, opened 2020-06-10
Common protos

jskeet:


jskeet:
It looks like there's a breaking change in google.rpc.ErrorDetails - I'm following up on that.

PR #414, opened 2020-06-23
Adds support for QuotaProject as a client builder option for both GAX and REST.

amanda-tarafa:
Marking as draft because:

- Google.Apis.Auth dependencies need to be reverted to packages instead of project.
- I still need to write the content for the file explaining why we are deprecataing per call credentials.

This won't build on Travis or AppVeyor until we have updated the dependencies.
As usual one commit at a time will make the most sense.

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -89,6 +89,12 @@ namespace Google.Api.Gax.Grpc
             set => _emulatorDetection = GaxPreconditions.CheckEnumValue(value, nameof(value));
         }
 
+        /// <summary>
+        /// The GCP project ID that should be used for quota and billing purposes.
+        /// May be null.
+        /// </summary>
+        public string QuotaProject { get; set; }
+
         // Note: when adding any more properties, CopyCommonSettings must also be updated.
 
         /// <summary>
@@ -115,6 +121,7 @@ namespace Google.Api.Gax.Grpc
             CallInvoker = source.CallInvoker;
             UserAgent = source.UserAgent;
             GrpcAdapter = source.GrpcAdapter;
+            QuotaProject = source.QuotaProject;
 
             // Note that we may be copying from one type that supports emulators (e.g. FirestoreDbBuilder)
             // to another type that doesn't (e.g. FirestoreClientBuilder). That ends up in a slightly odd situation,
@@ -131,7 +138,7 @@ namespace Google.Api.Gax.Grpc
         protected virtual void Validate()
         {
             // If there's a call invoker, we shouldn't have any credentials-related information or an endpoint.
-            ValidateOptionExcludesOthers("CallInvoker cannot be specified with credentials settings or an endpoint", CallInvoker,
+            ValidateOptionExcludesOthers($"{nameof(CallInvoker)} cannot be specified with credentials settings or an endpoint", CallInvoker,
                 ChannelCredentials, CredentialsPath, JsonCredentials, Scopes, Endpoint, TokenAccessMethod);
 
             ValidateAtMostOneNotNull("Only one source of credentials can be specified",
@@ -139,6 +146,9 @@ namespace Google.Api.Gax.Grpc
 
             ValidateOptionExcludesOthers("Scopes are not relevant when a token access method or channel credentials are supplied", Scopes,
                 TokenAccessMethod, ChannelCredentials);
+
+            ValidateOptionExcludesOthers($"{nameof(QuotaProject)} cannot be specified if a {nameof(CallInvoker)} or a {nameof(ChannelCredentials)} is specified", QuotaProject,
+                CallInvoker, ChannelCredentials);
         }
 
         /// <summary>
@@ -201,6 +211,7 @@ namespace Google.Api.Gax.Grpc
                     CheckNotSet(JsonCredentials, nameof(JsonCredentials));
                     CheckNotSet(Scopes, nameof(Scopes));
                     CheckNotSet(TokenAccessMethod, nameof(TokenAccessMethod));
+                    CheckNotSet(QuotaProject, nameof(QuotaProject));
 
                     void CheckNotSet(object obj, string name)
                     {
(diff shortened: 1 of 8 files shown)


amanda-tarafa:
This is now rebased onto the #415 and ready for review.
Still a draft until we have published Google.Apis.Auth.

Google.Api.Gax.Rest.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -55,6 +56,23 @@ public async Task ApiKey_NoOtherCredentials()
...
+                ApiKey = SampleApiKey,
+                QuotaProject = SampleQuotaProject 
+            };
+
+            Action<BaseClientService.Initializer> validator = initializer =>
+            {
+                Assert.Equal(SampleApiKey, initializer.ApiKey);
+
+            };
+            await ValidateResultAsync(builder, validator);
```
jskeet, line 73:
```suggestion
            Action<BaseClientService.Initializer> validator = initializer =>
                Assert.Equal(SampleApiKey, initializer.ApiKey);
            await ValidateResultAsync(builder, validator);
```

jskeet: But should this also be testing that the quota project is used?

Google.Api.Gax.Rest/ClientBuilderBase.cs
```diff
@@ -131,13 +141,16 @@ protected virtual IConfigurableHttpClientInitializer GetHttpClientInitializer()
                 CredentialsPath != null ? GoogleCredential.FromFile(CredentialsPath) :
                 JsonCredentials != null ? GoogleCredential.FromJson(JsonCredentials) :
                 null; // Use default credentials (maybe - see below)
-            // While we accept any credentials that are specified even if there's an API key,
-            // we don't try to load default credentials.
+                      // While we accept any credentials that are specified even if there's an API key,
+                      // we don't try to load default credentials.
```
jskeet, line 145: Revert these lines - this isn't part of the comment on the line above; it's a comment for the lines below.

Google.Api.Gax.Rest/ClientBuilderBase.cs
```diff
@@ -154,13 +167,16 @@ protected async virtual Task<IConfigurableHttpClientInitializer> GetHttpClientIn
                 CredentialsPath != null ? await GoogleCredential.FromFileAsync(CredentialsPath, cancellationToken).ConfigureAwait(false) :
                 JsonCredentials != null ? GoogleCredential.FromJson(JsonCredentials) :
                 null; // Use default credentials (maybe - see below)
-            // While we accept any credentials that are specified even if there's an API key,
-            // we don't try to load default credentials.
+                        // While we accept any credentials that are specified even if there's an API key,
+                        // we don't try to load default credentials.
```
jskeet, line 171: Ditto

Google.Api.Gax.Rest/ClientBuilderBase.cs
jskeet: (Maybe add a blank line above to clarify that?)

Google.Api.Gax.Rest.IntegrationTests/ClientBuilderBaseTest.cs
amanda-tarafa: Yep, added code to check that quota project is used.

Google.Api.Gax.Rest/ClientBuilderBase.cs
amanda-tarafa: Done, added blank line.

Google.Api.Gax.Rest/ClientBuilderBase.cs
amanda-tarafa: Yep

bshaffer:
Is there anything blocking here?

amanda-tarafa:
Waiting for a gRPC release that will happen on 28 July, and then we can merge this and release.

amanda-tarafa:
@jskeet Now that Grpc released I've updated the dependency versions to both Grpc and Google.Apis and this should be ready for merge.
The first commit contains the dependency update, the last commit addresses your original review comments. The other commits are the same as when you last reviewed.
Thanks!

amanda-tarafa:
Hmmm, Grpc.Net.Client only updated to 2.31.0-pre2, I don't think we need to update that dependency, because it doesn't depend on Grpc.Auth which is the one we really care about, but, for uniformity, it would be nice to have all Grpcs at the same version, no?

jskeet:
Yes, I'd really like to have all gRPC dependencies on the same version if possible.

I'd suggest that if this is a matter of waiting a day or two, we wait - but if we find there are reasons why it might be significantly longer, we *could* update just Grpc.Auth (and transitively Grpc.Core.Apis).

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -422,15 +438,22 @@ protected async virtual Task<ChannelCredentials> GetChannelCredentialsAsync(Canc
...
+            internal DelegatedTokenAccess(Func<string, CancellationToken, Task<string>> tokenAccessMethod, string quotaProject) =>
+                (_tokenAccessMethod, _quotaProject) = (tokenAccessMethod, quotaProject);
 
             public Task<string> GetAccessTokenForRequestAsync(string authUri, CancellationToken cancellationToken) =>
                 _tokenAccessMethod(authUri, cancellationToken);
+
+            public async Task<AccessTokenWithHeaders> GetAccessTokenWithHeadersForRequestAsync(string authUri = null, CancellationToken cancellationToken = default)
+            {
+                string token = await _tokenAccessMethod(authUri, cancellationToken).ConfigureAwait(false);
+                return new AccessTokenWithHeaders.Builder { QuotaProject = _quotaProject }.Build(token);
```
jskeet, line 455: If this is going to be called on every request, and *normally* QuotaProject isn't set, might it make sense to optimize for that case. So something like:

```csharp
return QuotaProject is null
    ? token
    : new AccessTokenWithHeaders.Builder { QuotaProject = _quotaProject }.Build(token);
```

Alternatively, could we keep a single `AccessTokenWithHeader.Builder` for all requests? I can't remember whether it's safe to reuse that.

amanda-tarafa: The `AccessTokenWithHeader.Builder.Build` method is not thread safe right now, so reusing the builder is not an option. I can make it thread safe, but I'd rather do that later, rerelease Google.Apis, and the come back here if that's OK.

I will optimize for `QuotaProject` not being set though by only using the builder if it is. Otherwise I can use the `AccessTokenWithHeader` constructor.

Let me know if this is fine.

jskeet: Yup, that sounds reasonable.

PR #422, opened 2020-09-24
Release GAX version 3.2.0 (all packages)

jskeet:
The only change is in dependencies. With 3.2.0, we have dependencies
on the latest versions of gRPC and Google.Api.CommonProtos, which in
turn has a dependency on the latest version of Google.Protobuf.

This reduces the risk of Google.Api.CommonProtos 2.1 being used with
more-recently-generated protos, which would cause failures with gRPC
due to an optimization in Grpc.Core.


amanda-tarafa (approved):
LGTM

PR #427, opened 2020-11-17
feat: Configures credentials so that JWTs are used as access tokens where possible.

amanda-tarafa:
This is the Gax only solution.
I will shortly submit a pair of PRs for the combined Google.Apis.Auth+Gax solution.

My two cents on pros and cons of this solution:

- Pros:
  - Is simple.
  - Requires no Auth library changes.
- Cons:
  - This is effectively Auth logic bleeding into client libs.
  - Might get ugly if the conditions for emitting JWTs change in the future.
  - Users wanting to create standalone credentials that emit JWTs need to know about this logic.
  - These JWT emitting credentials cannot be reused across endpoints.

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -344,8 +344,26 @@ namespace Google.Api.Gax.Grpc
                 CredentialsPath != null ? GoogleCredential.FromFile(CredentialsPath) :
                 JsonCredentials != null ? GoogleCredential.FromJson(JsonCredentials) :
                 GoogleCredential.GetApplicationDefault();
-            GoogleCredential scoped = unscoped.CreateScoped(Scopes ?? GetDefaultScopes());
-            GoogleCredential maybeWithProject = QuotaProject is null ? scoped : scoped.CreateWithQuotaProject(QuotaProject);
+
+            var effectiveScopes = Scopes ?? GetDefaultScopes();
+            var effectiveEndpoint = Endpoint ?? GetDefaultEndpoint();
+
+            // Skip scoping if conditions are met for JWT emission as access tokens.
+            GoogleCredential maybeScoped =
+                // Only service accounts emit JWTs as access tokens.
+                unscoped.UnderlyingCredential is ServiceAccountCredential &&
+                // JWTs are only valid access tokens if scopes are at most the default scopes.
+                !effectiveScopes.Except(GetDefaultScopes()).Any() &&
+                // JWTs are only valid access tokens if the endpoint is the default endpoint.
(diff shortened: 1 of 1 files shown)


jskeet:
Yes, the checking for it being a service account credential is slightly annoying. A small auth library change to indicate "can we generate JWTs with this credential" would hide that. I really would love it if we could avoid separating out user scopes and "default scopes".

> These JWT emitting credentials cannot be reused across endpoints.

Are you sure? I thought we cached the JWT per endpoint, so they'd just work across multiple endpoints by generating multiple JWTs - but you know this code better than I do.

We also need support for custom audiences as well, of course.

amanda-tarafa:
> > These JWT emitting credentials cannot be reused across endpoints.
> 
> Are you sure? I thought we cached the JWT per endpoint, so they'd just work across multiple endpoints by generating multiple JWTs - but you know this code better than I do.

The problem is the default scopes. We are not scoping the credential if the scopes are a subset of the default scopes, but that changes per API (or could change), so, what's a default scope for API A might not be for API B. A credential created with no scopes because of API A, would also generate a JWT for API B and that might be wrong (if some of the default scopes of API A are not default for API B).

> We also need support for custom audiences as well, of course.

Yes, for that we would need to have something similar to the combined solution, because we would need to register the mapping from host to audience, as the host is the only thing we get as a parameter when we are emitting the access token.

jskeet:
> The problem is the default scopes. We are not scoping the credential if the scopes are a subset of the default scopes, but that changes per API (or could change), so, what's a default scope for API A might not be for API B. A credential created with no scopes because of API A, would also generate a JWT for API B and that might be wrong (if some of the default scopes of API A are not default for API B).

But when we create a scoped credential, that's a new object - it doesn't affect the existing one. If API B can't use JWTs, then presumably we'd provide scopes - which means it would be creating a new object.

I can definitely see how your more complex approach addresses that with the constraint registration, without creating more objects... but I think I need to look in more detail and probably write a document with a bunch of different scenarios.

amanda-tarafa:
> But when we create a scoped credential, that's a new object - it doesn't affect the existing one. If API B can't use JWTs, then presumably we'd provide scopes - which means it would be creating a new object.

Ah yes, but I meant the same credential instance is not reusable across endpoints. When created on the ClientBuilder, there's (almost) no risk of the credential leaking, but it complicates things for standalone credentials if someone wanted to use this approach.

> If API B can't use JWTs

Is not that API B cannot use JWTs, is that it uses them under different conditions than API A, and that information is not stored anywhere on the credential, is just known when the credential is being created, so, once the credential is created, you don't know for which API it is correctly configure for JWTs and for which API it is not.

amanda-tarafa:
I'll close this one as well and keep the branch. #429 is an improved version of this one.

PR #429, opened 2020-11-25
Use self-signed JWTs by default

jskeet:
This change only affects service account credentials loaded either
as the application default credentials or via the
JsonCredentials or CredentialsPath properties.

(See internal document for further discussion.)

### DIFF
--- a/Google.Api.Gax.Grpc/ChannelPool.cs
+++ b/Google.Api.Gax.Grpc/ChannelPool.cs
@@ -23,8 +23,6 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class ChannelPool
     {
-        private readonly IEnumerable<string> _scopes;
-
         /// <summary>
         /// Lazily-created task to retrieve the default application channel credentials. Once completed, this
         /// task can be used whenever channel credentials are required. The returned task always runs in the
@@ -32,6 +30,9 @@ namespace Google.Api.Gax.Grpc
         /// The same channel credentials are used by all pools. The field is initialized in the constructor, as it uses
         /// _scopes, and you can't refer to an instance field within an instance field initializer.
         /// </summary>
+        /// <remarks>
+        /// The name here indicates that the credentials are scoped if they need to be - which currently, they never are.
+        /// </remarks>
         private readonly Lazy<Task<ChannelCredentials>> _lazyScopedDefaultChannelCredentials;
 
         // TODO: See if we could use ConcurrentDictionary instead of locking. I suspect the issue would be making an atomic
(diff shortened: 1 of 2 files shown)


amanda-tarafa (commented):
Left a few notes on the internal doc, but LGTM.

bshaffer (approved):

amanda-tarafa (approved):
LGTM

amanda-tarafa:
I've added the `do not merge` label so we don't merge accidentally.

jskeet:
Just looking into this more - we'll need to adjust ChannelPool as well, or the generator. By default, we'll end up with a channel pool that still uses scopes (so we still don't use self-signed JWTs).

amanda-tarafa:
Ah yes, I actually think we might want to modify both. We might want to have two channel pools, one unscoped for JWTs and one scoped with default scopes in case someone specified those explicitly?

jskeet:
If someone specifies scopes explicitly, we currently don't use the channel pool and I think it's probably okay to continue with that. It's pretty easy to change ChannelPool just to ignore the scopes it was provided with. We could potentially add a constructor accepting scopes and a flag to say "and yes, really use the scopes". All needs a bit more thought, but it's not urgent.

bshaffer:
@jskeet @amanda-tarafa the ESF fix should now be rolled out. Could you test using self-signed JWTs with regional endpoints and audiences? If everything works accordingly we can merge this!

jskeet:
@bshaffer: As noted above, we need to make a change to the channel pool to disable scoping there as well. I hadn't noticed that until I was trying to test this with the Recaptcha API today (which did work, although given a comment from a customer a couple of weeks ago, I'm somewhat surprised by that).

> Could you test using self-signed JWTs with regional endpoints and audiences?

We can test with a specific endpoint if you could let us know a good example to test. We don't allow specific audiences though, so I wouldn't expect to need to test anything other than endpoints.

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -43,12 +44,20 @@ public sealed class ChannelPool
...
         {
-            // Always take a copy of the provided scopes, then check the copy doesn't contain any nulls.
-            _scopes = GaxPreconditions.CheckNotNull(scopes, nameof(scopes)).ToList();
-            GaxPreconditions.CheckArgument(!_scopes.Any(x => x == null), nameof(scopes), "Scopes must not contain any null references");
+            // We validate the scopes parameter, even though we don't actually use it any more. This means we
+            // can start using them again later if necessary, without breaking clients.
+            GaxPreconditions.CheckNotNull(scopes, nameof(scopes));
+            GaxPreconditions.CheckArgument(!scopes.Any(x => x == null), nameof(scopes), "Scopes must not contain any null references");
             // In theory, we don't actually need to store the scopes as field in this class. We could capture a local variable here.
             // However, it won't be any more efficient, and having the scopes easily available when debugging could be handy.
```
amanda-tarafa, line 62: This comment does not apply anymore.

amanda-tarafa (approved):
LGTM

PR #430, opened 2020-11-25
First stage of the prototype of "REGAPIC" - a gRPC adapter to use REST/JSON

jskeet:
This is not ready to merge, but should give an idea of how I expect it to hang together.

(This PR is so that it's off my local machine, and folks can have a look at the progress.)


Google.Api.Gax.Grpc/Rest/GrpcRestAdapter.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Api.Gax.Grpc.Rest
+{
+    /// <summary>
+    /// Implementation of <see cref="GrpcAdapter"/> that uses HTTP/1.1 and JSON,
+    /// but via a gRPC <see cref="CallInvoker"/>.
+    /// </summary>
+    public sealed class RestGrpcAdapter : GrpcAdapter
```
chrisdunelm, line 19: The file is named `GrpcRestAdapter`, whereas this class is named `RestGrpcAdapter`; I assume that at least one of these should be changed ;)

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -0,0 +1,137 @@
...
+            HttpStatusCode.InternalServerError => StatusCode.Internal,
+            HttpStatusCode.NotFound => StatusCode.NotFound,
+            HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
+            // TODO: Others!
+            _ => StatusCode.Unknown
+        };
+
+        private Metadata GetTrailers(Task<HttpResponseMessage> httpResponseTask)
+        {
+            if (httpResponseTask.IsCompleted)
```
chrisdunelm, line 129: I think this needs a `!`, if the exception message is to be believed.

chrisdunelm (commented):
I don't have all the context here, but the approach looks reasonable.

Just a couple of minor comments.

jskeet:
Doh - thanks for both of those. Addressed.

Google.Api.Gax.Grpc/Rest/RestChannelCredentialsConfigurator.cs
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Api.Gax.Grpc.Rest
+{
+    class RestChannelCredentialsConfigurator : ChannelCredentialsConfiguratorBase
```
amanda-tarafa, line 15: There's another class named the same and very similar in CredentialExtensions. I believe this one here is not being used, only the one in CredentialExtensions.

Google.Api.Gax.Grpc/Rest/CredentialExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private sealed class RestChannelCredentialsConfigurator : ChannelCredentialsConfiguratorBase
+        {
+            internal AsyncAuthInterceptor Interceptor { get; private set; }
+
+            // TODO: Validate that we're okay to discard the ChannelCredentials.
+            // This isn't very clearly documented...
+            public override void SetCompositeCredentials(object state, ChannelCredentials channelCredentials, CallCredentials callCredentials)
+            {
+                Interceptor = callCredentials.ToAsyncAuthInterceptor();
```
amanda-tarafa, line 52:
```suggestion
                var configurator = ChannelCredentialsOnlyConfigurator();
                // If SetComposite is called here, then this fails with NotSupportedException.
                channelCredentials.InternalPopulateConfiguration(configurator, null);
                
                Interceptor = callCredentials.ToAsyncAuthInterceptor();
```

amanda-tarafa (commented):
I've left a couple of comments as well, this is neat, and simpler than I thought it would be.

Google.Api.Gax.Grpc/Rest/RestChannelCredentialsConfigurator.cs
jskeet: Deleted.

Google.Api.Gax.Grpc/Rest/CredentialExtensions.cs
jskeet: I don't follow this... will discuss when we get closer to actually merging into the main branch.

jskeet:
Merged into regapic-branch. Will all need reviewing again before merging into the main branch, but I've tried to clean up some of the comments.

PR #435, opened 2021-01-12
Add CopySettingsForEmulator method

jskeet:
When libraries pick up a released version of this change, they can
then address https://github.com/googleapis/google-cloud-dotnet/issues/5671

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -95,7 +95,8 @@ namespace Google.Api.Gax.Grpc
         /// </summary>
         public string QuotaProject { get; set; }
 
-        // Note: when adding any more properties, CopyCommonSettings must also be updated.
+        // Note: when adding any more properties, CopyCommonSettings must also be updated,
+        // and potentially CopySettingsForEmulator.
 
         /// <summary>
         /// Creates a new instance with no settings.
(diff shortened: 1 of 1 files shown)


jskeet:
(I haven't included QuotaProject here, given your previous comment on it. I could potentially add a comment in the method explaining that it's a credential-based setting, if you think that's worth it.)

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -130,6 +131,18 @@ protected void CopyCommonSettings<TOther>(ClientBuilderBase<TOther> source)
             EmulatorDetection = source.EmulatorDetection;
         }
 
+        /// <summary>
+        /// Copies common settings from the specified builder, expecting that any settings around
+        /// credentials and endpoints will be set by the caller, along with any client-specific settings.
+        /// Emulator detection is not copied.
```
amanda-tarafa, line 137: nit: just so we remember, maybe add "to avoid infinite recursion on build"

amanda-tarafa (approved):
LGTM. And fine with not copying quota project, for emulator only builders we already check it is not set.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: Done, thanks.

PR #440, opened 2021-03-24
Revert self-signed JWT PR

jskeet:
This reverts both of the commits in #429, as we now believe we'll need a different approach, and don't want to accidentally release a new version with that change.


amanda-tarafa (approved):
LGTM

PR #442, opened 2021-04-07
Add ClientBuilderBase.GrpcChannelOptions property

### LINKED ISSUE #437: Consider allowing gRPC options to be specified in client builders
https://github.com/googleapis/google-cloud-dotnet/issues/5987 is an example of where this would be useful.

It's pretty rare, but it would be nice for it to be simpler.

jskeet:
Fixes #437

Alternative property name suggestions welcome; I wondered about
CustomChannelOptions, but GrpcChannelOptions itself has a
CustomOptions property, and I didn't want those to be confused.


amanda-tarafa (approved):
LGTM

PR #447, opened 2021-04-15
Add partial class to support Decimal conversions

jskeet:
We *could* make these C# operator conversions, but I *think* it's
clearer to keep them as methods. We can always add the operators
later if we want.

### DIFF
--- a/Google.Api.CommonProtos.Tests/Type/DecimalTest.cs
+++ b/Google.Api.CommonProtos.Tests/Type/DecimalTest.cs
@@ -1,0 +1,106 @@
+﻿/*
+ * Copyright 2021 Google LLC All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+using Xunit;
+
+namespace Google.Api.CommonProtos.Tests.Type
+{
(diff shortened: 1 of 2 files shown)


jskeet:
Suggestions for more test cases are very welcome, btw.

Google.Api.CommonProtos.Tests/Type/DecimalTest.cs
```diff
@@ -0,0 +1,103 @@
...
+            // Exponents
+            { "2.3e4", 23000m },
+            { "2.3e+4", 23000m },
+            { "2.3e-4", 0.00023m },
+            { "2.3E4", 23000m },
+            { "2.3E-4", 0.00023m },
+            { "-2.3e4", -23000m },
+            { "-2.3e-4", -0.00023m },
+            // Explicit positive sign
+            { "+1.23", 1.23m },
```
amanda-tarafa, line 46: "-0.00"?

amanda-tarafa (approved):
LGTM

Google.Api.CommonProtos.Tests/Type/DecimalTest.cs
jskeet: Ooh, good one. That doesn't round-trip (it ends up as "0.00"). Will add two tests...

amanda-tarafa (approved):
LGTM again :)

PR #450, opened 2021-04-20
ci: Moves away from Keystore for secret storage.

amanda-tarafa:
This needs an acompanying CL that will be submitted soon.


jskeet (approved):
Worth checking that populatesecrets.sh has the +x bit (I don't think I can tell in code review) but basically LGTM :)

amanda-tarafa:
> Worth checking that populatesecrets.sh has the +x bit (I don't think I can tell in code review) but basically LGTM :)

Yep, I did remembered this time :)

PR #454, opened 2021-04-23
Make the single public Grpc-Rest class conditionally defined

jskeet:
With this in place, we can merge the regapic branch into master:

- Normal releases of GAX are fine (the presence of some internal
  classes isn't a problem)
- Releasing a version of GAX containing REGAPIC (e.g. another alpha)
  is fairly simple - we'd need to permit the REGAPIC environment
  variable to be set in the Kokoro job, but that's all we need
- Most REGAPIC development can still proceed without GrpcRestAdapter
- To develop *with* GrpcRestAdapter, just set REGAPIC=true and then
  start the solution from the command line

### DIFF
--- a/CommonProperties.Test.xml
+++ b/CommonProperties.Test.xml
@@ -3,6 +3,9 @@
 
   <!-- Build information -->
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


amanda-tarafa (approved):
LGTM

viacheslav-rostovtsev (approved):

PR #456, opened 2021-04-28
Merge regapic branch to master branch

jskeet:
This does not (or should not - please validate) include any Compute code


ReleaseVersion.xml
```diff
@@ -5,6 +5,6 @@
     - divergent versions, but for the moment this keeps things simpler.
     -->
   <PropertyGroup>
-    <Version>3.3.0-alpha01</Version>
+    <Version>3.3.0-alpha02</Version>
```
amanda-tarafa, line 8: Just double checking that this should be here. I think it should because now we'll have only one branch, but to make sure.

amanda-tarafa (approved):
LGTM

ReleaseVersion.xml
jskeet: Yup, that's correct. It's about to be changed to 3.3.0 though :)

PR #461, opened 2021-05-06
[fix] use presence testing instead of rejecting default

viacheslav-rostovtsev:


Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -115,8 +116,7 @@ internal static RestMethod Create(MethodDescriptor method, JsonParser parser)
                 var path = pathFactory(message);
                 var body = bodyFactory(message);
 
-                var fieldsToEncode = unusedEligibleFields.Where(field =>
-                    field.IsRequired || !IsDefaultValueForTranscoding(field, field.Accessor.GetValue(message)));
+                var fieldsToEncode = unusedEligibleFields.Where(field => !field.HasPresence || field.HasPresence && field.Accessor.HasValue(message));
```
jskeet, line 119: I would *definitely* use extra brackets here to make the precedence clear. I assume you want `!field.HasPresence || (field.HasPresence && field.Accessor.HasValue(message))`.

Alternatively, given that `||` is short-circuiting, just skip the second HasPresence check: `!field.HasPresence || field.Accessor.HasValue(message)`.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -142,11 +142,14 @@ internal HttpRequestMessage CreateRequest(IMessage protoRequest, string host)
             
             var uri = host is null ? new Uri(uriPathWithParams, UriKind.Relative) : new UriBuilder { Host = host, Path = uriPathWithParams }.Uri;
             
+            var content = transcodingResult.Body is null ? null : new StringContent(transcodingResult.Body, Encoding.UTF8, s_applicationJsonMediaType);
+            
             return new HttpRequestMessage
             {
+                Headers = { { HttpRequestHeader.ContentType.ToString(), "application/json" } },
```
jskeet, line 149: I don't *think* you need this - the content type header is logically part of the content, and I believe it will be set automatically with the way we're constructing the StringContent... unless we don't *have* any content, in which case I don't think we should be setting the content type anyway.

jskeet (commented):
Generally fine with the presence testing approach (just a clarity nit) but I'd like to know whether there was a specific reason for the content-type header change (which at the moment I'd prefer to reject).

Google.Api.Gax.Grpc/Rest/RestMethod.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestMethod.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -11,6 +11,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
```
jskeet, line 14: I suspect this using directive is no longer required.

jskeet (approved):
Looks good, thanks. Feel free to squash down to a single commit and then merge.

PR #462, opened 2021-05-06
Automatically include REGAPIC in alpha releases

jskeet:
This should make it easier to iterate on REGAPIC code.

(The change to build.sh is just to remove a pointless echo that had been irritating me.)


amanda-tarafa (approved):
LGTM

PR #464, opened 2021-06-01
feat: Exposes Google.Apis.Http.IHttpClientFactory in Rest.ClientBuilderBase

amanda-tarafa:
See googleapis/google-api-dotnet-client#1864

After this has been merged and release, users can either reference the new Gax version directly or wait for an update Storage release. Then their code should look something like this:

```cs
// Users implementation of Google.Apis.Http.IHttpClientFactory.
var proxiedHttpClientFactory = new ProxyHttpClientFactory(proxy);

var scopes = new string[] { "https://www.googleapis.com/auth/devstorage.full_control" };
var serviceCredentialFromFile = (ServiceAccountCredential) GoogleCredential.FromFile(sa_file).UnderlyingCredential;
var initializer = new ServiceAccountCredential.Initializer(serviceCredentialFromFile.Id)
{
    ProjectId = serviceCredentialFromFile.ProjectId,
    Key = serviceCredentialFromFile.Key,
    User = serviceCredentialFromFile.User
    HttpClientFactory = proxiedHttpClientFactory,
    Scopes = scopes
};
var credential = GoogleCredential.FromServiceAccountCredential(new ServiceAccountCredential(initializer));

var clientBuilder = new StorageClientBuilder
{
    Credential = credential,
    // All HTTP client instances for API calls will be obtained from this factory
    HttpClientFactory = proxiedHttpClientFactory
}

var client = clientBuilder.Build();
```

Notice also that once googleapis/google-api-dotnet-client#1851 is in, setting a proxy for a credential will be much more simple as well.

### DIFF
--- a/Google.Api.Gax.Rest.IntegrationTests/ClientBuilderBaseTest.cs
+++ b/Google.Api.Gax.Rest.IntegrationTests/ClientBuilderBaseTest.cs
@@ -240,6 +240,14 @@ ZUp8AsbVqF6rbLiiUfJMo2btGclQu4DEVyS+ymFA65tXDLUuR9EDqJYdqHNZJ5B8
             Assert.Throws<InvalidOperationException>(builder.ValidateForTest);
         }
 
+        [Fact]
+        public async Task HttpClientFactory()
+        {
+            var factory = new HttpClientFactory();
+            var builder = new SampleClientBuilder { HttpClientFactory = factory };
+            await ValidateResultAsync(builder, initializer => Assert.Same(factory, initializer.HttpClientFactory));
+        }
+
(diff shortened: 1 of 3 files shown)


jskeet (approved):
Nice! I'm impressed it's this simple :)

We may want to think about anything else we want to get into a new GAX release before releasing this... presumably as a workaround, users could create the `StorageService` themselves and then call the `StorageClientImpl` constructor, right?

amanda-tarafa:
Yes, it's good that we really use the client factory for everything :).

And yes, users should be able to create the `StorageService` and then `StorageClientImp` themselves while we haven't released this.

PR #466, opened 2021-06-02
Support GKE on Windows as well as we're likely to be able to

### LINKED ISSUE #431: Detect GKE platform with Windows containers
Currently `Platform.Instance()` fails to correctly detect GKE on Windows.

jskeet:
Fixes #431

Note that this doesn't get full details - it can't detect the
container name, because it doesn't have mount info. However, it
looks like it gets everything else (somewhat remarkably).


amanda-tarafa (approved):
LGTM

PR #467, opened 2021-06-03
Improve derivation of GKE container name from mountinfo

### LINKED ISSUE #411: Determine container name with no access to kube-apiserver
Initially posted in #305.

When first working with this package, the library was not determining the containerName in my GKE environment. This was caused by my Istio mTLS policy preventing access to kube-apiserver, in addition to my k8s service account not have enough permissions.

I note the code that retrieves the containerName runs this Regex over the contents of /proc/self/mountinfo:

```cs
var regex = new Regex($"/var/lib/kubelet/pods/{podUid}/containers/([^/]+)/.*/dev/termination-log");
```

In my case, this had no matches, as `podUid` could not be determined from kube-apiserver.

One cat and grep later (within a gke container), I noted that there was only one possible match for this regex in /proc/self/mountinto (when substituting `{podUid}` with `[^/]+`).

As pods and containers have their own file system, might it be suitable to disregard the podUid when determining the containerName (even as a fallback)? This should allow the determination of containerName even with no kube-apiserver access.

Cheers,
Shane.

jskeet:
If we find exactly one entry in mountinfo that looks like it should
give us the container name, but we don't know what the pod UID is, we
can assume it's correct. We still verify that it's correct if we
*do* know what the pod UID is.

Note that if there are multiple mountinfo entries that look like
they could give us the information, we *could* pick the one that has
the right pod UID if we wanted (and if we know the pod UID, and
there's only one matching entry) but we're in unexpected territory
anyway at that point - if we see it happening in the wild, we can
judge whether that would be a good idea or not.

Fixes #411

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/MonitoringResourceBuilderTest.cs
@@ -119,7 +119,7 @@ namespace Google.Api.Gax.Grpc.Tests
             {
                 NamespaceJson = namespaceJson,
                 PodJson = podJson,
-                MountInfo = new[] { "/var/lib/kubelet/pods/podid/containers/containername/ /dev/termination-log" }
+                MountInfo = new[] { "1 2 3 /var/lib/kubelet/pods/podid/containers/containername/ /dev/termination-log xyz" }
             };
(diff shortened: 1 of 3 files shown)


amanda-tarafa (approved):
LGTM

PR #470, opened 2021-06-10
feat: send correct x-goog-api-client headers for the diregapic

viacheslav-rostovtsev:
This makes the `x-goog-api-client` header for the diregapic libraries to have a `rest` component and sends the headers with the request.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -8,6 +8,8 @@
 using Google.Protobuf;
 using Grpc.Core;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Threading;
@@ -64,10 +66,24 @@ namespace Google.Api.Gax.Grpc.Rest
 
         private async Task<HttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request, CancellationToken cancellationToken)
         {
-            // TODO: Add headers from options.Headers, but intercept the x-goog-api-client header.
+            var headers = options.Headers
+                .Where(mh => !mh.IsBinary)
+                .Where(mh=> mh.Key != VersionHeaderBuilder.HeaderName)
+                .Select(mh => new KeyValuePair<string, string>(mh.Key, mh.Value))
+                .ToList();
+
+            var restVersion = new VersionHeaderBuilder()
+                .AppendDotNetEnvironment()
+                .AppendAssemblyVersion("gapic", GetType())
+                .AppendAssemblyVersion("gax", typeof(CallSettings))
+                .AppendAssemblyVersion("rest", typeof(HttpClient))
+                .ToString();
+
+            headers.Add(new KeyValuePair<string, string>(VersionHeaderBuilder.HeaderName, restVersion));
+
             // Ideally, add the header in the client builder instead of in the ServiceSettingsBase...
             // TODO: Use options. How do we set the timeout for an individual HTTP request? We probably need to create a linked cancellation token.
-            var httpRequest = restMethod.CreateRequest((IMessage) request, host);
+            var httpRequest = restMethod.CreateRequest((IMessage) request, host, headers);
             // TODO: How do we cancel this?
             await AddAuthHeadersAsync(httpRequest, restMethod).ConfigureAwait(false);
 
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -24,6 +24,13 @@ namespace Google.Api.Gax.Grpc.Rest
     /// </summary>
     internal sealed class RestChannel : ChannelBase
     {
+        private static readonly string RestVersion = new VersionHeaderBuilder()
+            .AppendDotNetEnvironment()
+            .AppendAssemblyVersion("gapic", typeof(RestChannel))
+            .AppendAssemblyVersion("gax", typeof(CallSettings))
+            .AppendAssemblyVersion("rest", typeof(HttpClient))
+            .ToString();
+
         private readonly AsyncAuthInterceptor _channelAuthInterceptor;
         private readonly HttpClient _httpClient;
         private readonly RestServiceCollection _serviceCollection;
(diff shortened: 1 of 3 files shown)


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -64,10 +66,24 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
         private async Task<HttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request, CancellationToken cancellationToken)
         {
-            // TODO: Add headers from options.Headers, but intercept the x-goog-api-client header.
+            var headers = options.Headers
+                .Where(mh => !mh.IsBinary)
+                .Where(mh=> mh.Key != VersionHeaderBuilder.HeaderName)
+                .Select(mh => new KeyValuePair<string, string>(mh.Key, mh.Value))
+                .ToList();
+
+            var restVersion = new VersionHeaderBuilder()
```
jskeet, line 75: We could presumably construct this once and store it as a static variable, right? I think that would be better, as I don't know what the efficiency of these methods is like.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -64,10 +66,24 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
+                .AppendAssemblyVersion("gax", typeof(CallSettings))
+                .AppendAssemblyVersion("rest", typeof(HttpClient))
+                .ToString();
+
+            headers.Add(new KeyValuePair<string, string>(VersionHeaderBuilder.HeaderName, restVersion));
+
             // Ideally, add the header in the client builder instead of in the ServiceSettingsBase...
             // TODO: Use options. How do we set the timeout for an individual HTTP request? We probably need to create a linked cancellation token.
-            var httpRequest = restMethod.CreateRequest((IMessage) request, host);
+            var httpRequest = restMethod.CreateRequest((IMessage) request, host, headers);
```
jskeet, line 86: I'm not sure that I'd put the header addition into RestMethod, given that nothing about the headers depends on the method. If we keep the header addition in this class instead, we don't need to materialize the list - just add all text headers (other than the version) from the options, then add the version header separately.

Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -62,6 +62,12 @@ public VersionHeaderBuilder AppendAssemblyVersion(string name, System.Type type)
         /// Appends the .NET environment information to the list.
         /// </summary>
         public VersionHeaderBuilder AppendDotNetEnvironment() => AppendVersion("gl-dotnet", s_environmentVersion.Value);
+        
+        /// <summary>
+        /// Whether the name or value that are supposed to be included in a header are valid
+        /// </summary>
+        private static bool IsHeaderNameValueValid(string nameOrValue)
+            => !nameOrValue.Contains(" ") && !nameOrValue.Contains("/");
```
jskeet, line 70: Nit: we normally put the `=>` at the end of the previous line.

Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -103,16 +109,15 @@ private static string GetEntryAssemblyVersionOrNull()
...
+        
         private static string FormatAssemblyVersion(System.Type type)
         {
             // Prefer AssemblyInformationalVersion, then AssemblyFileVersion,
             // then AssemblyVersion.
 
             var assembly = type.GetTypeInfo().Assembly;
             var info = assembly.GetCustomAttributes<AssemblyInformationalVersionAttribute>().FirstOrDefault()?.InformationalVersion;
-            if (info != null)
+            if (info != null && IsHeaderNameValueValid(info)) // Skip informational version if it's not a valid header value
```
jskeet, line 120: I assume this is because you've seen this going wrong somewhere? Out of interest, what happened?

viacheslav-rostovtsev: The informational version for `System.Net.Http` is `4.6.30015.01 @BuiltBy: AzDevOps-vseng20170000ZL @Branch: release/2.1-MSRC @SrcCode: https://github.com/dotnet/corefx/tree/96d2fdd8610e6eead9b357789266be9d149dd0b8`
So the code would take it and immediately throw when checking it for spaces and slashes.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: I've moved it but it feels strange to me that there is a method named `CreateRequest` and it creates a request object … except for the headers.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -143,13 +144,15 @@ internal HttpRequestMessage CreateRequest(IMessage protoRequest, string host)
             var uri = host is null ? new Uri(uriPathWithParams, UriKind.Relative) : new UriBuilder { Host = host, Path = uriPathWithParams }.Uri;
             
             var content = transcodingResult.Body is null ? null : new StringContent(transcodingResult.Body, Encoding.UTF8, s_applicationJsonMediaType);
-            
-            return new HttpRequestMessage
+
+            var httpRequestMessage = new HttpRequestMessage
```
jskeet, line 148: Revert all the changes to this file now?

Google.Api.Gax/VersionHeaderBuilder.cs
jskeet: Ick. That's horrible.

jskeet (approved):
While I see your point about creating a semi-incomplete request, I'd say that RestMethod is creating a request with all the information that comes from the rest method itself... other code can contribute other aspects. (There could even be further sources of information eventually, I guess.)

I definitely like it better now - I'd just suggest resetting RestMethod itself back to the previous version.

Google.Api.Gax/VersionHeaderBuilder.cs
viacheslav-rostovtsev: :+1:

Google.Api.Gax.Grpc/Rest/RestMethod.cs
viacheslav-rostovtsev: :+1:

PR #473, opened 2021-06-29
Set the gRPC max-receive-size to 2GB by default

jskeet:
Some APIs require large receive limits, and this change is in line
with other languages.


amanda-tarafa (approved):
LGTM

PR #474, opened 2021-07-07
Fix ClientBuilderBase integration tests

jskeet:
This also avoids future issues of the same nature - we obtain the
default settings programmatically instead of having to keep the two
in sync.


amanda-tarafa (approved):
LGTM

PR #485, opened 2021-07-28
feat: allow scoped Self-Signed JWT

TimurSadykov:
We keep both ChannelPool, ClientBuilderBase and GetCallInvoker needing to opt int for JWTs. ChannelPool gets a property that can be set on construction, ClientBuilderBase has a property that can be overwritten by descendants, and can be initialized in the constructor.

- Currently they are not opted-in by default. No one using GAX will be affected.
- Add a condition to ClientBuilderBase.CanUseChannelPool so that the channel pool can be used only if ChannelPool.UseJwt == ClientBuilder.UseJwt.


Gax.sln
```diff
@@ -48,6 +48,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "MSBuild", "MSBuild", "{68EA
 		ReleaseVersion.xml = ReleaseVersion.xml
 	EndProjectSection
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Apis.Auth", "..\google-api-dotnet-client\Src\Support\Google.Apis.Auth\Google.Apis.Auth.csproj", "{8F05BBAB-7B26-44DF-A139-4B631A5C6A85}"
```
jskeet, line 51: Google.Apis.Auth has now been released, so you should be able to revert this and use the new package.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,12 +59,22 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
...
                 }));
         }
 
+        internal DefaultChannelCredentialsCache(IEnumerable<string> scopes) : this(scopes, false)
+        {
+        }
+
         internal ChannelCredentials GetCredentials() =>
-            GetCredentialsAsync().ResultWithUnwrappedExceptions();
+        GetCredentialsAsync().ResultWithUnwrappedExceptions();
```
jskeet, line 77: Please revert this indentation change.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -114,6 +116,15 @@ protected ClientBuilderBase()
         {
         }
 
+        /// <summary>
```
jskeet, line 119: I think it would be simpler to add a virtual Boolean property that defaults to returning true. We can then use partial classes to disable it for specific APIs.

jskeet (commented):
Slightly annoying to have the same logic in three places, but I can see why we need it.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,12 +59,22 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
                     {
                         appDefaultCredentials = appDefaultCredentials.CreateScoped(_scopes);
                     }
+
+                    if (_useUseJwtAccessWithScopes && appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential)
```
amanda-tarafa, line 63: Asking here for `_useJwtAccessWithScopes` in order to change the credential makes assumption about what the default is in the credential itself. I'd rather not so we can make changes independently.

If you want to optimize instance creation, then it's better if we check that in `ServiceAccountCredential.WithUseJwtAccessWithScopes`, if the value being received as a parameter is the same as the current value, then we return `this`.

and then:
```suggestion
                    if (appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential serviceCredential)
```
and you can delete the first line inside the if.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,12 +59,22 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
...
+                    {
+                        ServiceAccountCredential serviceCredential = appDefaultCredentials.UnderlyingCredential as ServiceAccountCredential;
+                        appDefaultCredentials = GoogleCredential.FromServiceAccountCredential(serviceCredential.WithUseJwtAccessWithScopes(true));
+                    }
                     return appDefaultCredentials.ToChannelCredentials();
                 }));
         }
 
+        internal DefaultChannelCredentialsCache(IEnumerable<string> scopes) : this(scopes, false)
```
amanda-tarafa, line 72: Do we need this internal overload, or can we just call the other one always specifiying true or false?

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,12 +59,22 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
                     {
                         appDefaultCredentials = appDefaultCredentials.CreateScoped(_scopes);
                     }
+
+                    if (_useUseJwtAccessWithScopes && appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential)
+                    {
+                        ServiceAccountCredential serviceCredential = appDefaultCredentials.UnderlyingCredential as ServiceAccountCredential;
+                        appDefaultCredentials = GoogleCredential.FromServiceAccountCredential(serviceCredential.WithUseJwtAccessWithScopes(true));
```
amanda-tarafa, line 66:
```suggestion
                        appDefaultCredentials = GoogleCredential.FromServiceAccountCredential(serviceCredential.WithUseJwtAccessWithScopes(_useJwtAccessWithScopes));
```

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -39,8 +41,11 @@ internal sealed class DefaultChannelCredentialsCache
         /// if they require any.
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
+        /// <param name="useUseJwtAccessWithScopes"></param>
+        internal DefaultChannelCredentialsCache(IEnumerable<string> scopes, bool useUseJwtAccessWithScopes)
         {
+            _useUseJwtAccessWithScopes = useUseJwtAccessWithScopes;
```
amanda-tarafa, line 47:
```suggestion
            _useJwtAccessWithScopes = useJwtAccessWithScopes;
```
and everywhere

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -25,6 +25,8 @@ public sealed class ChannelPool
     {
         private readonly IEnumerable<string> _scopes;
 
+        private readonly bool _useUseJwtAccessWithScopes;
```
amanda-tarafa, line 28:
```suggestion
        private readonly bool _useJwtAccessWithScopes;
```

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
amanda-tarafa: Also, this whole class is internal. But you haven't changed code using this class, and by passing false here, there's effectively no change in behaviour at all. Is this in preparation for when jwts are the default? Then I'd rather we move that preparation to client code. Keep here just one constructor and have the client class pass true or false depending on what's the default for it.

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -54,13 +59,27 @@ public ChannelPool(IEnumerable<string> scopes)
...
+
         private async Task<ChannelCredentials> CreateChannelCredentialsUncached()
         {
             var appDefaultCredentials = await GoogleCredential.GetApplicationDefaultAsync().ConfigureAwait(false);
             if (appDefaultCredentials.IsCreateScopedRequired)
             {
                 appDefaultCredentials = appDefaultCredentials.CreateScoped(_scopes);
             }
+
+            if (_useUseJwtAccessWithScopes && appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential)
```
amanda-tarafa, line 78: Same comments as before.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -27,6 +27,8 @@ public abstract class ClientBuilderBase<TClient>
             .WithEnableServiceConfigResolution(false)
             .WithMaxReceiveMessageSize(int.MaxValue);
 
+        private readonly bool _useUseJwtAccessWithScopes;
```
amanda-tarafa, line 30:
```suggestion
        private readonly bool _useJwtAccessWithScopes;
```

Google.Api.Gax.Grpc/ClientBuilderBase.cs
amanda-tarafa: +1, and it would be more in line with this being a builder, which is, this is mutable by defintion.

Not sure wether we want to validat the value being set or not when a credential is explicitly specified. @jskeet what do you think?

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -368,6 +379,13 @@ protected virtual ChannelCredentials GetChannelCredentials()
                 GoogleCredential.GetApplicationDefault();
             GoogleCredential scoped = unscoped.CreateScoped(Scopes ?? GetDefaultScopes());
             GoogleCredential maybeWithProject = QuotaProject is null ? scoped : scoped.CreateWithQuotaProject(QuotaProject);
+
+            if (_useUseJwtAccessWithScopes && maybeWithProject.UnderlyingCredential is ServiceAccountCredential)
```
amanda-tarafa, line 383: And same comment as before.

amanda-tarafa (commented):
I've left a few comments. Makes sense overall.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: Perhaps I wasn't clear - I wasn't expecting this to be publicly mutable, although I guess we *could* do that. I was going to keep it read-only, but with subclasses able to override the default, effectively.

I think it's fine for this to be true, but ignored due to an explicit credential being set.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
TimurSadykov: thanks... funny typo :)

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
TimurSadykov: I was not sure about this one yet. Now I see that it is used only by GcpCallInvokerPool and that class isn't referenced by Gapic generator. It probably mean this class is out of scope for now? 

Or if we want to provide general functionality to enable SS Jwt if needed, it means, as you suggested, I can modify GcpCallInvokerPool accordingly and explicitly pass true/false into here.

Gax.sln
TimurSadykov: Yep, will do

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
jskeet: GcpCallInvokerPool is only used by Spanner and PubSub, I believe. (It should probably go away entirely in the next version of GAX, to be honest.)

I need to have a think about how all of this works with channel pools in general... I'll *try* to do that this afternoon, but unfortunately it may need to wait until late next week. (I'm on vacation from tomorrow until Wednesday afternoon.)

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -44,8 +46,11 @@ public sealed class ChannelPool
         /// if they require any.
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        /// <param name="useUseJwtAccessWithScopes"></param>
+        public ChannelPool(IEnumerable<string> scopes, bool useUseJwtAccessWithScopes)
```
jskeet, line 50: Just to clarify my own thinking: we'd expect a generator change to call this constructor in a way that's consistent with what the client builder says about using JWTs.

I think this is fine *if* we don't allow a user to override. If we *do* allow a user to override the use of JWTs, we'd need to know not to use the channel pool in that case.

I think the implication of this is that the ClientBuilderBase property that I've proposing should be protected, virtual and read-only, probably *defaulting* to false (as otherwise old libraries taking a dependency on the new GAX would do the wrong thing) and being overridden in *every* generated library... and a separate `bool?` property for using JWTs that would be null by default, forcing use or non-use.

If none of that is clear, I can come up with another PR showing what I mean :)

amanda-tarafa: If the second bool? property is for allowing users to overwrite the behaviour and us to know not to use the pool if it has been set, then I agree this is a good solution. The same if we have one inmutable property only and users cannot decide wether to use the JWTs or not (I like it better if users can choose, but I'm fine with either).

An alternative that we might want to look at for a next major version is to maybe make useJwts (and scopes which would be the breaking bit) part of the key used to fetch a channel from the ChannelPool, that way we can facilitate reusing of channels for more use cases.

TimurSadykov: Sorry if I don't get some details, my thoughts: 

I think we can keep things simple and NOT allow an override. We effectively have only GAPIC libraries in scope. If JWT is supported by the API, what would be the reason to override JWT? It is faster and more reliable. We can provide a way to override once the scope becomes broader than GAPIC libraries. 

Also, question, why we care about Pool is being set only in case of override? Or we always care? Then it means I miss something in current draft ? Something that disables pool in case of JWT or use JWT only if there is no pool ?

amanda-tarafa: > Also, question, why we care about Pool is being set only in case of override? Or we always care? Then it means I miss something in current draft ? Something that disables pool in case of JWT or use JWT only if there is no pool ?

The default pool uses very specific credentials, with the API default scopes, and even with your code as it is, not opted in to JWTs. So there are many scenarion in which we cannot use the pool, for instance, if the client provides specific scopes, or if the client provides a quota project, etc.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
TimurSadykov: done

Google.Api.Gax.Grpc/ChannelPool.cs
TimurSadykov: fixed

Google.Api.Gax.Grpc/ClientBuilderBase.cs
TimurSadykov: fixed

Google.Api.Gax.Grpc/ClientBuilderBase.cs
TimurSadykov: done

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
TimurSadykov: I have updated changes to this file, as its hard for me to estimate how soon it will go away. The solution is basically the same, so not much overhead to simply have it. Let me know if it is safe to assume this code as gone and skip it :)

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -441,14 +451,21 @@ protected virtual GrpcChannelOptions GetChannelOptions()
         /// Returns whether or not a channel pool can be used if a channel is required. The default behavior is to return
         /// true if and only if no quota project, scopes, credentials or token access method have been specified. Derived classes should
         /// override this property if there are other reasons why the channel pool should not be used.
+        /// We do not enable self-signed Jwts with channel pools until credentials are not part of the key
```
amanda-tarafa, line 454: This is an implementation detail that we don't need to include on the doc.

I would change two senteces up to say something like:

/// true if and only if no quota project, scopes, credentials or token access method have been specified and UseJwt... == GetChannelPool().UseJwt....

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -441,14 +451,21 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
             TokenAccessMethod == null &&
             Scopes == null &&
-            QuotaProject == null;
+            QuotaProject == null &&
+            UseJwtAccessWithScopes == GetChannelPool()._useJwtAccessWithScopes;
+
+        /// <summary>
+        ///  Returns whether or not self-signed JWTs will be used over OAuth tokens when OAuth scopes are explicitly set.
+        /// </summary>
+        protected virtual bool UseJwtAccessWithScopes => false;
```
amanda-tarafa, line 468:
```suggestion
        protected virtual bool UseJwtAccessWithScopes { get; }
```

amanda-tarafa (changes requested):
A few comments but mostly cosmetic.

@jskeet this is a solution that @TimurSadykov and I discussed last week, it goes like this:

- As initially discussed, keep both ChannelPool and ClientBuilderBase needing to opt int for JWTs. ChannelPool has a property that can be set on construction, ClientBuilderBase has a property that can be overwritten by descedandts, and can be initialized in the constructor.
- Currently they are **not** opted-in by default. No one using GAX will be affected.
- Add a condition to ClientBuilderBase.CanUseChannelPool so that the channel pool can be used only if ChannelPool.UseJwt == ClientBuilder.UseJwt.
- When we want GAPICs opted in
  - In the generator, for all clients:
    - Have the [default ChannelPool](https://github.com/googleapis/google-cloud-dotnet/blob/61c5c17c1b1a17549da6f2ae6a819440d3982f41/apis/Google.Cloud.AccessApproval.V1/Google.Cloud.AccessApproval.V1/AccessApprovalServiceClient.g.cs#L281) be initialized opted in to JWTs, passing true in the constructor.
    - Add a constructor to the [client builder](https://github.com/googleapis/google-cloud-dotnet/blob/61c5c17c1b1a17549da6f2ae6a819440d3982f41/apis/Google.Cloud.AccessApproval.V1/Google.Cloud.AccessApproval.V1/AccessApprovalServiceClient.g.cs#L171) that sets the UseJwt to true. (Don't override here, as we want to override later on in partials).
  - After generation, for the 3 APIs that still don't support JWTs, write two partial classes:
    - One of the client itself with a static constructor reinitializing the ChannelPool to **not opted in** to JWTs.
    - One of the client builder to override the UseJwt property to return always false.

This will achieve:
- No one using GAX independently is affected by the change, they'd need to opt in.
- Gapics will continue to use the Channel Pool under the same circumstances as before, regardless if their APIs support JWTs or not.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet, line 468: This indicates the *default*, right? Is the plan not to allow users to customize this via another property, at the moment?

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -25,6 +25,8 @@ public sealed class ChannelPool
     {
         private readonly IEnumerable<string> _scopes;
 
+        internal readonly bool _useJwtAccessWithScopes = false;
```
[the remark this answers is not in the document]
jskeet: +1 for adding a constructor overload. (The current constructor can just delegate to that one.)

jskeet (commented):
I'm not sure about the idea of reinitializing the channel pool in a partial class. I think we can do better than that using a partial method... we'll need to change the generator a bit, but that's doable.

Other than that, and Amanda's existing comments, I think this is fine. (We really should look into removing the Gcp pacakge for the next major version...)

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,14 +59,23 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
...
                 }));
         }
 
         internal ChannelCredentials GetCredentials() =>
-            GetCredentialsAsync().ResultWithUnwrappedExceptions();
+        GetCredentialsAsync().ResultWithUnwrappedExceptions();
 
         internal Task<ChannelCredentials> GetCredentialsAsync() =>
-            _lazyScopedDefaultChannelCredentials.Value;
+        _lazyScopedDefaultChannelCredentials.Value;
```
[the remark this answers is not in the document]
TimurSadykov: I have reverted one already, it is my VS settings apparently... i'll check, thanks!

Google.Api.Gax.Grpc/ClientBuilderBase.cs
TimurSadykov: We have to remember that ourselves then :) Ok.

Google.Api.Gax.Grpc/ChannelPool.cs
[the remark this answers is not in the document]
TimurSadykov: @amanda-tarafa Ok, I agree that property would fit better, but why we want to make Jwt flag configurable? Could you please explain the scenario in mind? 

As I understand, right now Pools does not support Jwt because its not part of the key. I thought we better add constructor overload along with the update to the Pool implementation.

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
```diff
@@ -54,14 +59,23 @@ internal DefaultChannelCredentialsCache(IEnumerable<string> scopes)
...
                     }
+
+                    if (appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential)
+                    {
+                        ServiceAccountCredential serviceCredential = appDefaultCredentials.UnderlyingCredential as ServiceAccountCredential;
+                        if (serviceCredential.UseJwtAccessWithScopes != _useJwtAccessWithScopes)
+                        {
+                            appDefaultCredentials = GoogleCredential.FromServiceAccountCredential(serviceCredential.WithUseJwtAccessWithScopes(_useJwtAccessWithScopes));
+                        }
+                    }
```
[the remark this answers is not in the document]
TimurSadykov: applied, thanks

Google.Api.Gax.Grpc.Gcp/DefaultChannelCredentialsCache.cs
TimurSadykov: fixed again

Google.Api.Gax.Grpc/ChannelPool.cs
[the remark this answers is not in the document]
amanda-tarafa: Just for the record @TimurSadykov and I discussed this offline:

- Currently each pool either supports JWTs or not.
- If we, at some point, make the credential config part of the key on which the channels are pooled, then the same pool will be able to serve channels using JWTs and channels that don't use JWTs.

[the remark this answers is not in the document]
TimurSadykov: synched offline, fixed

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -61,6 +63,15 @@ private async Task<ChannelCredentials> CreateChannelCredentialsUncached()
             {
                 appDefaultCredentials = appDefaultCredentials.CreateScoped(_scopes);
             }
+
+            if (appDefaultCredentials.UnderlyingCredential is ServiceAccountCredential)
```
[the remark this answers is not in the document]
TimurSadykov: fixed

Google.Api.Gax.Rest/Google.Api.Gax.Rest.csproj
```diff
@@ -16,6 +16,6 @@
   <ItemGroup>
     <ProjectReference Include="..\Google.Api.Gax\Google.Api.Gax.csproj" />
 
-    <PackageReference Include="Google.Apis.Auth" Version="[1.52.0, 2.0.0)" />
+    <PackageReference Include="Google.Apis.Auth" Version="1.53.0" />
```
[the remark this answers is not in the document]
TimurSadykov: fixed

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -368,6 +368,16 @@ protected virtual ChannelCredentials GetChannelCredentials()
                 GoogleCredential.GetApplicationDefault();
             GoogleCredential scoped = unscoped.CreateScoped(Scopes ?? GetDefaultScopes());
             GoogleCredential maybeWithProject = QuotaProject is null ? scoped : scoped.CreateWithQuotaProject(QuotaProject);
+
+            if (maybeWithProject.UnderlyingCredential is ServiceAccountCredential)
```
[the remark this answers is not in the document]
TimurSadykov: updated

Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
```diff
@@ -37,7 +37,7 @@ public sealed class GcpCallInvokerPool
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
         public GcpCallInvokerPool(IEnumerable<string> scopes) =>
-            _credentialsCache = new DefaultChannelCredentialsCache(scopes);
```
jskeet, line 40: Are you going to add another constructor here? At the moment `DefaultChannelCredentialsCache` will never use JWTs.

Google.Api.Gax.Grpc/ChannelPool.cs
```diff
@@ -44,8 +46,11 @@ public sealed class ChannelPool
         /// if they require any.
         /// </summary>
         /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public ChannelPool(IEnumerable<string> scopes)
+        /// <param name="useJwtWithScopes"></param>
```
jskeet, line 49: Please fill in the documentation comment here.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -439,16 +446,23 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
         /// </summary>
         protected virtual bool CanUseChannelPool =>
             ChannelCredentials == null &&
             CredentialsPath == null &&
             JsonCredentials == null &&
             TokenAccessMethod == null &&
             Scopes == null &&
-            QuotaProject == null;
+            QuotaProject == null &&
+            UseJwtAccessWithScopes == GetChannelPool().UseJwtAccessWithScopes;
```
jskeet, line 460: Note to self (not a request for changes): some APIs make GetChannelPool throw an exception, but that's okay because they also override CanUseChannelPool.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -439,16 +446,23 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
+
+        /// <summary>
+        ///  Returns whether or not self-signed JWTs will be used over OAuth tokens when OAuth scopes are explicitly set.
+        /// </summary>
+        protected virtual bool UseJwtAccessWithScopes { get; }
```
jskeet, line 465: This will never be set to true at the moment unless it's overridden. I think we either need to add another constructor accepting the value, or give this a setter as well and make it non-virtual. The latter is probably simpler - and means if we ever want to make it public, derived classes can just declare a new public property that delegates to this one.

(I tend to avoid virtual automatically-implemented properties - if it's overridden, the field in the base class is still there but unused, which is a bit confusing.)

jskeet (commented):
Really nearly there now - just a few nits.

I'm going to start looking at what we want to generate, with a plan of:

- Existing library source code still compiles, and defaults to non-JWT
- Newly generated source code defaults to JWT
- A library can opt out (e.g. Compute) just by adding a partial class with a partial method.

I hope to get that done today. The precise details will depend on how ClientBuilderBase.UseJwtAccessWithScopes is changed, but it should be easy to change that.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: Also, we need to copy the value in CopyCommonSettings - which is easier if it's a regular automatically-implemented property.

jskeet:
(Yet another option would be an automatically implemented property that is initialized in the parameterless constructor via a virtual property. That feels overly complicated.)

jskeet:
Prototype of the library changes I'd envisage: https://github.com/googleapis/google-cloud-dotnet/pull/6950

Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
TimurSadykov: If I understood it correctly, it is not used by GAPIC libraries and might be removed soon, therefore, setting to default. Let me know if you think it is not a good idea.

jskeet: It's used by the Spanner and Bigtable libraries. If we don't mind those APIs not getting self-signed JWTs, then let's revert everything within this project. There seems to be little point in implementing the functionality but not actually allowing it to be used.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -437,18 +441,41 @@ protected virtual GrpcChannelOptions GetChannelOptions()
                 : defaultOptions.MergedWith(GrpcChannelOptions);
         }
 
+        /// <summary>
+        /// Sets the UseJwtAccessWithScopes for the provided credential 
+        /// </summary>
+        /// <param name="credential">a <see cref="GoogleCredential"/></param>
+        /// <returns>A <see cref="GoogleCredential"/> with UseJwtAccessWithScopes flag set</returns>
+        protected virtual GoogleCredential SetUseJwtWithScopesFlag(GoogleCredential credential)
```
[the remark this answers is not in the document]
amanda-tarafa: Do not make this protected virtual just for the sakes of tests. At the very most, internal.

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -340,7 +391,29 @@ private protected override ChannelBase CreateChannel(string endpoint, ChannelCre
                 return ChannelCreated;
             }
 
+            protected override GoogleCredential SetUseJwtWithScopesFlag(GoogleCredential credential)
```
amanda-tarafa, line 394: I would rather we don't know much about the credentials, than do something like this, wich requieres the method to be unnecessarily virtual.

You can check CanUseChannelPool instead.

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -340,7 +391,29 @@ private protected override ChannelBase CreateChannel(string endpoint, ChannelCre
...
+            {
+                GoogleCredentialsUsedToCreateChannel = base.SetUseJwtWithScopesFlag(credential);
+
+                return GoogleCredentialsUsedToCreateChannel;
+            }
+
             public override string ToString() => _name;
         }
+
+        public class SampleClientBuilderWithJwt : SampleClientBuilder
```
amanda-tarafa, line 404: Instad of adding all these different classes, add a constructor to the existing one that takes 2 bools `public SampleClientBuilder(bool clientUsesJwt, bool poolUsesJwt)`, and then you can test all your scenrios with the same client builder.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -391,6 +393,8 @@ protected async virtual Task<ChannelCredentials> GetChannelCredentialsAsync(Canc
                 await GoogleCredential.GetApplicationDefaultAsync(cancellationToken).ConfigureAwait(false);
             GoogleCredential scoped = unscoped.CreateScoped(Scopes ?? GetDefaultScopes());
             GoogleCredential maybeWithProject = QuotaProject is null ? scoped : scoped.CreateWithQuotaProject(QuotaProject);
+            maybeWithProject = SetUseJwtWithScopesFlag(maybeWithProject);
```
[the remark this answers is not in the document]
amanda-tarafa: Or even maybe just duplicate the code as is for all the rest instead of calling a method.

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
amanda-tarafa: @jskeet how would you feel about adding in ClientBuilderBase:

```
internal GoogleCredential GetGoogleCredential() {...}
internal Task<GoogleCredential> GetGoogleCredentialAsync() {...}
```
so we can write tests for the GoogleCredential that's used. This would be done in the same spirit as GetChannelCredentials in that validation has already been done, so we don't care if the builder really can use a GoogleCredential or not, we just give the best one that we can?

Or maybe we can have those be `protected virtual` to allow other types of GoogleCredentials in descendants, for instance impersonate credentials which we currently don't accept?

Of course, all of this can be done in another PR, and in this one keep the checks to the minimum.

amanda-tarafa (commented):
Left a few more comments, mainly on tests...

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
amanda-tarafa: Yes, I always forget about not exposing Google.Apis.Auth in Gax, hmmm, ...

OK, I can work on adding the internal methods next week or the other. @TimurSadykov, are you OK to test the bare minimum now, like CanUseChannelPool, etc. and then we can improve testing separately?

Google.Api.Gax.Grpc/ClientBuilderBase.cs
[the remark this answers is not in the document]
jskeet: Indeed, and presumably it's only internal for the sake of testing?

Google.Api.Gax.Grpc/ClientBuilderBase.cs
jskeet: Yes, I think it's fine to assume that GetChannelPool is "the one we've provided" and that we won't do anything weird.

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
TimurSadykov: @amanda-tarafa yes, that can work right now, I'll remove the SetUseJwtWithScopesFlag method

Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
TimurSadykov: thanks for the clarification, i'll add a constructor so Spanner and BigTable can use it

Google.Api.Gax.Grpc/ChannelPool.cs
TimurSadykov: done, thanks!

Google.Api.Gax.Grpc/ClientBuilderBase.cs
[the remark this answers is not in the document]
TimurSadykov: yes, the intent was just internal use for testing and some remove some duplicate code. But we are removing it for now...

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -391,6 +393,8 @@ protected async virtual Task<ChannelCredentials> GetChannelCredentialsAsync(Canc
...
+
             return maybeWithProject.ToChannelCredentials();
```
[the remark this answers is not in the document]
TimurSadykov: skipping since we are reverting this

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -282,27 +335,40 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
         {
             public static string DefaultEndpoint { get; } = "default.nowhere.com";
             public static string[] DefaultScopes { get; } = new[] { "scope1", "scope2" };
-            public ChannelPool ChannelPool { get; } = new ChannelPool(DefaultScopes);
+            public virtual ChannelPool ChannelPool { get; }
```
amanda-tarafa, line 338: No need for the virtual anymore.
```suggestion
            public ChannelPool ChannelPool { get; } 
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -282,27 +335,40 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
...
-            public ChannelPool ChannelPool { get; } = new ChannelPool(DefaultScopes);
+            public virtual ChannelPool ChannelPool { get; } 
 
             // The default options are private in ClientBuilderBase, but we can access them by getting the effective
             // options if we don't apply any modifications.
             public static GrpcChannelOptions DefaultOptions => new SampleClientBuilder().GetChannelOptions();
 
             public string EndpointUsedToCreateChannel { get; private set; }
             public ChannelCredentials CredentialsUsedToCreateChannel { get; private set; }
+            public GoogleCredential GoogleCredentialsUsedToCreateChannel { get; private set; }
```
amanda-tarafa, line 346: No need for this anymore either.
```suggestion
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -282,27 +335,40 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
...
 
+            /// <summary>
+            /// Constructor assigning Jwt flags for the sake of theory tests.
+            /// </summary>
+            public SampleClientBuilder(bool clientUsesJwt, bool poolUsesJwt)
+            {
+                ChannelPool = new ChannelPool(DefaultScopes, poolUsesJwt);
+                UseJwtAccessWithScopes = clientUsesJwt;
+            }
+
```
amanda-tarafa, line 371:
```suggestion
            public SampleClientBuilder(bool clientUsesJwt, bool poolUsesJwt)
                : this("Unnamed", clientUsesJwt, poolUsesJwt)
            {
            }
            
            public SampleClientBuilder(string name, bool clientUsesJwt, bool poolUsesJwt)
            {
                _name = name;
                ChannelPool = new ChannelPool(DefaultScopes, poolUsesJwt);
                UseJwtAccessWithScopes = clientUsesJwt;
            }

```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -282,27 +335,40 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
...
             private readonly string _name;
 
             /// <summary>
             /// Constructor assigning a "name" to a builder for the sake of theory tests.
             /// </summary>
-            public SampleClientBuilder(string name) => _name = name;
+            public SampleClientBuilder(string name) : this(false, false)
+            {
+                _name = name;
+            }
```
amanda-tarafa, line 357:
```suggestion
            public SampleClientBuilder(string name) : this(name, false, false)
            {
            }
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -226,6 +227,58 @@ public async Task TokenAccessMethod()
...
         }
 
+        [Fact]
+        public async Task JwtClientEnabledTest()
+        {
+            var builder = new SampleClientBuilder(true, false) { JsonCredentials = DummyServiceAccountCredentialFileContents };
+            ChannelBase channelFromPool = builder.ChannelPool.GetChannel(GrpcCoreAdapter.Instance, SampleClientBuilder.DefaultEndpoint, SampleClientBuilder.DefaultOptions);
+
+            // Jwt enabled for client and disabled for pool
+            // We won't use channel pool
```
amanda-tarafa, line 237:
```suggestion
        [Theory]
        [InlineData(true, false)]
        [InlineData(false, true)]
        public async Task DoesNotUsePool(bool clientUsesJwt, bool poolUsesJwt)
        {
            var builder = new SampleClientBuilder(clientUsesJwt, poolUsesJwt) { JsonCredentials = DummyServiceAccountCredentialFileContents };
            ChannelBase channelFromPool = builder.ChannelPool.GetChannel(GrpcCoreAdapter.Instance, SampleClientBuilder.DefaultEndpoint, SampleClientBuilder.DefaultOptions);

            // We won't use channel pool because it's either just
            // enabled for the client or just enabled for the pool.
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -226,6 +227,58 @@ public async Task TokenAccessMethod()
...
+            Action<CallInvoker> validator = invoker =>
+            {
+                var channelFromBuilder = GetChannel(invoker);
+                Assert.NotSame(channelFromPool, channelFromBuilder);
+                Assert.NotNull(builder.ChannelCreated);
+            };
+            await ValidateResultAsync(builder, validator);
+        }
+
```
amanda-tarafa, line 263: This is now part of the Theory
```suggestion
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -226,6 +227,58 @@ public async Task TokenAccessMethod()
...
+        [Fact]
+        public async Task JwtClientAndPoolEnabledTest()
+        {
+            var builder = new SampleClientBuilder(true, true);
+            ChannelBase channelFromPool = builder.ChannelPool.GetChannel(GrpcCoreAdapter.Instance, SampleClientBuilder.DefaultEndpoint, SampleClientBuilder.DefaultOptions);
+
+            // Jwt enabled for both client and pool
+            // We use channel pool
```
amanda-tarafa, line 271:
```suggestion
        [Theory]
        [Combinatorial]
        public async Task UsesChannelPool(bool enableJwts)
        {
            var builder = new SampleClientBuilder(enableJwts, enableJwts);
            ChannelBase channelFromPool = builder.ChannelPool.GetChannel(GrpcCoreAdapter.Instance, SampleClientBuilder.DefaultEndpoint, SampleClientBuilder.DefaultOptions);

            // Jwt enabled or disabled for both client and pool.
            // We use channel pool.
```

TimurSadykov: ah yes... with params I can use it.

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
TimurSadykov: ah yes

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
TimurSadykov: applied, added parameterless

Google.Api.Gax.Grpc.Gcp/GcpCallInvokerPool.cs
TimurSadykov: added

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -295,11 +333,26 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
...
+
+            public SampleClientBuilder(bool clientUsesJwt, bool poolUsesJwt)
+                : this("Unnamed", clientUsesJwt, poolUsesJwt)
+            {
+            }
+
+            public SampleClientBuilder(string name) : this(false, false)
+            {
+                _name = name;
+            }
```
amanda-tarafa, line 353: tiny nit: 
```suggestion
            public SampleClientBuilder(string name) : this(name, false, false)
            {
            }
```

Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs
```diff
@@ -295,11 +333,26 @@ public class SampleClientBuilder : ClientBuilderBase<CallInvoker>
...
 
-            public SampleClientBuilder() : this("Unnamed")
+            public SampleClientBuilder() : this("Unnamed", false, false)
```
amanda-tarafa, line 355: tiny nit:
```suggestion
            public SampleClientBuilder() : this(false, false)
```

amanda-tarafa (approved):
LGTM. And two tiny nits.

Also, you'll need to rebase the PR before it's ready for merge.

PR #488, opened 2021-07-29
feat: Surface better error information for REGAPIC

amanda-tarafa:
Draft as there are no tests, etc. Just to agree on the overall approach first.

fyi: @meteatamel

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -8,10 +8,10 @@
 using Google.Protobuf;
 using Grpc.Core;
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -71,7 +71,7 @@ namespace Google.Api.Gax.Grpc.Rest
             return new AsyncUnaryCall<TResponse>(responseTask, responseHeadersTask, statusFunc, trailersFunc, cancellationTokenSource.Cancel);
         }
 
-        private async Task<HttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request, CancellationToken cancellationToken)
+        private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request, CancellationToken cancellationToken)
         {
             // Ideally, add the header in the client builder instead of in the ServiceSettingsBase...
             // TODO: Use options. How do we set the timeout for an individual HTTP request? We probably need to create a linked cancellation token.
@@ -87,8 +87,20 @@ namespace Google.Api.Gax.Grpc.Rest
             // TODO: How do we cancel this?
             await AddAuthHeadersAsync(httpRequest, restMethod).ConfigureAwait(false);
 
-            var task = _httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseContentRead, cancellationToken);
-            return await task.ConfigureAwait(false);
+            var httpResponseMessage = await _httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseContentRead, cancellationToken).ConfigureAwait(false);
+
+            try
+            {
+                string content = await httpResponseMessage.Content.ReadAsStringAsync().ConfigureAwait(false);
+                return new ReadHttpResponseMessage(httpResponseMessage, content);
+            }
+            catch (Exception ex)
+            {
+                // Let's defer the throwing of this exception to when it's actually needed,
+                // so that we can at least read headers and other metadata.
+                var exInfo = ExceptionDispatchInfo.Capture(ex);
+                return new ReadHttpResponseMessage(httpResponseMessage, exInfo);
+            }
         }
 
         private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod method)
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -87,8 +87,20 @@ private async Task<HttpResponseMessage> SendAsync<TRequest>(RestMethod restMetho
...
+            var httpResponseMessage = await _httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseContentRead, cancellationToken).ConfigureAwait(false);
+
+            try
+            {
+                string content = await httpResponseMessage.Content.ReadAsStringAsync().ConfigureAwait(false);
+                return new ReadHttpResponseMessage(httpResponseMessage, content);
+            }
+            catch (Exception ex)
+            {
+                // Let's differ the throwing of this exception to when it's actually needed,
```
jskeet, line 99: differ => defer

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
-            var httpResponse = await httpResponseTask.ConfigureAwait(false);
-            // TODO: This could be very wrong. I don't know what headers we should really return, and I don't know about semi-colon joining.
-            var metadata = new Metadata();
-            foreach (var header in httpResponse.Headers)
-            {
-                metadata.Add(header.Key, string.Join(";", header.Value));
-            }
-            return metadata;
-        }
+        private async Task<Metadata> ReadHeadersAsync(Task<ReadHttpResponseMessage> httpResponseTask) =>
```
jskeet, line 120: I haven't checked, but if this is only called in one place we should probably just inline it.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
+        /// <summary>
+        /// In <see cref="AsyncUnaryCall{TResponse}"/> the functions to obtain the TResponse
+        /// and the <see cref="Status"/> of the call are two different functions.
+        /// The function to obtain the response is async, but the function to obtain the
+        /// <see cref="Status"/> is not.
+        /// For being able to surface error details in <see cref="Status"/> we need to be
+        /// able to call <see cref="HttpContent.ReadAsStringAsync"/> which is an async method,
+        /// and thus cannot be done, without blocking, on the sync function that obtains the 
+        /// <see cref="Status"/> in the <see cref="AsyncUnaryCall{TResponse}"/>.
+        /// So we need to make async content reading part of sending the call and not part of
```
jskeet, line 149: I think we'd want to remove this sentence when we make it a real PR.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
+        /// building the TResponse as it is now.
+        /// This class is just a convenient wrapper for passing together the <see cref="HttpResponseMessage"/>
+        /// and it's read response.
```
jskeet, line 152: it's => its

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
+
+            internal ReadHttpResponseMessage(HttpResponseMessage response, string content) =>
+                (OriginalResponseMessage, _content) = (response, content);
+
+            internal ReadHttpResponseMessage(HttpResponseMessage response, ExceptionDispatchInfo readException) =>
+                (OriginalResponseMessage, _readException) = (response, readException);
+
+            internal Metadata GetHeaders()
+            {
+                // TODO: This could be very wrong. I don't know what headers we should really return, and I don't know about semi-colon joining.
```
jskeet, line 178: We should check this before we go to GA...

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
+                var grpcStatus = OriginalResponseMessage.StatusCode switch
+                {
+                    HttpStatusCode.OK => StatusCode.OK,
+                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
+                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
+                    HttpStatusCode.BadRequest => StatusCode.InvalidArgument,
+                    HttpStatusCode.InternalServerError => StatusCode.Internal,
+                    HttpStatusCode.NotFound => StatusCode.NotFound,
+                    HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
+                    // TODO: Others!
```
jskeet, line 198: Another thing to sort out before GA.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -105,51 +117,97 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod me
...
+                    _ => StatusCode.Unknown
+                };
+
+                return new Status(grpcStatus,
+                    // Notice that here, if there was an exception reading the content
```
jskeet, line 203: Right - so this is an exception from "we can't even get the error message" rather than "the call failed and the error message is present, yes?

Is the error message likely to be in JSON that we should parse?

jskeet (commented):
Fine with the general approach here. A few tweaks may be required, and in particular we may want to parse JSON to create a more useful exception format, but it's a start...

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: I removed the "as it is now" bit. But I'm happy to remove the whole sentence, or even to turn this into a normal comment as a reminder for us.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Done

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Agreed.
Note that this was actually your note, I just moved this code here, but I made no changes to how we handle headers.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Agreed.
In particular I think we should map `StatusCode.OK` to `HttpResponseMessage.IsSuccessStatusCode`, since this is the measure of success that we use in Apiary.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Yes to the first part. This is "something went wrong and the request probably didn't even make it to the backend".

And yes, the error message is likely to be in JSON. My original intention was to do the same as we do in Apiary, which is, we attempt to parse, if we succeed then we set the parsed object, else we set the content as a string on the GoogleApiException.

The problem here is that `Grpc.Core.Status` only has a `StatusCode` and `string Detail` so we cannot set anything more complex there. And the `AsyncUnaryCall` expects `Grpc.Core.Status`.

Of course I could inherit from `Status` and add anything. I could even inherit from `RpcException` to have an exception type that only accepts the more specific `Status` type. Not sure if it is worth it, specially since the JSON will have some standard properties that we can expect, but API teams can add their own that we have no idea of, so, in Apiary, we end up adding those as a string I think.
What do you think?

amanda-tarafa (commented):
I've addressed the nits and left comments elsewhere. PTAL, thanks!

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: Just realised this is only internal anyway, so we don't need to worry *too* much.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: Yup, absolutely. This is a note to myself (and vicariously to Slava) really.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: I don't quite follow... but I think the "OK" path is fine at the moment.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: I think what's here is sufficient for a beta. Let's see what it looks like in reality, and find out what other languages are doing.

jskeet (approved):

amanda-tarafa (commented):
Tests for this change are included here: https://github.com/googleapis/google-cloud-dotnet/pull/6898. I've tested locally and they are green.

PR #490, opened 2021-08-09
Add more status code handling for the REST adapter

jskeet:
This should really be standardized across implementations.

This is still up for debate, but it's now consistent with [the Java implementation](https://github.com/googleapis/gax-java/blob/main/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonStatusCode.java) where feasible:

- Java throws a separate exception on 411
- Java uses the error message to determine the RPC code for a few cases

Using the error message would be feasible (as we have already read it by this point) but it feels brittle.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -186,16 +186,112 @@ namespace Google.Api.Gax.Grpc.Rest
 
             internal Status GetStatus()
             {
-                var grpcStatus = OriginalResponseMessage.StatusCode switch
+                var grpcStatus = (int) OriginalResponseMessage.StatusCode switch
                 {
-                    HttpStatusCode.OK => StatusCode.OK,
-                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
-                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
-                    HttpStatusCode.BadRequest => StatusCode.InvalidArgument,
-                    HttpStatusCode.InternalServerError => StatusCode.Internal,
-                    HttpStatusCode.NotFound => StatusCode.NotFound,
-                    HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
-                    // TODO: Others!
+                    // Note: this switch statement is organized numerically for ease of comparison with other implementations,
+                    // and to group 1xx, 2xx, 3xx etc codes together. The names in each comment correspond
+                    // with the names in the HttpStatusCode enumeration
+
+                    // 1xx range
+                    // Continue
+                    100  => StatusCode.Unknown, // We don't expect to get this
+                    // SwitchingProtocols
+                    101 => StatusCode.Unknown,
+
+                    // 2xx range
+                    // OK
+                    200 => StatusCode.OK,
+                    // Created
+                    201 => StatusCode.OK, // In the 200 range
+                    // Accepted
+                    202 => StatusCode.OK,
+                    // NonAuthoritativeInformation
+                    203 => StatusCode.OK, // In the 200 range
+                    // NoContent
+                    204 => StatusCode.OK, // In the 200 range,
+                    // ResetContent
+                    205 => StatusCode.Unknown, // While this isn't normally an error, our response won't contain what we expect.
+                    // PartialContent
+                    206 => StatusCode.Unknown, // While this isn't normally an error, our response won't contain what we expect.
+
+                    // 3xx range
+                    // Ambiguous, MultipleChoices
+                    300 => StatusCode.Unknown,
+                    // Moved, MovedPermanently
+                    301 => StatusCode.Unknown,
+                    // Redirect, Found
+                    302 => StatusCode.Unknown,
+                    // RedirectMethod, SeeOther
+                    303 => StatusCode.Unknown,
+                    // NotModified
+                    304 => StatusCode.Unknown, // While this isn't normally an error, our response won't contain what we expect.
+                    // UseProxy
+                    305 => StatusCode.Unknown,
+                    // Unused
+                    306 => StatusCode.Unknown,
+                    // TemporaryRedirect, RedirectKeepVerb
+                    307 => StatusCode.Unknown,
+
+                    // 4xx range
+                    // BadRequest
+                    400 => StatusCode.InvalidArgument, // Note: Java uses OutOfRange, FailedPrecondition or InvalidArgument based on the message
+                    // Unauthorized
+                    401 => StatusCode.Unauthenticated,
+                    // PaymentRequired
+                    402 => StatusCode.Unknown, // This is reserved for future use; we really shouldn't see it.
+                    // Forbidden
+                    403 => StatusCode.PermissionDenied,
+                    // NotFound
+                    404 => StatusCode.NotFound,
+                    // MethodNotAllowed
+                    405 => StatusCode.InvalidArgument,
+                    // NotAcceptable
+                    406 => StatusCode.InvalidArgument,
+                    // ProxyAuthenticationRequired
+                    407 => StatusCode.Unauthenticated,
+                    // RequestTimeout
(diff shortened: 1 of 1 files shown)


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
-                    HttpStatusCode.OK => StatusCode.OK,
-                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
-                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
+                    // Note: this switch statement contains all the unique named values within HttpStatusCode (as of 2021-08-09).
+
+                    HttpStatusCode.Accepted => StatusCode.OK,
+                    HttpStatusCode.Ambiguous => StatusCode.Unknown,
+                    HttpStatusCode.BadGateway => StatusCode.InvalidArgument,
                     HttpStatusCode.BadRequest => StatusCode.InvalidArgument,
+                    // Note: this is ambiguous between FailedPrecondition, Aborted or Unavailable.
```
amanda-tarafa, line 197: Should we go with Aborted so that clients maybe retry (Aborted is the one that's usually OK to retry).
On the other hand if it's really a FailedPrecondition (as in it's an erroneous argument) retrying won't solve it. But still client code should retry only a few times, etc.

So, I would favour giving client code the opportunity to recover by retrying.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
+                    HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
+                    HttpStatusCode.Continue => StatusCode.Unknown, // We don't expect to get this
+                    HttpStatusCode.Created => StatusCode.OK, // In the 200 range
+                    HttpStatusCode.ExpectationFailed => StatusCode.FailedPrecondition,
+                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
+                    // Note: Found and Redirect are both 302
```
amanda-tarafa, line 203: Is this note here out of place, or is there something missing?

amanda-tarafa: Ah, Redirect is done, maybe add that to the note.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
+                    HttpStatusCode.GatewayTimeout => StatusCode.DeadlineExceeded,
+                    HttpStatusCode.Gone => StatusCode.NotFound,
+                    HttpStatusCode.HttpVersionNotSupported => StatusCode.Unknown,
```
amanda-tarafa, line 206: Should this one be InvalidArgument?

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
                     HttpStatusCode.InternalServerError => StatusCode.Internal,
+                    HttpStatusCode.LengthRequired => StatusCode.InvalidArgument,
+                    HttpStatusCode.MethodNotAllowed => StatusCode.InvalidArgument,
```
amanda-tarafa, line 209: Unimplemented? Although I'm fine with InvalidArgument as well.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
+                    HttpStatusCode.PaymentRequired => StatusCode.Unknown, // This is reserved for future use; we really shoudln't see it.
+                    HttpStatusCode.PreconditionFailed => StatusCode.FailedPrecondition,
+                    HttpStatusCode.ProxyAuthenticationRequired => StatusCode.Unauthenticated,
+                    HttpStatusCode.Redirect => StatusCode.Unknown,
+                    // Note: RedirectKeepVerb and TemporaryRedirect are both 307
+                    HttpStatusCode.RedirectMethod => StatusCode.Unknown,
+                    HttpStatusCode.RequestEntityTooLarge => StatusCode.InvalidArgument,
+                    HttpStatusCode.RequestTimeout => StatusCode.DeadlineExceeded,
+                    HttpStatusCode.RequestUriTooLong => StatusCode.InvalidArgument,
+                    HttpStatusCode.RequestedRangeNotSatisfiable => StatusCode.FailedPrecondition,
```
amanda-tarafa, line 230: InvalidArgument?

amanda-tarafa (approved):
LGTM as is, although with a couple of alternatives.

+1 having these be the same for all languages.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -188,14 +188,60 @@ internal Status GetStatus()
...
                 var grpcStatus = OriginalResponseMessage.StatusCode switch
                 {
-                    HttpStatusCode.OK => StatusCode.OK,
-                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
-                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
+                    // Note: this switch statement contains all the unique named values within HttpStatusCode (as of 2021-08-09).
+
+                    HttpStatusCode.Accepted => StatusCode.OK,
+                    HttpStatusCode.Ambiguous => StatusCode.Unknown,
+                    HttpStatusCode.BadGateway => StatusCode.InvalidArgument,
```
vchudnov-g, line 195: This doesn't seem consistent with https://grpc.github.io/grpc/core/md_doc_http-grpc-status-mapping.html

jskeet: That table seems to be losing a *lot* of information though, and isn't really aimed at our use case, IMO. Mapping almost everything to unknown seems like a bad idea to me.

I'm happy to revisit specific ones like BadGateway (and Unavailable seems fine for that) but I don't think we should follow that table for everything. Where there's a really clear mapping (e.g. BadRequest to InvalidArgument) I don't see why we'd want to lose that information.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: I generally expect InvalidArgument to mean that there's something the caller can do about it - and that wouldn't be the case here.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: I've reorganized everything - it should be clearer now.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: Java uses different error codes based on the message. (Not entirely happy with that, really.) We'll see what consensus we get. I've made a note in the code.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: Done.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: (Mind you, it's not the case in various other things either...)

jskeet:
I've now updated this (completely changing the layout, in a way that I hope makes it easier to read).

@vchudnov-g: Are you happy to circulate this to everyone else who should be aware of it, so we can make this canonical? (I'm not suggesting it's already "right" - but that we make it right, then rely on this.)

vchudnov-g:
@jskeet Yes, I'll spread the word on this so we can get agreement.

jskeet:
Closing in favour of #502.

PR #492, opened 2021-08-16
Google.Api.Gax version 3.5.0

jskeet:
Changes since 3.4.0:

- Set the gRPC max-receive-size to 2GB by default
- Support for self-signed JWTs in gRPC client builders


amanda-tarafa (approved):
LGTM

PR #500, opened 2021-09-21
feat: Add extension methods to access extra error information

jskeet:
Extension methods are the least breaking option here: we don't
control the exception type (because we've always allowed
RpcException to just be thrown directly). While we could add a new
subclass of RpcException, that would have knock-on effects for
anything else that's already extending RpcException.

Any attempt to access extended information that is missing or
present but invalid will return null.

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/RpcExceptionExtensionsTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/RpcExceptionExtensionsTest.cs
@@ -1,0 +1,167 @@
+﻿/*
+ * Copyright 2021 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Google.Rpc;
+using Grpc.Core;
+using System;
+using Xunit;
+
+using Status = Google.Rpc.Status;
+using GrpcStatus = Grpc.Core.Status;
+using Google.Protobuf.WellKnownTypes;
+using Google.Protobuf;
+
+namespace Google.Api.Gax.Grpc.Tests
+{
+    public class RpcExceptionExtensionsTest
+    {
+        private static readonly GrpcStatus s_status = new GrpcStatus(StatusCode.InvalidArgument, "Bad request");
+
+        [Fact]
+        public void NullArgument_AllMethodsThrow()
+        {
+            RpcException ex = null;
+            Assert.Throws<ArgumentNullException>(() => ex.GetRpcStatus());
+            Assert.Throws<ArgumentNullException>(() => ex.GetBadRequest());
+            Assert.Throws<ArgumentNullException>(() => ex.GetErrorInfo());
+            Assert.Throws<ArgumentNullException>(() => ex.GetStatusDetail<BadRequest>());
+        }
+
+        [Fact]
+        public void NoTrailers_AllMethodsReturnNull()
+        {
+            RpcException ex = new RpcException(s_status);
+            AssertAllMethodsReturnNull(ex);
+        }
+
+        [Fact]
+        public void IrrelevantTrailer_AllMethodsReturnNull()
+        {
+            var metadata = new Metadata();
+            metadata.Add("key", "value");
+            RpcException ex = new RpcException(s_status, metadata);
+            AssertAllMethodsReturnNull(ex);
+        }
+
+        [Fact]
+        public void InvalidProtobufStatusTrailer_AllMethodsReturnNull()
+        {
+            var metadata = new Metadata();
+            metadata.Add(RpcExceptionExtensions.StatusDetailsTrailerName, new byte[] { 1, 2, 3, 4 });
+            RpcException ex = new RpcException(s_status, metadata);
+            AssertAllMethodsReturnNull(ex);
+        }
+
+        [Fact]
+        public void ValidTrailer_GetRpcStatus()
+        {
+            var status = new Status
+            {
+                Code = 123,
+                Message = "This is a message",
+            };
+
+            var metadata = new Metadata();
+            metadata.Add(RpcExceptionExtensions.StatusDetailsTrailerName, status.ToByteArray());
+            RpcException ex = new RpcException(s_status, metadata);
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <returns></returns>
+        public static T GetStatusDetail<T>(this RpcException ex) where T : class, IMessage<T>, new()
+        {
+            var status = GetRpcStatus(ex);
+            if (status is null)
+            {
+                return null;
+            }
+            var expectedName = new T().Descriptor.FullName;
+            var any = status.Details.FirstOrDefault(a => Any.GetTypeName(a.TypeUrl) == expectedName);
```
amanda-tarafa, line 68: If we can have many of these (not sure that's the case though), maybe add a version of this method that returns all of them? `IEnumerable<T> GetStatusDetails`?

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
jskeet: While in theory it would be possible to have multiple values for the same message type, I think it vanishingly unlikely. I suggest we wait until it's requested - expecting that it never will be :)

(Users can always do it themselves, of course.)

Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+            catch
+            {
+                // If the message is malformed, just report there's no information.
+                return null;
+            }
+        }
+
+        private static Metadata.Entry GetTrailer(RpcException ex, string key)
+        {
+            GaxPreconditions.CheckNotNull(ex, nameof(ex));
```
summer-ji-eng, line 86: Try to understand the behavior of `CheckNotNull`. It is possible gRPC metadata do not have key `grpc-status-details-bin`, then we should return the original error instead throw new error.

Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+                return null;
+            }
+            try
+            {
+                return any.Unpack<T>();
+            }
+            catch
+            {
+                // If the message is malformed, just report there's no information.
+                return null;
```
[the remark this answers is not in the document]
jskeet: I would suggest that Ads could add their own Ads-specific extension method in their library. That's one of the nice things about extension methods - anyone can add them within their own library.

Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
jskeet: `CheckNotNull` is checking that you haven't passed a null value for `ex` - that's all. Basically, this should fail:

```csharp
RpcException ex = null;
ex.GetErrorInfo();
```

But (as per the tests) it's absolutely fine to call it on an exception that doesn't have a trailer. There's no notion of "return the original error" though - the caller already has the `RpcException`; that's never returned. The only thing that's returned is extra information from within that exception.

summer-ji-eng: Perfect. Thanks for clarification. 👍

Google.Api.Gax.Grpc.Tests/RpcExceptionExtensionsTest.cs
```diff
@@ -0,0 +1,167 @@
...
+            Assert.Equal(badRequest, ex.GetStatusDetail<BadRequest>());
+        }
+
+        [Fact]
+        public void GetErrorInfo_Present()
+        {
+            var errorInfo = new ErrorInfo
+            {
+                Domain = "googleapis.com",
+                Reason = "broken"
```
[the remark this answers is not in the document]
jskeet: It doesn't matter - this is just a sample to make sure we can get the `ErrorInfo` back. Adding extra fields would be testing protobuf, not this code.

Google.Api.Gax.Grpc/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        public static BadRequest GetBadRequest(this RpcException ex) => GetStatusDetail<BadRequest>(ex);
+
+        /// <summary>
+        /// Retrieves the <see cref="ErrorInfo"/> message containing extended error information
+        /// from the trailers in an <see cref="RpcException"/>, if present.
+        /// </summary>
+        /// <param name="ex">The RPC exception to retrieve details from. Must not be null.</param>
+        /// <returns>The <see cref="ErrorInfo"/> message specified in the exception, or null
+        /// if there is no such information.</returns>
+        public static ErrorInfo GetErrorInfo(this RpcException ex) => GetStatusDetail<ErrorInfo>(ex);
```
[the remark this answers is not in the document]
jskeet: I would personally prefer not to do that to start with. It's easy to add them later on, but we couldn't take them away.

@amanda-tarafa do you agree, or do you think it's worth putting them on immediately?

(Note that `GetMetadata` would have to return an `IDictionary<string, string>` or similar, not just a string.)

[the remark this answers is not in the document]
summer-ji-eng: > (Note that GetMetadata would have to return an IDictionary<string, string> or similar, not just a string.)
Agree, the metadata is map<string, string> error_details.proto.

Node.js, Go and Ruby promote the values now.
It's up to @jskeet and @amanda-tarafa decide, thanks to [help](https://cloud.google.com/dotnet/docs/reference/help/troubleshooting) document, it could surface ErrorInfo to user.

summer-ji-eng (approved):
LGTM, except the promote ErrorInfo itself or promote ErrorInfo individual field. Either way, help documentation could draw attention for this.
 
Again, thanks a million make this happend.

jskeet:
Merged now - but I'd like to try to revisit #490 and get that done before we release a new version of GAX. Hopefully won't be more than a week or two though.

PR #502, opened 2021-10-08
Use canonical status code mapping from HTTP to RPC

jskeet:
This is now *not* consistent with [the Java implementation](https://github.com/googleapis/gax-java/blob/main/gax-httpjson/src/main/java/com/google/api/gax/httpjson/HttpJsonStatusCode.java) in some cases.

This is an alternative to #490

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -186,16 +186,36 @@ namespace Google.Api.Gax.Grpc.Rest
 
             internal Status GetStatus()
             {
-                var grpcStatus = OriginalResponseMessage.StatusCode switch
+                var httpStatus = (int) OriginalResponseMessage.StatusCode;
+                var grpcStatus = (httpStatus / 100) switch
                 {
-                    HttpStatusCode.OK => StatusCode.OK,
-                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
-                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
-                    HttpStatusCode.BadRequest => StatusCode.InvalidArgument,
-                    HttpStatusCode.InternalServerError => StatusCode.Internal,
-                    HttpStatusCode.NotFound => StatusCode.NotFound,
-                    HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
-                    // TODO: Others!
+                    // All 2xx responses map to OK
+                    2 => StatusCode.OK,
+
+                    // 4xx defaults to FailedPrecondition, with specific exceptions
+                    4 => httpStatus switch
+                    {
+                        400 => StatusCode.InvalidArgument,
+                        401 => StatusCode.Unauthenticated,
+                        403 => StatusCode.PermissionDenied,
+                        404 => StatusCode.NotFound,
+                        409 => StatusCode.Aborted,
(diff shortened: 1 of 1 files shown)


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -186,16 +186,36 @@ internal Metadata GetHeaders()
...
-                    HttpStatusCode.OK => StatusCode.OK,
-                    HttpStatusCode.Unauthorized => StatusCode.Unauthenticated,
-                    HttpStatusCode.Forbidden => StatusCode.PermissionDenied,
-                    HttpStatusCode.BadRequest => StatusCode.InvalidArgument,
-                    HttpStatusCode.InternalServerError => StatusCode.Internal,
-                    HttpStatusCode.NotFound => StatusCode.NotFound,
-                    HttpStatusCode.Conflict => StatusCode.FailedPrecondition,
-                    // TODO: Others!
+                    // All 2xx responses map to OK
+                    2 => StatusCode.OK,
```
amanda-tarafa, line 193: +1 to this.

The Apiary libraries marks all 2xx as success, if we didn't do this here, we could get the exact same operation succeed in Apiary but "fail" in REGAPIC.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -186,16 +186,36 @@ internal Metadata GetHeaders()
...
+                    // 5xx defaults to Internal, with specific exceptions
+                    5 => httpStatus switch
+                    {
+                        501 => StatusCode.Unimplemented,
+                        503 => StatusCode.Unavailable,
+                        504 => StatusCode.DeadlineExceeded,
+                        _ => StatusCode.Internal
+                    },
+
+                    // Anything else (including all 1xx and 3xx) maps to Unknown
```
amanda-tarafa, line 218: We handle redirects in Apiary.

amanda-tarafa: I have no idea if they are used, tbh. I've never seen an actual redirect while debugging code, but of course, I've onle ever debugged a low percent of operations. Maybe we can at least chat to the Compute API folk and ask, or maybe to ESF?

vchudnov-g: Any progress on the redirect issue? I would assume we'd want to handle it like Apiary and just follow the redirect. Ideally, that means HttpClient would do this and redirection codes would not show up here in practice....

jskeet: Sorry, I haven't chased the redirect issue at all. We could definitely "ask" HttpClient to do it for us, although I don't know whether it gives up after a few redirects.

jskeet: I suggest we can merge this change without worrying about redirects for now, but I'll file an issue to make sure we remember.

jskeet:
This has now been agreed internally, and is ready to merge if you're happy, @amanda-tarafa

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Just for completeness, this is being tracked in #511.

amanda-tarafa (approved):
LGTM

PR #503, opened 2021-10-08
feat: ForwardingCallInvoker, currently REGAPIC-specific

jskeet:
This is a CallInvoker implementation which is able to forward a
single unary method call to another CallInvoker. This will be used
by REGAPIC to allow non-standard LRO implementations to "look" like
standard ones, in terms of polling operations.

cc @viacheslav-rostovtsev


Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// </para>
+    /// <para>
+    /// Currently, only unary methods are supported, and only a single method can be forwarded. Any
+    /// other method results in an <see cref="RpcException"/> with a status code of <see cref="StatusCode.Unimplemented"/>.
+    /// The type parameters of this class would make it hard to support multiple calls, which is why the factory
+    /// class method has a return type of <see cref="CallInvoker"/> rather than the concrete class: we may implement
+    /// multi-method support via composition of multiple call invokers.
+    /// </para>
+    /// </remarks>
+    internal sealed class ForwardingCallInvoker<TSourceRequest, TSourceResponse, TTargetRequest, TTargetResponse> : CallInvoker
```
amanda-tarafa, line 68: nit: Any reason why this one is not inside the `#if REGAPIC`?

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
```diff
@@ -0,0 +1,155 @@
...
+
+        public override AsyncServerStreamingCall<TResponse> AsyncServerStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options, TRequest request) =>
+            throw new RpcException(CreateUnimplementedStatus("Streaming calls are unsupported in this CallInvoker"));
+
+        public override AsyncUnaryCall<TResponse> AsyncUnaryCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options, TRequest request)
+        {
+            if (method.FullName != _sourceMethodFullName)
+            {
+                // Rather than throw directly, here we *do* create a call that fails as if it had been
+                // rejected by a server. This is more likely to be handled appropriately by calling code.
```
amanda-tarafa, line 114: If this fails, it's a coding error, don't we want to just throw inmediatly? It could literally be ArgumentException I think.

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
jskeet: Yes - it makes testing a lot simpler. Basically, the only things we've got in `#if REGAPIC` are the public entry points for the REGAPIC work. Everything else is internal - so we're shipping extra unused code in our current GA releases, but that's not really a big problem - the important point is that we can still change the API.

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
```diff
@@ -0,0 +1,155 @@
...
+        // This is because there are multiple ways of consuming a stream, and throwing directly is more
+        // likely to expose the problem immediately. Also, *all* streaming calls will fail with this
+        // CallInvoker, so making the failure a more catastrophic one is probably appropriate.
+        public override AsyncClientStreamingCall<TRequest, TResponse> AsyncClientStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options) =>
+            throw new RpcException(CreateUnimplementedStatus("Streaming calls are unsupported in this CallInvoker"));
+
+        public override AsyncDuplexStreamingCall<TRequest, TResponse> AsyncDuplexStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options) =>
+            throw new RpcException(CreateUnimplementedStatus("Streaming calls are unsupported in this CallInvoker"));
+
+        public override AsyncServerStreamingCall<TResponse> AsyncServerStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options, TRequest request) =>
```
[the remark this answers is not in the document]
jskeet: I'm okay with changing this to InvalidOperationException (which I think is generally more suitable).

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
jskeet: I'm torn on this one. I'm worried that there might be plenty of error handling code that expects the initial call to go through, and for the failure to be thrown asynchronously.

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
amanda-tarafa: Ok

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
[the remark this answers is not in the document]
amanda-tarafa: OK

Google.Api.Gax.Grpc/ForwardingCallInvoker.cs
amanda-tarafa: I just see this as a coding error, not a runtime error. It shouldn't be handle, it should be fix.
But again, I'm fine with this as is, it's just a preference.

PR #505, opened 2021-10-14
Added support dotnet core 5 for CommonProtos

careless6666:
We are faced with the problem that we have long gone from dotnet standard, and everywhere we use dotnet core 5


jskeet:
.NET 5 supports .NET Standard 2.0, which there's already a target for. There is no need for this.

careless6666:
Package Aer.Platform.GrpcImports1.0.2080-feature-pl-366 is not compatible with netstandard2.0 (.NETStandard,Version=v2.0). Package Aer.Platform.GrpcImports.GoG1.0.2080-feature-pl-366 supports: net5.0 (.NETCoreApp,Ver ... : net5.0 (.NETCoreApp,Version=v5.0)

PR #508, opened 2021-11-01
gRPC convergence

jskeet:
This feels a little too easy to be "it", and there are some aspects we need to look more closely at:

- Implementing the Grpc.Gcp side of things
- Check the use of options for GrpcNetClient (lack of PrimaryUserAgent could be problematic)
- Check the logic of GrpcAdapter.DefaultAdapter
- Lots of testing in different scenarios
- Consider GrpcNetClientAdapter integration with IHttpClientFactory and the like

However, this is a good starting point for discussion.

cc @jtattermusch


Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
```diff
@@ -0,0 +1,165 @@
...
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Implementation of <see cref="GrpcAdapter"/> for Grpc.Core.
+    /// </summary>
+    public sealed class GrpcCoreAdapter : GrpcAdapter
+    {
+        private const string GrpcCoreAssemblyName = "Grpc.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=d754f35622e28bad";
+
+        // Option names, copied from ChannelOptions
```
amanda-tarafa, line 25: It's a little weird that we have these harcoded here (ultimately a dependency). I don't mind at all tbh, it just jumped at me a little be.

I know it will be more reflection, but maybe use the `Grpc.Core.ChannelOptions` class with all the constants? That way, in the unlikely case those change, we don't have to change the code.

Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            // Result should be equivalent to
+            // new Channel(endpoint, credentials, ConvertOptions(options));
+
+            // TODO: Benchmark using expressions vs invoking the Channel constructor by reflection.
+            // The latter may be more friendly to environments where expression trees do not work -
+            // but in those environments, it's unlikely that Grpc.Core is being used anyway.
+            // The other methods can be invoked with delegates, but delegates can't refer directly to methods.
+
+            // TODO: Somehow cache the result of ConvertOptions? It's probably not called terribly frequently.
```
amanda-tarafa, line 74: - Is ChannelOption inmutable?
- Consider caching each individual option instead of the list? And you can pick and choose based on the value?

Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -38,5 +43,27 @@ public ChannelBase CreateChannel(string endpoint, ChannelCredentials credentials
...
+
+        // TODO: Is this really what we want? Definitely simple, but not great in other ways...
+        // Perhaps have an environment variable that can be used to force one or the other?
+        // (In some situations it may be better to see "Grpc.Core fails immediately, but that's the one we want"
+        // rather than "Grpc.Net.Client looks like it will work, but then fails".)
+        private static GrpcAdapter CreateDefaultAdapter()
+        {
+            try
+            {
+                GrpcChannel.ForAddress("https://ignored.com");
```
amanda-tarafa, line 60: If this succeeds, can we guarantee that Grpc.Net.Client is fully usable?

amanda-tarafa (commented):
Looks good overall.

Google.Api.Gax.Grpc/GrpcAdapter.cs
jskeet: That's something we need to check with Jan. But I'm pretty sure we'll want some what of forcing it one way or another globally - an environment variable being the simplest approach.

Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
jskeet: I'd prefer not to - to be honest, I think it's more likely that the names in the C# code will change e.g. in a new major version than that the underlying values will change, given that the whole ecosystem will depend on those.

Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
jskeet: Yes, both ChannelOption and GrpcChanneloptions are immutable. Caching individual options would be trickier, given that we still need to come up with a sequence of the results. I suspect it's probably simpler to do all or nothing (and let's do nothing until we know we need to do anything...)

jskeet:
This is now ready for real review with an eye to merging, but before I merge I'll create new issues for:

- Considering whether clients should be able to *set* the default adapter (maybe only before first use?)
- Adding an environment variable for forcing the use of one adapter or another
- More options for GrpcNetClient, e.g. with an IHttpClientFactory and a logger.

jskeet:
(I'll then probably try to get the Grpc.Gcp code in before addressing those issues.)

Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        private GrpcCoreAdapter()
+        {
+        }
+
+        /// <inheritdoc />
+        protected override ChannelBase CreateChannelImpl(string endpoint, ChannelCredentials credentials, GrpcChannelOptions options) =>
+            channelFactory.Value.Invoke(endpoint, credentials, options);
+
+        private static Func<string, ChannelCredentials, GrpcChannelOptions, ChannelBase> CreateFactory()
```
amanda-tarafa, line 58: tiny nit: CreateChannelFactory

And also, I think a comment here with something like "A function is created, instead of just having a factory method, to avoid the reflection bits/expression building every time."

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/GrpcCoreAdapter.cs
jskeet: SGTM.

PR #512, opened 2021-11-05
feat: Expose HTTP to gRPC status code conversions in REGAPIC

jskeet:
This will allow custom LROs to be converted to standard ones more accurately.


amanda-tarafa (approved):
LGTM

PR #515, opened 2021-12-08
feat: escape variable expansion in URI path

viacheslav-rostovtsev:


Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -41,14 +41,21 @@ internal static HttpRulePathPattern Parse(string pattern, MessageDescriptor requ
...
                 int endOfName = match.Value.IndexOf('=');
-                if (endOfName == -1)
+                if (endOfName != -1)
                 {
-                    // If we don't have an '=', just go as far as the closing brace.
-                    endOfName = match.Value.Length - 1;
+                    // `startIndex: 1` for `{`; `endOfName - 1` for `=`
+                    boundFieldPath = match.Value.Substring(1, endOfName - 1);
+                    // `endOfName + 1` for `=`; `match.Value.Length - 3` for `={}`
+                    boundFieldPattern = match.Value.Substring(endOfName+1, match.Value.Length - 3 - boundFieldPath.Length);
```
jskeet, line 55: Nit - spaces before and after `+`

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -103,15 +109,30 @@ internal sealed class HttpRulePathPatternSegment
...
                 bool singleFieldPath = periodIndex == -1;
                 string fieldName = singleFieldPath  ? propertyPath : propertyPath.Substring(0, periodIndex);
 
-                return new HttpRulePathPatternSegment(fieldName, CreatePropertyAccessor(descriptor, propertyPath));
+                var propertyAccessor = CreatePropertyAccessor(descriptor, propertyPath);
+
+                // Per `google/api/http.proto`, if the pattern has  multiple path segments, such as `"{var=foo/*}"`,
+                // the `/` symbol should not be encoded, otherwise it should.
+                //TODO: [virost, 2021-12] Match pattern on the bound field, in addition to selecting the escape function
+                Func<string, string> escape_regular = value => Uri.EscapeDataString(value);
```
jskeet, line 127: Nit: please rename to `escapeRegular` and `escapeMultisegment`

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -103,15 +109,30 @@ internal sealed class HttpRulePathPatternSegment
...
                 int periodIndex = propertyPath.IndexOf('.');
                 bool singleFieldPath = periodIndex == -1;
                 string fieldName = singleFieldPath  ? propertyPath : propertyPath.Substring(0, periodIndex);
 
-                return new HttpRulePathPatternSegment(fieldName, CreatePropertyAccessor(descriptor, propertyPath));
+                var propertyAccessor = CreatePropertyAccessor(descriptor, propertyPath);
+
+                // Per `google/api/http.proto`, if the pattern has  multiple path segments, such as `"{var=foo/*}"`,
+                // the `/` symbol should not be encoded, otherwise it should.
+                //TODO: [virost, 2021-12] Match pattern on the bound field, in addition to selecting the escape function
```
jskeet, line 126: Super-nit: space after `//`

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -103,15 +109,30 @@ internal sealed class HttpRulePathPatternSegment
...
+                string boundFieldPattern, MessageDescriptor descriptor)
             {
                 int periodIndex = propertyPath.IndexOf('.');
                 bool singleFieldPath = periodIndex == -1;
                 string fieldName = singleFieldPath  ? propertyPath : propertyPath.Substring(0, periodIndex);
 
-                return new HttpRulePathPatternSegment(fieldName, CreatePropertyAccessor(descriptor, propertyPath));
+                var propertyAccessor = CreatePropertyAccessor(descriptor, propertyPath);
+
+                // Per `google/api/http.proto`, if the pattern has  multiple path segments, such as `"{var=foo/*}"`,
```
jskeet, line 124: What about `{var=foo/**}` with a match of `a/b/c`? At that point I'd expect something like `foo/a%2Fb%2Fc`... although I'm not entirely sure whether that's correct.

Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
```diff
@@ -18,7 +18,10 @@ public class HttpRulePathPatternTest
             { "x/y:custom", new RuleTestRequest(), "x/y:custom" },
             { "firstPart/{x}/secondPart/{y}", new RuleTestRequest { X = "x1", Y = "y2" }, "firstPart/x1/secondPart/y2" },
             { "combined/{x}-{y}/end", new RuleTestRequest { X = "xx", Y = "yy" }, "combined/xx-yy/end" },
+            { "pattern/{x}", new RuleTestRequest { X = "abc/def" }, "pattern/abc%2Fdef" },
             { "pattern/{x=abc/*}", new RuleTestRequest { X = "abc/def" }, "pattern/abc/def" },
+            { "pattern/{x=abc/*}", new RuleTestRequest { X = "abc/San Jose" }, "pattern/abc/San%20Jose" },
```
jskeet, line 23: Possibly use San Francisco or New York as an example instead of San Jose? I had to look up whether San Jose had an accent or not usually... and given that we're talking about escaping, I was expecting it to *be* an example of escaping accents.

Possibly add another example that *does* have an accent, e.g. "abc/caf\u00e9"? That would help to validate that we're escaping using UTF-8.

jskeet (approved):
Generally fine, just a few nits.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: I meant the variable part was a/b/c, so the complete value being "foo/a/b/c".

Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
viacheslav-rostovtsev: It was `San José` initially but the `é` gets mangled somewhere, and in the CI it is shown as a replacement character (`�`), so the test no longer works. I decided against trying to figure out where the problem happens at the moment.
Changing to New York.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
viacheslav-rostovtsev: That being said, `{var=**}` is specifically called in `http.proto` as leaving `/` unescaped, so adding this now.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
viacheslav-rostovtsev: If the pattern is `foo/**` and the variable is `foo/a/b/c`, the slashes should remain unescaped. I have added that as a test case as well.

Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
jskeet: Right, that's why I was suggesting `caf\u00e9` to make the source code fully ASCII, but still representing a non-ASCII value. I still think it would be good to do that.

viacheslav-rostovtsev: Thank you, implemented.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: Great, thanks. That test is exactly what I was looking for (assuming you've read the spec correctly - I haven't checked it).

jskeet (approved):

PR #516, opened 2021-12-08
feat: timeout http calls and add async headers calls

viacheslav-rostovtsev:


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class RestChannel : ChannelBase
             .AppendAssemblyVersion("gax", typeof(CallSettings))
             .AppendAssemblyVersion("rest", typeof(HttpClient))
             .ToString();
-
+
```
jskeet, line 32: nit: Please delete this extra space.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -59,7 +59,12 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
         {
             var restMethod = _serviceCollection.GetRestMethod(method);
 
-            var cancellationTokenSource = new CancellationTokenSource();
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
```
jskeet, line 62: Not sure why you've changed the `var` to an explicit type here... I think it's clear enough as it is.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -59,7 +59,12 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
+            if (options.Deadline.HasValue)
+            {
+                var timeout = (int) Math.Floor((options.Deadline.Value - DateTime.Now).TotalMilliseconds);
```
jskeet, line 65: This should definitely be DateTime.UtcNow - or better, use an injected `IClock` if possible. (We can leave it at `DateTime.UtcNow` for the moment though.)

jskeet (commented):
I'm fine with the general approach... I had vague memories of having helper methods to check for cancellation like this, but I'm not sure whether we actually do.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: This one is still in development and I should've converted it to Draft since I am still trying to test it with Compute.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -60,9 +62,12 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
             var restMethod = _serviceCollection.GetRestMethod(method);
 
             var cancellationTokenSource = new CancellationTokenSource();
+            if (options.Deadline.HasValue)
+            {
+                cancellationTokenSource = new CancellationTokenSource(options.Deadline.Value - DateTime.UtcNow);
```
jskeet, line 67: Please add a TODO here to inject an IClock in the future. That doesn't need to block GA though.

jskeet: (We should probably consider what happens if the deadline has already passed, too. If we end up with a TimeSpan with TotalMilliseconds of -1, that ends up as an infinite deadline instead of one just passed. I think we can fix that shortly after GA though.)

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -83,10 +87,23 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
             }
             httpRequest.Headers.Add(VersionHeaderBuilder.HeaderName, RestVersion);
 
-            // TODO: How do we cancel this?
-            await AddAuthHeadersAsync(httpRequest, restMethod).ConfigureAwait(false);
+            var addAuthHeadersTask = AddAuthHeadersAsync(httpRequest, restMethod);
+            var delayTask = Task.Delay(-1, cancellationToken);
+            if (delayTask == await Task.WhenAny(addAuthHeadersTask, delayTask).ConfigureAwait(false))
+            {
+                throw new InvalidOperationException("Timeout was reached when waiting for auth headers to be added for a method {restMethod.FullName}.");
```
jskeet, line 94: Throw OperationCanceledException instead of InvalidOperationException?

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -83,10 +87,23 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
+            }
 
-            var httpResponseMessage = await _httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseContentRead, cancellationToken).ConfigureAwait(false);
+            HttpResponseMessage httpResponseMessage;
+            try
+            {
+                httpResponseMessage = await _httpClient.SendAsync(httpRequest,
+                    HttpCompletionOption.ResponseContentRead, cancellationToken).ConfigureAwait(false);
+            }
+            catch(TaskCanceledException ex)
```
jskeet, line 103: I think we can remove this catch block - just let the cancellation propagate.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: Not an infinite deadline, but yet another exception of a different type. Fixed.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: I changed to the TimeoutException.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: The TaskCancelledException is extremely uninformative in Core 2.1 -- does not have an inner exception or any reference to a timeout. (At later versions I think they made it wrap a TimeoutException).
For the consistency I think wrapping it into a TimeoutException would be best.

jskeet: Tiny nit: space after catch.

But more importantly, I'm not at all sure about all the uses of TimeoutException here.

I've just tried making an RPC via gRPC, and it failed with a DeadlineExceeded RpcException. I don't know whether that came from the server or whether it was generated locally... would it be reasonable to synthesize that here? (It's easy enough to do.)

@amanda-tarafa any thoughts.

amanda-tarafa: There's nowhere else where we use TimeoutException that I recall, so I don't think this would be consitent with anything we already have. In Apiary, we just [buble up the cancellation exception](https://github.com/googleapis/google-api-dotnet-client/blob/main/Src/Support/Google.Apis.Core/Http/ConfigurableMessageHandler.cs#L423) regardless of whether it's because of a timeout or something else.

I think here I'd like best to be consistent with the rest of error mapping we are doing, so +1 to throwing an RpcException with a DeadlineExceeded status.

Wether we throw inmediately after we know that the deadline is exceeded, or whether we fake a 503 response, is another thing. I' would lean towards throwing inmediately, but then I think we should modify the GetStatus method (line 145) to take into account the case in which the task is faulted with an RpcException, so we can return the Status contained by that RpcException itself.

jskeet: We use TimeoutException in Polling, but only there.

I *thought* we had cunning code in RetryAttempt to avoid sleeping if we know that by the time we've finished sleeping the deadline will have passed, but I can't find that now.

viacheslav-rostovtsev: RpcException does not allow wrapping an inner exception, so we either throw two different exception types: RPC+TaskCancelled (a very uninformative TaskCancelled at that) or throw the TaskCancelled away.

jskeet: Hmm. We do have a slight issue here - while it's tempting to throw away the TaskCanceledException, if this is *actually* because the caller passed in a CancellationToken and that got canceled, it's weird to change that.

What we could potentially do is use the deadline and current time to work out what to throw - that's pretty weird, but would probably give the right expected semantics.

viacheslav-rostovtsev: There is no `caller`, only us. We generate a token from a source we control. The httpClient.SendAsync docs say that TaskCancelled exception happens because `The request failed due to timeout.`
I would throw an RpcException here, then we can ask the Grpc team to give us a constructor that takes an InnerException, and update the code to wrap a TaskCancelled later.

jskeet: I'd argue that this *is* a gRPC context, in terms of the abstraction GAX exposes - we're implementing a gRPC CallInvoker. So it makes sense to talk in gRPC concepts - just like we translate HTTP status codes into gRPC ones.

Even though we generate one CancellationToken, there's another which may be present in the CallOptions... it looks like we're not observing that at the moment, which also needs to be fixed. (The TODO was meant to cover all aspects of CallOptions; a deadline is only part of that.) But we can probably differentiate between the cancellation token *we've* created (just for the deadline) and one that has been provided by the caller. We can check whether our deadline-oriented cancellation token has been cancelled, and translate that into an RpcException, but leave anything else.

viacheslav-rostovtsev: Updated to account for this

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -60,9 +62,18 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
             var restMethod = _serviceCollection.GetRestMethod(method);
 
             var cancellationTokenSource = new CancellationTokenSource();
+            if (options.Deadline.HasValue)
+            {
+                // TODO: [virost, 2021-12] Use IClock.
+                var delayInterval = options.Deadline.Value - DateTime.UtcNow;
+                if (delayInterval.TotalMilliseconds <= 0)
+                {
+                    throw new TimeoutException($"The timeout was reached when calling a method `{restMethod.FullName}`");
```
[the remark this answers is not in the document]
amanda-tarafa: Hmmm, if we were going to be that strict, then we should continue to inspect the deadline even after the HTTP request returned succesfully, i.e. while we parse the response, etc.
I think a deadline so small that we cannot even do prep for the part of the process that takes the significant amount of time could be considered a bug, and different from a "really small deadline" that some times will not be enough to complete the time consuming part of the process. And I think we could signal that differently.

But I'm happy to disagree on this.

amanda-tarafa (commented):
I left these comments earlier but forgot to submit the review...

Google.Api.Gax.Grpc/Rest/RestChannel.cs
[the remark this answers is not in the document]
jskeet: It's not a matter of being strict as such - we're only throwing something because we know the deadline has already passed before we make the RPC; this is only a check to make sure we don't pass an invalid argument to `Task.Delay`.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -83,10 +93,25 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
+            await AddAuthHeadersAsync(httpRequest, restMethod, options, deadlineToken).ConfigureAwait(false);
+ 
+            HttpResponseMessage httpResponseMessage;
+            using (CancellationTokenSource linkedCts = CancellationTokenSource.CreateLinkedTokenSource(options.CancellationToken, deadlineToken))
+            {
+                try
+                {
+                    httpResponseMessage = await _httpClient.SendAsync(httpRequest, HttpCompletionOption.ResponseContentRead, linkedCts.Token).ConfigureAwait(false);
+                }
+                catch(TaskCanceledException)
```
jskeet, line 105: Nit: space after catch. But actually, we can simplify this anyway:

```csharp
catch (TaskCanceledException) when (deadline.IsCancellationRequested)
{
    throw new RpcException(new Status(StatusCode.DeadlineExceeded, $"The timeout was reached when calling a method `{restMethod.FullName}`"));
}
```

No need to rethrow explicitly.

jskeet (approved):
Okay, I think I'm now happy barring one nit. Would like approval from @amanda-tarafa as well though.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -102,14 +127,28 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
+            var deadlineTask = Task.Delay(-1, deadlineToken);
+            var clientCancellationTask = Task.Delay(-1, options.CancellationToken);
+            var channelTask = _channelAuthInterceptor(context, metadata);
+            var resultTask = await Task.WhenAny(channelTask, deadlineTask, clientCancellationTask).ConfigureAwait(false);
+            if (deadlineTask == resultTask)
+            {
+                throw new RpcException(new Status(StatusCode.DeadlineExceeded, $"Timeout was reached when waiting for auth headers to be added for a method `{restMethod.FullName}`."));
+            }
+
+            if (clientCancellationTask == resultTask)
```
amanda-tarafa, line 147: channelTask may have failed as well, in which case we are not failing, we are just not adding any headers, I guess we should buble the channelTask failure if there was any?

I wonder, why don't you pass the linked token as a parameter here and then do something like:

if (resultTask.IsFaulted)
{
    throw resultTask.Exception; 
}

And then you catch that the same way you catch the one in the calling _httpClient.SendAsync(...). If you want to keep the different messages you can set a flag after you have added the headers to diferentiate which operation threw.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -83,10 +93,23 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
             httpRequest.Headers.Add(VersionHeaderBuilder.HeaderName, RestVersion);
 
-            // TODO: How do we cancel this?
-            await AddAuthHeadersAsync(httpRequest, restMethod).ConfigureAwait(false);
+            var addAuthHeadersTask = AddAuthHeadersAsync(httpRequest, restMethod);
+            var delayTask = Task.Delay(-1, cancellationToken);
+            if (delayTask == await Task.WhenAny(addAuthHeadersTask, delayTask).ConfigureAwait(false))
+            {
+                throw new TimeoutException($"Timeout was reached when waiting for auth headers to be added for a method `{restMethod.FullName}`.");
+            }
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: I've implemented a version of this.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -102,14 +121,33 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
+            {
+                throw new TaskCanceledException(
+                    $"The task was cancelled by the caller while waiting for the auth headers to be added for a method `{restMethod.FullName}`.");
+            }
+
+            if (channelTask.IsFaulted)
+            {
+                if (channelTask.Exception is object)
+                {
+                    throw channelTask.Exception;
```
jskeet, line 145: I think we'd want `ExceptionDispatchInfo.Capture(channelTask.Exception).Throw();` here, to preserve the stack trace.

jskeet (approved):
This is looking good to me - but Amanda has been better at spotting problems than I have on this one, so I'll defer to her. Nearly there!

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -102,14 +121,33 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
+                }
+
+                throw new InvalidOperationException($"An unknown error has occurred when adding auth headers to a method `{restMethod.FullName}`.");
```
amanda-tarafa, line 148: Does this happen? Can the task be faulted and the Exception be null? I would have though not.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -102,14 +121,33 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
-            await _channelAuthInterceptor(context, metadata).ConfigureAwait(false);
+            var context = new AuthInterceptorContext(schemeAndAuthority, restMethod.FullName);
+
+            var combinedCancellationTask = Task.Delay(-1, combinedCancellationToken);
+            var channelTask = _channelAuthInterceptor(context, metadata);
+            var resultTask = await Task.WhenAny(channelTask, combinedCancellationTask).ConfigureAwait(false);
+            if (combinedCancellationTask == resultTask)
+            {
+                throw new TaskCanceledException(
+                    $"The task was cancelled by the caller while waiting for the auth headers to be added for a method `{restMethod.FullName}`.");
```
amanda-tarafa, line 138: nit: By the caller, or because we hit the deadline, right?

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -102,14 +121,33 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
...
 
             var metadata = new Metadata();
-            var context = new AuthInterceptorContext(schemeAndAuthority, method.FullName);
-            await _channelAuthInterceptor(context, metadata).ConfigureAwait(false);
+            var context = new AuthInterceptorContext(schemeAndAuthority, restMethod.FullName);
+
+            var combinedCancellationTask = Task.Delay(-1, combinedCancellationToken);
+            var channelTask = _channelAuthInterceptor(context, metadata);
+            var resultTask = await Task.WhenAny(channelTask, combinedCancellationTask).ConfigureAwait(false);
+            if (combinedCancellationTask == resultTask)
```
amanda-tarafa, line 135: I think that if this is the case, then `resultTask` will be faulted and `resultTask.Exception` will contain a `TaskCanceledException`. The only possible state in which combinedCancellationTask may run to completion is faulted.

Or am I missing something?

My point is that I think lines from 135 to 149 may be substituted by:

```csharp
if(resultTask.IsFaulted)
{
    ExceptionDispatchInfo.Capture(resultTask.Exception).Throw();
}
```
Or what am I missing?

jskeet: No, I think you're right. (I thought before that we'd want to provide a different exception on timeout for auth headers, but in this case I think it's fine for the task cancellation to just propagate.)

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: I don't *think* so.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: The `Task.Delay(-1, combinedCancellationToken)` does not get faulted and does not have an exception when the cancellation gets called on the token. It gets the status: Cancelled and that's it.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: If we hit a deadline, we'll wrap this in the exception that says that the timeout was reached. If we did not hit the deadline, there is no need for this `or`.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: The docs agree. Fixed.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: Ah, then you can change lines from 135 to 149 with:

```
await resultTask;
```

This will throw the TaskCancelledException if cancellation happened, or whatever exception happened if the real task failed.
If you get past this line of code, then there was no error or cancellation.

I think that will be cleaner and easier to understand than all the code we have now.

viacheslav-rostovtsev: Yes, this makes sense. Updated.

amanda-tarafa (approved):
LGTM! Thanks, it looks really way simpler now.

jskeet (approved):
Great - I'll squash and merge.

PR #517, opened 2021-12-30
feat: runtime support for explicit routing headers

viacheslav-rostovtsev:
Explicit routing header construction might require matching one or multiple regexes to the field and extracting part of it to put into the header.
This also includes adding Routing annotation class to the `Google.Api.CommonProtos` which might be better as a separate PR.


jskeet:
It's not clear to me why we want to use regular expressions. The routing headers are always generated from resource name templates, aren't they? We already have support for those in both GAX and the generator, and we can extract the relevant parts from the result of parsing a resource name (including failing gracefully). What benefit is there in adding a different way of handling resource name templates?

Google.Api.Gax.Grpc/ApiServerStreamingCall.cs
```diff
@@ -107,5 +108,33 @@ public sealed class ApiServerStreamingCall<TRequest, TResponse>
...
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="parameterName"></param>
+        /// <param name="extractionRegex"></param>
+        /// <param name="valueSelector"></param>
+        /// <returns></returns>
+        public ApiServerStreamingCall<TRequest, TResponse> WithExtractedGoogleRequestParam(string parameterName, Regex extractionRegex, Func<TRequest, string> valueSelector) =>
```
jskeet, line 119: How would you feel about creating the HeaderParameterExtractor in the generated code instead, so we only need to expose a single method here?

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public sealed class HeaderParameterExtractor<TRequest>
```
jskeet, line 18: I'd suggest making this immutable: pass all the regexes and all the selectors into the constructor in one go (possibly having an overload for the single-pattern version if that simplifies some generation situations). We can then create a `List(Regex, Func<TRequest, string>))` which has already been reversed. That feels better than doing the zipping and reversing on every request. (And immutability generally feels nicer...)

jskeet (commented):
I still need to consider the whole regex aspect, but if we do go with the regex form, I'm fine with the overall design - with a couple of suggestions.

I'll now take a slightly longer look at whether I still object to the use of regexes in the first place...

jskeet:
(Oh, and I agree we should do the CommonProtos change as a separate PR. I'm happy to do that and release a new version if that would be useful.)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public string Extract(TRequest request)
+        {
+            foreach (var (extractionRegex, valueSelector) in _regexes.Zip(_selectors, (regex, func) => (regex, func))
+                .AsEnumerable().Reverse())
+            {
+                var value = valueSelector(request);
+
+                if (extractionRegex.IsMatch(value))
```
jskeet, line 59: I suspect we can call `Match` once, check whether or not it was successful, then use `Value` if so. (At the moment we're matching twice.)

jskeet: Except I'm not sure we actually want the complete value of the matching regex, do we? Don't we sometimes only want one group within it? (I'm still trying to work that out...)

viacheslav-rostovtsev: 👍, taking a group

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: 👍

Google.Api.CommonProtos/Api/Routing.g.cs
```diff
@@ -0,0 +1,902 @@
+/*
```
jskeet, line 1: Note: if you rebase off main, this diff will go away.

Google.Api.Gax.Grpc/ApiCall.cs
```diff
@@ -185,5 +188,33 @@ public sealed class ApiCall<TRequest, TResponse>
...
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="parameterName"></param>
+        /// <param name="extractionRegex"></param>
+        /// <param name="valueSelector"></param>
+        /// <returns></returns>
+        public ApiCall<TRequest, TResponse> WithExtractedGoogleRequestParam(string parameterName, Regex extractionRegex, Func<TRequest, string> valueSelector) =>
```
jskeet, line 199: Presumably this one will go away too?

jskeet (commented):
It looks like you haven't actually pushed your change...

Google.Api.Gax.Grpc/ApiCall.cs
viacheslav-rostovtsev: 👍

viacheslav-rostovtsev:
Yes, I totally forgot to push /facepalm. Rebased and pushed.

Google.Api.Gax.Grpc/ApiCall.cs
```diff
@@ -185,5 +186,24 @@ public sealed class ApiCall<TRequest, TResponse>
             GaxPreconditions.CheckNotNull(valueSelector, nameof(valueSelector));
             return WithCallSettingsOverlay(request => CallSettings.FromGoogleRequestParamsHeader(parameterName, valueSelector(request)));
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="extractor"></param>
+        /// <returns></returns>
+        public ApiCall<TRequest, TResponse> WithExtractedGoogleRequestParam(HeaderParameterExtractor<TRequest> extractor)
```
jskeet, line 195: We might want to consider moving the body of the lambda expression to HeaderParameterExtractor, to avoid the duplication between this and ApiServerStreamingCall (and others?). That can be a later refactor though.

Google.Api.Gax.Grpc/CallSettings.cs
```diff
@@ -325,6 +327,14 @@ public static CallSettings FromFieldMask(string fieldMask)
         internal static CallSettings FromGoogleRequestParamsHeader(string parameterName, string value) =>
             FromHeader(RequestParamsHeader, parameterName + "=" + Uri.EscapeDataString(value ?? ""));
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="paramNameValues"></param>
+        /// <returns></returns>
+        internal static CallSettings FromGoogleRequestParamsHeader(Dictionary<string, string> paramNameValues) =>
+            FromHeader(RequestParamsHeader, string.Join("&", paramNameValues.Select(nameVal => nameVal.Key + "=" + Uri.EscapeDataString(nameVal.Value ?? ""))));
```
jskeet, line 336: Is the order important here? Would it make sense to take a list of key/value pairs instead? (Potentially even just a list of values that is formatted by the HeaderParameterExtractor?)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
jskeet: Glad to to see it's now immutable - but we're still doing a lot of work on a per request basis. It might be easier for me to suggest a refactoring after merging though.

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public Dictionary<string, string> Extract(TRequest request)
+        {
+            return _extractions.GroupBy(pe => pe.ParamName)
```
jskeet, line 57: This is the bit that feels like it can be done once on construction instead of repeatedly. It may require keeping more state than we would otherwise do (so that we've got the original extractions as well as the "processed" version) but I doubt that'll be a problem.

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="paramName"></param>
+        /// <param name="regex"></param>
+        /// <param name="selector"></param>
+        /// <returns></returns>
+        public HeaderParameterExtractor<TRequest> Add(string paramName, Regex regex, Func<TRequest, string> selector)
```
jskeet, line 45: Rename to `WithParameter`? (`Add` sounds like it modifies the existing one.)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                ParamName = paramName;
+                _regex = regex;
+                _selector = selector;
+            }
+
+            public string Extract(TRequest request)
+            {
+                var match = _regex.Match(_selector(request) ?? "");
+                return match.Success ? match.Groups[1].Value : null;
```
jskeet, line 82: This requires that the regex itself *has* one capturing group, yes? We'll want to make that really clear in the documentation.

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
```diff
@@ -132,8 +136,310 @@ public void WithGoogleRequestParam()
             call1.Sync(new SimpleRequest { Name = "test" }, null);
             call1.Async(new SimpleRequest { Name = "test" }, null);
 
-            CallSettingsTest.AssertSingleHeader(syncCallSettings, CallSettings.RequestParamsHeader, "parent=test");
-            CallSettingsTest.AssertSingleHeader(asyncCallSettings, CallSettings.RequestParamsHeader, "parent=test");
+            var expectedHeader = "parent=test";
```
jskeet, line 139: I'd suggest doing the vast majority of these tests as unit tests of HeaderParameterExtractor - that keeps the test closer to the logic, and will reduce the setup required (and duplication of testing with sync and async). I would expect just a single test for each of ApiCall and ApiServerStreamingCall would be sufficient.

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
```diff
@@ -132,8 +136,310 @@ public void WithGoogleRequestParam()
...
+        }
+
+        /// <summary>
+        /// Extracting multiple routing header key-value pairs by matching
+        /// several non-conflicting path templates on (parts of) a single request field.
+        /// Make the templates strict, so that if the `table_name` does not
+        /// have an instance information, nothing is sent.
+        /// </summary>
+        [Theory]
+        [InlineData("projects/100/instances/200/tables/300", "project_id=projects/100&instance_id=instances/200")]
```
jskeet, line 259: This test is sensitive to the order in which they're reported by the dictionary - this sort of thing is why I think would be better to just keep a list of parameter/value pairs, or put everything into the parameter extractor (it could just produce the complete header value).

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
```diff
@@ -132,8 +136,310 @@ public void WithGoogleRequestParam()
...
+                (req, cs) => { asyncCallSettings = cs; return null; },
+                (req, cs) => { syncCallSettings = cs; return null; },
+                null);
+
+            // call corresponding to the following routing parameters:
+            // { field: "table_name", path_template: "{project_id=projects/*}/**" }
+            // { field: "app_profile_id", path_template: "{routing_id=**}" }
+            var call1 = call0
+                .WithExtractedGoogleRequestParam(
+                    new HeaderParameterExtractor<ExtractedRequestParamRequest>("project_id",
```
jskeet, line 344: If we had a parameterless constructor, and could accept strings instead of Regex (maybe just as a convenience) I think we could format these tests more simply (also simpler when just testing the extractor):

```csharp
var extractor = new HeaderParameterExtractor<ExtractedRequestParamRequest>()
    .WithParameter("project_id", "^(?<project_id>projects/[^/]+)(?:/.*)?$", request => request.TableName)
    .WithParameter("routing_id", "^(?<routing_id>.*)$", request => request.AppProfileId);
```

It could be even shorter (and less repetitive) if we either omitted the group name or used it as the parameter name...

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="paramName"></param>
+        /// <param name="regex"></param>
+        /// <param name="selector"></param>
+        public HeaderParameterExtractor(string paramName, Regex regex, Func<TRequest, string> selector)
+        {
+            _extractions = new List<ParamExtraction> { new ParamExtraction(paramName, regex, selector) };
```
jskeet, line 30: All the tests show the parameter name being the same as the group name. If that's what we expect to do, let's build that in and simplify things:

- Call regex.GetGroupNames()
- Assert that the result has a single value
- Use the value as the parameter name

jskeet (commented):
Getting a lot closer to what I'd like :) I don't know how much we want to iterate on this PR vs getting it in and then I can suggest changes via additional PRs... I'm happy to be guided by you on that.

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/ApiCall.cs
viacheslav-rostovtsev: This one I am happy to leave to the later refactor. I would like to separate combining the header parameter from extracting them as well, and I was not able to get a clean version that I would like there.

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: Done, in the documentation and added the check for it.

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
viacheslav-rostovtsev: My counter-argument would be that if we decide to move around parts when refactoring -- e.g. where does the lambda go, who is responsible for merging and escaping the parameters --  this file will give us a complete suite of tests that won't have to be modified, which is very helpful. That is why I would prefer to keep these here.

I'm not 100% married to the idea, so if you still think these should be moved I am open to it.

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
viacheslav-rostovtsev: I've fixed the test to not be order sensitive. Keeping the string here because (a) it is simpler to read and (b) parity with Ruby tests

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
viacheslav-rostovtsev: 👍  done

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: 👍 done

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
jskeet: I really think it'll be simpler to provide good coverage with focused tests. We should definitely keep *some* tests in here, as effectively integration tests - but they don't need to cover every detail.

I'm happy to make those test changes as a later refactoring PR though.

Google.Api.Gax.Grpc.Tests/CallSettingsTest.cs
```diff
@@ -273,7 +274,7 @@ internal static void AssertSingleHeader(CallSettings callSettings, string expect
             Assert.Equal(expectedHeaderValue, metadata[0].Value);
         }
 
-        internal static void AssertRoutingHeader(CallSettings callSettings, string expectedHeaderValue)
+        internal static void AssertRoutingHeaderIgnoringOrder(CallSettings callSettings, string expectedHeaderValue)
```
jskeet, line 277: To be honest I'd have preferred to make the order deterministic instead, but we can go with this for now.
(I'm hoping that once most tests have moved, everything will be simpler all round.)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -13,54 +13,89 @@
...
             _extractions = new List<ParamExtraction>(extractions);
+
+            // It is important that we Reverse the parameters within the group.
+            // The extractions follow the `last successfully matched wins` rule for
+            // conflict resolution when there are multiple extractions for the same parameter name.
+            // (see `google/api/routing.proto` for further details)
+            // But we would prefer to stop evaluating on the first successfully matched extraction,
+            // thus the reverse.
+            _groupings = _extractions.GroupBy(pe => pe.ParamName)
+                .ToDictionary(paramGroup => paramGroup.Key,
```
jskeet, line 45: Nit - I'd unwrap this last method call to a single line. I'm happy to do that later though.

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -13,54 +13,89 @@
...
+            GaxPreconditions.CheckNotNullOrEmpty(paramName, nameof(paramName));
+            GaxPreconditions.CheckNotNull(regexStr, nameof(regexStr));
+            GaxPreconditions.CheckNotNull(selector, nameof(selector));
+
+            var regex = new Regex(regexStr);
+
+            // All regexes have a capturing group named `0` that captures the whole regex
+            if (regex.GetGroupNames().Length != 2)
+            {
+                var errMsg =
```
jskeet, line 72: Nit: unwrap, or just inline it into the throw statement. (I don't think it's worth having a separate local variable.)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -13,54 +13,89 @@
...
-        /// <param name="regex"></param>
-        /// <param name="selector"></param>
+        /// <param name="paramName">The name of the parameter in the routing header.</param>
+        /// <param name="regexStr">The regular expression (in the string form) used to extract the value of the parameter.
+        /// Should have exactly one named capturing group.</param>
+        /// <param name="selector">A function to call on each request, to determine the string to extract the header value from.
+        /// The parameter must not be null, but may return null.</param>
         /// <returns></returns>
-        public HeaderParameterExtractor<TRequest> Add(string paramName, Regex regex, Func<TRequest, string> selector)
+        public HeaderParameterExtractor<TRequest> WithParameter(string paramName, string regexStr, Func<TRequest, string> selector)
```
jskeet, line 61: It's still like the parameter name to be inferred from the regex - I think that'll simplify calling code. Again, I'm happy to do that myself, but it'll need to be before the generator change is merged.

jskeet (approved):
I'm happy for this to be merged now. I'd then like to create a PR that:

- Makes most of the tests for HeaderParameterExtractor unit tests of *just* the extractor (leaving just a few tests for the integration of CallSettings and HeaderParameterExtractor)
- Uses the name of the capturing group in the regex as the parameter name, to avoid duplication
- Makes the extractor return the header value as a string, in a deterministic way with all necessary escaping already performed

(That could be three separate PRs, of course - or at the very least three separate commits so they can be individually accepted or rejected.)

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
viacheslav-rostovtsev: We can't have parameter name directly match regex group name, because we can have parameters with `.` in their name (actual case from Spanner), just like in the implicit headers. I can add an override of `WithParameter` that will work for the majority of the cases.

jskeet: Is that not a valid capturing group name? Okay... in that case, I'd argue we should move in the opposite direction: use an unnamed capturing group instead. (It still deduplicates, just in the opposite direction.)

Google.Api.Gax.Grpc.Tests/ApiCallTest.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc.Tests/CallSettingsTest.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/HeaderParameterExtractor.cs
```diff
@@ -13,54 +13,89 @@
...
+                    "The regex used for the routing header extraction should have exactly one named capturing group.";
+                throw new ArgumentException(errMsg, nameof(regex));
+            }
+
             return new HeaderParameterExtractor<TRequest>(_extractions.Concat(new[] { new ParamExtraction(paramName, regex, selector) }));
         }
 
         /// <summary>
-        /// 
+        /// Extracts the dictionary of parameter names and values to add to the routing header.
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

PR #526, opened 2022-02-09
Experimental: First pass at bidi streaming routing

jskeet:
This basically allows a "sample request" to be passed in the initial call. That request is only used for extracting routing headers.
I think this would work, but requires significant documentation.


jskeet:
Closing for now, but leaving the branch present to examine later.

PR #528, opened 2022-02-23
Prepare for next major version

jskeet:
Review one commit at a time as normal. In summary, the changes are:

- Version will be 4.x
- Target frameworks will be .NET 4.6.2 and .NET Standard 2.1 (for GAX; CommonProtos remains the same)
- Test frameworks will be .NET 4.6.2, .NET Core 3.1 and .NET 6
- Dependencies are updated
- GkePlatformDetails requires an odd change, but was probably slightly broken before on .NET 4.6.1

### DIFF
--- a/Google.Api.Gax.Grpc.GrpcNetClient.Tests/Google.Api.Gax.Grpc.GrpcNetClient.Tests.csproj
+++ b/Google.Api.Gax.Grpc.GrpcNetClient.Tests/Google.Api.Gax.Grpc.GrpcNetClient.Tests.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="..\CommonProperties.Test.xml" />
 
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Linq.Async" Version="5.0.0" />
+    <PackageReference Include="System.Linq.Async" Version="6.0.1" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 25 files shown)


Google.Api.Gax.Tests/Json/JsonTokenizerTest.cs
```diff
@@ -120,8 +120,10 @@ public void NumberValue(string json, double expectedValue)
         [InlineData("1e-")]
         [InlineData("--")]
         [InlineData("--1")]
-        [InlineData("-1.7977e308")]
-        [InlineData("1.7977e308")]
+        // TODO: Investigate these when we actually start using the JSON code.
```
amanda-tarafa, line 123: This is and AssertThrows, so the test failing means that this number format is now valid, which is not a bad thing.

I think it's OK to remove these two cases.

Google.Api.Gax.Grpc.GrpcNetClient.Tests/Google.Api.Gax.Grpc.GrpcNetClient.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
amanda-tarafa, line 1: tiny nit: Is this changing EOL?

amanda-tarafa (approved):
LGTM!

I would have thought we had more conditional code in Gax, specially for .net461. I'm probably really thinking about Google.Apis.Auth.

Google.Api.Gax.Tests/Json/JsonTokenizerTest.cs
jskeet: I think it's a bad thing if the numbers can't actually be represented due to overflow (which I believe was the issue). I'd like to look more closely at this later, so I'll leave them in if that's okay.

Google.Api.Gax.Grpc.GrpcNetClient.Tests/Google.Api.Gax.Grpc.GrpcNetClient.Tests.csproj
jskeet: More likely to be a BOM introduced or removed. But this project will be going away in the next PR anyway, so I won't bother changing it :)

Issue #529, opened 2022-02-23
Consider the GrpcAdapter.Default API


jskeet:
Should clients be able to set it manually? Clear the default?

amanda-tarafa:
I was also thinking more in terms of DI. But I think the entry point for that would be more on `ClientBuilderBase` and its implementations (which would benefit REGAPIC as well). Something like:

- Change `GrpcAdapter.Default` -> `GrpcAdapter.Fallback`, and keep the current heuristic of trying to create a channel + possibly and env variable to determine whether it's Grpc.Core or Grpc.Net.Client.

And o one of the following options:

1. At a minimum add a constructor to ClientBuilderBase that takes a GrpcAdapter, so the builder can be itself registered for DI, with an injected Adapter. If null, use `GrpcAdapter.Fallback`.
2. Or add static factory methods to the generated ClientBuilders (a little weird to have a factory for a builder) that take the Adapter and any other 
3. Or possibly the better option is to add DI extension methods for registering the builder and maybe even the client directly (even if we internally use the builder). This of course means that we need a dependency on MS's DI, which we might not want.
4. Do 3. but in a separate powerpack like library, generic on ClientBuilderBase and ClientBase, where we can take the DI dependency easily.
5. Do nothing in production code but write examples for approaches 2. (the factory methods to be outside of the classe) and 3.

I like 4, the best, what do you think?

jskeet:
Not sure yet... but I'll think about it. I suspect we should probably write a mini-PRD for this, so that we really know what we're trying to achieve...

jskeet:
Not sure yet... but I'll think about it. I suspect we should probably write a mini-PRD for this, so that we really know what we're trying to achieve...

jskeet:
We've now overhauled this a *lot*, and I think we're happy with it.

PR #537, opened 2022-02-24
Speculative approach to making the default adapter settable.

jskeet:
Is this too rigid?

This is only a draft PR, with no tests etc.

One motivating idea: it would be reasonably easy (I think) to set a Grpc.Net.Client-based adapter with a suitable `ILoggerFactory` on startup for an ASP.NET Core app. I could potentially prototype that before going any further, if you think it would help.

(More generally, if there's some good way we could hook into DI to avoid all this being static, that would be great...)


amanda-tarafa:
I'm leaving my thoughts on #529 because they are more general than a particular implementation. And yes, they are around trying to use DI.

PR #540, opened 2022-03-08
Enable error details in REGAPIC

jskeet:
Please review one commit at a time. If this all seems okay, I'll add a final commit with unit tests and we can merge.


Google.Api.Gax.Grpc/Rest/response_metadata.proto
```diff
@@ -0,0 +1,45 @@
...
+// Taken from https://cloud.google.com/apis/design/errors#error_mapping.
+// This allows detailed error information to be parsed from HTTP/1.1 responses
+// and converted into an appropriate google.rpc.Status. The message names
+// aren't ideal, but it's better to just take the proto definition directly
+// from the page than to modify it. This is only generated as internal classes.
+
+// This message defines the error schema for Google's JSON HTTP APIs.
+message Error {
+  // Deprecated. This message is only used by error format v1.
+  message ErrorProto {}
```
amanda-tarafa, line 23: Why don't you add the annotation?

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/Rest/response_metadata.proto
jskeet: Because it's not in the proto in the referenced page. If I were going to change what's there, I'd change a lot more :) I figured that copy/paste would be the simplest way of helping us stay up to date with changes. This is only generated as an internal class anyway, so it has no public impact.

amanda-tarafa: Ah yes, OK.

jskeet:
This is now ready for full review: @amanda-tarafa please review the final commit (that's the only change since your last review); @viacheslav-rostovtsev please review all commits.

viacheslav-rostovtsev (approved):
LGTM

PR #541, opened 2022-03-08
feat: Allow gRPC adapter detection to be overridden

jskeet:
This allows an environment variable with the name
"GRPC_DEFAULT_ADAPTER_OVERRIDE" to determine the adapter to use:
- a value of "Grpc.Core" will use the GrpcCoreAdapter
- a value of "Grpc.Net.Client" will use the GrpcNetClientAdapter
- a missing or empty/whitespace value will use normal detection
- any other value will throw an exception

### DIFF
--- a/Google.Api.Gax.Grpc.IntegrationTests/GrpcAdapterTest.cs
+++ b/Google.Api.Gax.Grpc.IntegrationTests/GrpcAdapterTest.cs
@@ -6,6 +6,7 @@
  */
 
 using Grpc.Core;
+using System;
 using Xunit;
 using static Google.Api.Gax.Grpc.IntegrationTests.TestService;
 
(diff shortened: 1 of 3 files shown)


jskeet:
Note: there are no tests for this at the moment as it's a "one shot" approach. I could potentially add a "ClearForTest" method to reset the lazy variable, if that would be deemed appropriate.

We may wish to rename the environment variable if we start viewing this as a fallback instead of as a default. (I'm about to start looking at how DI might work with gRPC adapters...)

amanda-tarafa (approved):
LGTM

Test wise, maybe we should continue to run at least some of our library integration tests with Grpc.Core, in which case we can just have a different CI job that sets the env variable to Grpc.Core?

amanda-tarafa (approved):
LGTM

Issue #543, opened 2022-03-10
Should we include an ILogger as part of ApiCall etc?


jskeet:
For GAX v4, we've got a dependency on the logging abstractions anyway.

If we added ILogger to ServiceSettingsBase (or potentially ClientBuilderBase, or both) we could add per-client logging, including retries etc.

If we added an ILoggerFactory instead, we could potentially create the logger ourselves with the right client name.

jskeet:
(We should probably prototype this before v4, to see whether we need to implement it properly *before* v4 ships or whether it could be added afterwards.)

jskeet:
After a few minutes of looking, we'd *either* need to add an `ILoggerFactory` or `ILogger` to `ServiceSettingsBase`, *or* add an `ILogger` parameter to the constructor of each client implementation type. The latter would probably be the most sensible approach, to be honest - and would definitely need to be done before the major release.

jskeet:
Weird idea: we could include ILogger in CallSettings. *Probably* not a good idea, but worth at least considering...

jskeet:
(CallSettings would actually make this somewhat easier to plumb through, although we'd also want the method name to be plumbed through too, which is trickier...)

jskeet:
Interesting decision: a single logger for the client, or a logger per method? The latter makes it easier to give different log levels to different methods...

jskeet:
Ha! Good point, yes. Not entirely sure the best way of doing that, but we can come to it later...

jskeet:
This was implemented in #559.

PR #544, opened 2022-03-10
Prototype of logging

jskeet:
This is for #543, and builds on #542.

It would require some changes in client generation, so we need to decide this before shipping GAX v4.

Currently it doesn't log very much - just "call starting", "call finished", "backing off before retry" but we could definitely add more.


jskeet:
It's gRPC-adapter agnostic - so if the gRPC adapter and the client builder are both specified via DI, I'd expect them to have the same LoggerFactory. But they don't absolutely have to.

jskeet:
Closing in favour of #558.

PR #545, opened 2022-03-17
Second pass at making GAX more DI-friendly

jskeet:
End-user code:

```csharp
builder.Services.AddGrpcNetClientAdapter();
builder.Services.AddGoogleTranslationServiceClient();
```

Translation service code:

```csharp
namespace Google.Cloud.Translate.V3
{
    public partial class TranslationServiceClientBuilder
    {
        internal static TranslationServiceClient ConfigureAndBuild(IServiceProvider provider, Action<TranslationServiceClientBuilder> configure)
        {
            var builder = new TranslationServiceClientBuilder();
            configure?.Invoke(builder);
            builder.Configure(provider);
            return builder.Build();
        }
    }
}

namespace Microsoft.Extensions.DependencyInjection
{
    using Google.Cloud.Translate.V3;

    public static class TranslationServiceClientExtensions
    {
        public static IServiceCollection AddGoogleTranslationServiceClient(this IServiceCollection collection, Action<TranslationServiceClientBuilder> configure) =>
            collection.AddSingleton(provider => TranslationServiceClientBuilder.ConfigureAndBuild(provider, configure));

        public static IServiceCollection AddGoogleTranslationServiceClient(this IServiceCollection collection) =>
            AddGoogleTranslationServiceClient(collection, null);
    }
}
```

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -8,6 +8,7 @@
 using Google.Apis.Auth.OAuth2;
 using Grpc.Auth;
 using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -507,6 +508,60 @@ namespace Google.Api.Gax.Grpc
         /// Builds the resulting client.
         /// </summary>
         public abstract TClient Build();
+
+        /// <summary>
+        /// Populates properties based on those set via dependency injection.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Credentials are only requested from dependency injection if they are not already set
+        /// via any of <see cref="ChannelCredentials"/>, <see cref="CredentialsPath"/>,
+        /// <see cref="JsonCredentials"/>, <see cref="Credential"/>, <see cref="GoogleCredential"/>
+        /// or <see cref="TokenAccessMethod"/>.
+        /// </para>
+        /// <para>
+        /// If credentials are requested, they are tried in the following order:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>ChannelCredentials</item>
+        /// <item>ICredential</item>
+        /// <item>GoogleCredential</item>
+        /// </list>
+        /// </remarks>
+        /// <param name="provider">The service provider to request dependencies from.</param>
+        protected virtual void Configure(IServiceProvider provider)
+        {
+            GaxPreconditions.CheckNotNull(provider, nameof(provider));
+
+            CallInvoker ??= provider.GetService<CallInvoker>();
+            if (CallInvoker is object)
+            {
+                return;
+            }
+
+            // TODO: Check for the concrete adapters as well, based on what is supported?
+            GrpcAdapter ??= provider.GetService<GrpcAdapter>();
+            GrpcChannelOptions ??= provider.GetService<GrpcChannelOptions>();
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            if (ChannelCredentials is null && CredentialsPath is null && JsonCredentials is null && TokenAccessMethod is null &&
+#pragma warning restore CS0618 // Type or member is obsolete
+                GoogleCredential is null && Credential is null)
+            {
+                if (provider.GetService<ChannelCredentials>() is ChannelCredentials channelCredentials)
+                {
+                    ChannelCredentials = channelCredentials;
+                }
+                else if (provider.GetService<ICredential>() is ICredential credential)
+                {
+                    Credential = credential;
+                }
+                else if (provider.GetService<GoogleCredential>() is GoogleCredential googleCredential)
+                {
+                    GoogleCredential = googleCredential;
+                }
+            }
+        }
 
         // Note: The implementation is responsible for performing validation before constructing the client.
         // The Validate method can be used as-is for most builders.
--- a/Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
+++ b/Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
@@ -1,0 +1,64 @@
+﻿/*
+ * Copyright 2022 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Google.Api.Gax.Grpc;
+using Google.Api.Gax.Grpc.Rest;
+using Grpc.Net.Client;
+using System;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods for dependency injection.
+    /// </summary>
+    public static class GaxGrpcServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds a singleton <see cref="GrpcNetClientAdapter"/> to the given service collection
+        /// as the preferred <see cref="GrpcAdapter"/> implementation,
+        /// using the default instance with any additional options configured via <paramref name="optionsConfigurer"/>.
+        /// Note that unlike the <see cref="AddGrpcNetClientAdapter(IServiceCollection)"/> overload without an action,
+        /// this does not implicitly set the <see cref="Grpc.Net.Client.GrpcChannelOptions.ServiceProvider"/> to the provider.
+        /// </summary>
+        /// <param name="services">The service collection to add the adapter to.</param>
+        /// <param name="optionsConfigurer">The configuration action to perform on each <see cref="Grpc.Net.Client.GrpcChannelOptions"/>
+        /// when it is used by the adapter to construct a channel.</param>
+        /// <returns>The same service collection reference, for method chaining.</returns>
+        public static IServiceCollection AddGrpcNetClientAdapter(this IServiceCollection services, Action<IServiceProvider, Grpc.Net.Client.GrpcChannelOptions> optionsConfigurer) =>
+            services.AddSingleton<GrpcAdapter>(provider => GrpcNetClientAdapter.Default.WithAdditionalOptions(options => optionsConfigurer(provider, options)));
+
+        /// <summary>
+        /// Adds a singleton <see cref="GrpcNetClientAdapter"/> to the given service collection
+        /// as the preferred <see cref="GrpcAdapter"/> implementation,
+        /// such that any <see cref="GrpcChannel"/> created uses the service provider from
+        /// created this service collection. This enables logging, for example.
+        /// </summary>
+        /// <param name="services">The service collection to add the adapter to.</param>
+        /// <returns>The same service collection reference, for method chaining.</returns>
+        public static IServiceCollection AddGrpcNetClientAdapter(this IServiceCollection services) =>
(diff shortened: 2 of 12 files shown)


Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            services.AddSingleton<GrpcAdapter>(provider => GrpcNetClientAdapter.Default.WithAdditionalOptions(options => optionsConfigurer(provider, options)));
+
+        /// <summary>
+        /// Adds a singleton <see cref="GrpcNetClientAdapter"/> to the given service collection,
+        /// such that any <see cref="GrpcChannel"/> created uses the service provider from
+        /// created this service collection. This enables logging, for example.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddGrpcNetClientAdapter(this IServiceCollection services) =>
```
amanda-tarafa, line 36: For solving the REST Adapter bits and following with the factory idea (restricting visibility as much as possible). Also, quickly here but I'm happy to push a PR for this.

```csharp
public interface IAdapterFactory { GrpcAdapter Get(GrpcAdapterOptions options); }

public class GrpcAdapterOptions
{ 
    public IEnumerable<FileDescriptors> Descriptors { get; set; }
    public SupportedFormat SupportedFormats { get; set; }

    [Flags] // Possibly better names here.
    public enum SupportedFormat { GRpc, Json}
}

// And a similar one for Grpc.Core if we wanted to.
public class GrpcNetClientAdapterFactory : IAdapterFactory
{
    private readonly GrcpNetClientAdapter _instance;

    public GrpcNetClientAdapterFactory(IServiceProvider provider) =>
        _instance = GrcpNetClientAdapter.Default.WithAdditionalOptions(options => options.ServiceProvider = provider);

    public GrpcAdapter Get(GrcpAdapterOptions options) => 
         options.SupportedFormats.HasFlag(SupportedFormat.Grpc) ? _instance : null;
}

public class RestGrpcAdapterFactory : IAdapterFactory
{
    private readonly ConcurrentDictionary<GrpcAdapterOptions, RestGrpcAdapter> _cache = new ....;
    
    // No need for explicit constructor.

    public GrpcAdapter Get(GrpAdapterOptions options) =>
        // if the supported format includes JSON, find or add the adapter on the cache using the descriptor, else return null.
}
```

On the extension methods we always inject the factories instead of the adapters.
On the client builders we look for the factories and always pass the descriptors (which we know, even if we have to add a new abstract property) and the formats we can support (i.e. Compute only supports JSON, gapics only support gRPC for now but will support JSON at some point). If we don't get a factory or an adapter, that's fine, we don't set it and default will be used.

amanda-tarafa (commented):
Yep, I like this very much.

A couple of suggestions on the remaining loose ends.

Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
jskeet: Maybe. I think we can defer that part a little... I'm slightly nervous about having too many abstractions. We could potentially make the single GrpcAdapter method that's used to create a ChannelBase accept the additional option of the file descriptors, and just ignore them for other adapters. Not sure...

Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public static class GaxGrpcServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds a singleton <see cref="GrpcNetClientAdapter"/> to the given service collection,
+        /// using the default with any additional options configured via <paramref name="optionsConfigurer"/>.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="optionsConfigurer"></param>
+        /// <returns></returns>
+        public static IServiceCollection AddGrpcNetClientAdapter(this IServiceCollection services, Action<IServiceProvider, global::Grpc.Net.Client.GrpcChannelOptions> optionsConfigurer) =>
```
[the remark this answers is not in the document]
jskeet: Done.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -481,6 +482,44 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
+
+            if (ChannelCredentials is null && CredentialsPath is null && JsonCredentials is null && TokenAccessMethod is null)
+            {
+                ChannelCredentials ??= provider.GetService<ChannelCredentials>();
+
+                if (ChannelCredentials is null)
+                {
+                    // TODO: There are other interfaces we could use here... as well as concrete credential classes.
+                    // TODO: Figure out how/when to apply the "default" behavior around JWTs, scopes etc.
+                    //       Possibly use the Options pattern? Lots of details to work out there.
```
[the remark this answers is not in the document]
jskeet: We've now got GoogleCredential separately from ICredential, via a different PR. I *believe* that makes all of this moot.

jskeet:
The production code is now ready for review. The tests should be easy enough to add afterwards, in terms of precedence etc - but they'd all change if we change that precedence, so I'll add them in another commit :)

jskeet:
Ooh - had missed that we could set the IHttpClientFactory in the REST builder base...

jskeet:
There are now tests in a second commit.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -508,6 +509,60 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
+        {
+            GaxPreconditions.CheckNotNull(provider, nameof(provider));
+
+            CallInvoker ??= provider.GetService<CallInvoker>();
+            if (CallInvoker is object)
+            {
+                return;
+            }
+
+            // TODO: Check for the concrete adapters as well, based on what is supported?
```
amanda-tarafa, line 542: I think we should use `IEnumerable<GrpcAdapter> GetServices<GrpcAdapter>` and grab the first one that supports this client. We can ask users to always register as the base type and we do so ourselves.

If we ask for specific types here, and we ever open the adapters for extension, then that'll be awkward.

Google.Api.Gax.Grpc/GaxGrpcServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Adds a singleton <see cref="GrpcNetClientAdapter"/> to the given service collection
+        /// as the preferred <see cref="GrpcAdapter"/> implementation,
+        /// using the default instance with any additional options configured via <paramref name="optionsConfigurer"/>.
+        /// </summary>
+        /// <param name="services">The service collection to add the adapter to.</param>
+        /// <param name="optionsConfigurer">The configuration action to perform on each <see cref="Grpc.Net.Client.GrpcChannelOptions"/>
+        /// when it is used by the adapter to construct a channel.</param>
+        /// <returns>The same service collection reference, for method chaining.</returns>
+        public static IServiceCollection AddGrpcNetClientAdapter(this IServiceCollection services, Action<IServiceProvider, Grpc.Net.Client.GrpcChannelOptions> optionsConfigurer) =>
+            services.AddSingleton<GrpcAdapter>(provider => GrpcNetClientAdapter.Default.WithAdditionalOptions(options => optionsConfigurer(provider, options)));
```
amanda-tarafa, line 30: Should't you add here the provider as well to the ChannelOptions, as there's no guarantee that the configurer action will do that.
Wether you do it before or after the configurer function executes, I'm fine with either, tending to doing it before to allow users to override that for some weird case.

Google.Api.Gax.Grpc/GrpcNetClientAdapter.cs
```diff
@@ -16,22 +17,40 @@ namespace Google.Api.Gax.Grpc
...
+        /// <returns>A new adapter based on this one, but with an additional channel options configuration action.</returns>
+        public GrpcNetClientAdapter WithAdditionalOptions(Action<global::Grpc.Net.Client.GrpcChannelOptions> configure) =>
+            new GrpcNetClientAdapter(_optionsConfigurationAction + configure);
+
         /// <inheritdoc />
         private protected override ChannelBase CreateChannelImpl(ServiceMetadata serviceMetadata, string endpoint, ChannelCredentials credentials, GrpcChannelOptions options)
         {
             var grpcNetClientOptions = ConvertOptions(credentials, options);
+            _optionsConfigurationAction?.Invoke(grpcNetClientOptions);
+            grpcNetClientOptions.LoggerFactory?.CreateLogger("Grpc.Net.Client.Internal.GrpcCall").LogTrace("Awooga");
```
amanda-tarafa, line 53: nit: remove

Google.Api.Gax.Grpc/GrpcNetClientAdapter.cs
```diff
@@ -16,22 +17,40 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public sealed class GrpcNetClientAdapter : GrpcAdapter
     {
+        private Action<global::Grpc.Net.Client.GrpcChannelOptions> _optionsConfigurationAction;
+
         // Note: this is "Default" rather than "Instance" as we expect to have other factory methods later, e.g. accepting
         // an HTTP client factory.
+        // TODO: Rename now that there's GrpcAdapter.DefaultAdapter?
```
amanda-tarafa, line 24: I think it's fine to have `GrpcAdapter.DefaultAdapter` and then this one to be `Fallback` and for GrpcCore and Rest, `Instance`.

Google.Api.Gax.Rest/ClientBuilderBase.cs
```diff
@@ -241,6 +242,43 @@ protected async virtual Task<IConfigurableHttpClientInitializer> GetHttpClientIn
...
+        /// <para>
+        /// If credentials are requested, they are tried in the following order:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>ICredential</item>
+        /// <item>GoogleCredential</item>
+        /// </list>
+        /// </remarks>
+        /// <param name="provider">The service provider to request dependencies from.</param>
+        protected virtual void Configure(IServiceProvider provider)
```
amanda-tarafa, line 263: We need to remember to handwrite the extension methods for BigQuery and Storage clients.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -508,6 +509,60 @@ protected virtual GrpcChannelOptions GetChannelOptions()
...
+        /// If credentials are requested, they are tried in the following order:
+        /// </para>
+        /// <list type="bullet">
+        /// <item>ChannelCredentials</item>
+        /// <item>ICredential</item>
+        /// <item>GoogleCredential</item>
+        /// </list>
+        /// </remarks>
+        /// <param name="provider">The service provider to request dependencies from.</param>
+        protected virtual void Configure(IServiceProvider provider)
```
amanda-tarafa, line 532: We need to remember to handwrite the extension methods for the libraries that have builder wrappers, I think PubSub is one of them?

amanda-tarafa (approved):
LGTM with a couple of comments.

jskeet (commented):
Thanks - will merge as a single commit.
We can revisit all the bits we're still not sure on.

PR #548, opened 2022-03-23
Prototype of change for GrpcAdapter to support REST

jskeet:
This bundles everything into GrpcAdapter; we could potentially
separate it out so that GrpcAdapter has a
CreateChannelFactory(GrpcApiDescriptor) method or similar, but I'm
nervous of adding too many abstractions.


Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -84,5 +104,29 @@ private static GrpcAdapter DetectDefaultPreferringGrpcNetClient()
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="grpcApiDescriptor"></param>
+        /// <returns></returns>
+        /// <exception cref="NotImplementedException"></exception>
+        public static GrpcAdapter GetFallbackAdapter(GrpcApiDescriptor grpcApiDescriptor)
+        {
+            // TODO: Some way of indicating a preference? Or just set the adapter in the client builder...?
```
amanda-tarafa, line 116: I think this is fine. For a specific adapter they have the builder property and DI. For the fallback is fine if we get to decide.

Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -84,5 +104,29 @@ private static GrpcAdapter DetectDefaultPreferringGrpcNetClient()
...
+            if (grpcApiDescriptor.Transports.HasFlag(GrpcTransports.Grpc))
+            {
+                // TODO: This is all a bit of a mess.
+                return CreateDefaultAdapter();
```
amanda-tarafa, line 120: Why not?
```suggestion
                return s_defaultFactory.Value;
```

amanda-tarafa: Or even 
```suggestion
                return DefaultAdapter;
```

Google.Api.Gax.Grpc/Rest/RestGrpcAdapter.cs
```diff
@@ -74,5 +75,14 @@ public sealed class RestGrpcAdapter : GrpcAdapter
...
             };
+
+        private class GrpcChannelFactory
+        {
+            private RestServiceCollection _serviceCollection;
+
+            internal GrpcChannelFactory(RestServiceCollection serviceCollection) =>
+                _serviceCollection = serviceCollection;
+
+        }
```
amanda-tarafa, line 86: Left over maybe?

Google.Api.Gax.Grpc/GrpcChannelFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+using Grpc.Core;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Interoperability layer for the aspects of gRPC that aren't covered by Grpc.Core.Api.
+    /// </summary>
+    internal abstract class GrpcChannelFactory
```
amanda-tarafa, line 15: Left over as well?

Google.Api.Gax.Grpc/GrpcApiDescriptor.cs
```diff
@@ -0,0 +1,45 @@
...
+using Google.Protobuf.Reflection;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Provides metadata about an API. This is expected to be constructed with a single instance
+    /// per API; equality is by simple identity.
```
amanda-tarafa, line 17: Maybe overriding equals and hashcode explicitly just calling base it's more visible than this comment here. Obviously keep this as well.

I was about to leave a comment about missing equals and hashcode :)

amanda-tarafa: Also, just looping the loop but, would it be possible to just want to call some operations with one adapter and some operations with another, and you want to make that explicit by only passing the specific FileDescriptors?

It would be great for workarounds where some operations are not working via gRPC once we support REGAPIC for all.

Granted that descriptors are not per operation so this wouldn't cover all cases.

Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -21,30 +22,49 @@ public abstract class GrpcAdapter
 
         private static readonly Lazy<GrpcAdapter> s_defaultFactory = new Lazy<GrpcAdapter>(CreateDefaultAdapter, LazyThreadSafetyMode.PublicationOnly);
 
+        private GrpcTransports _supportedTransports;
```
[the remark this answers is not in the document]
amanda-tarafa: The first point here ties in with my latest comment about wanting to do, for the same API, some operations with REST and some with gRPC.

I know this is a very cornerish case though.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -454,12 +454,11 @@ internal async Task<GoogleCredential> GetGoogleCredentialAsync(CancellationToken
...
         /// <summary>
-        /// Returns the default <see cref="GrpcAdapter"/> to use if
-        /// <see cref="GrpcAdapter"/> is not set.
+        /// 
         /// </summary>
-        protected abstract GrpcAdapter DefaultGrpcAdapter { get; }
+        protected abstract GrpcApiDescriptor GrpcApiDescriptor { get; }
 
-        private GrpcAdapter EffectiveGrpcAdapter => GrpcAdapter ?? DefaultGrpcAdapter;
+        private GrpcAdapter EffectiveGrpcAdapter => GrpcAdapter ?? GrpcAdapter.GetFallbackAdapter(GrpcApiDescriptor);
```
amanda-tarafa, line 461: This is not taking DI into acount right now, right? Which is fine as that hasn't been merged, just answered myself.

jskeet: Correct. Although the DI part would set `GrpcAdapter` anyway, so it might not make much difference.

Google.Api.Gax.Grpc/GrpcAdapter.cs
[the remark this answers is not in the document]
jskeet: The intent of this PR is to move in the opposite direction: allow a single GrpcAdapter instance to handle multiple APIs, even for the REST adapter. I'd prefer users to be able to say "I want to use GrpcNetClientAdapter" and be done. GrpcApiDescriptor is meant to be specific to an individual API. That consists of "extra information needed by the gRPC adapter" as well as "constraints on which gRPC adapters will work" - if we were to separate out those two aspects, we could accept the constraint part, but I'd rather not include something that looks like API-specific information.

In terms of private vs protected: I'm actually thinking this should probably be a public read-only property.

Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -21,30 +22,49 @@ public abstract class GrpcAdapter
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="supportedTransports"></param>
+        protected GrpcAdapter(GrpcTransports supportedTransports)
+        {
+            // TODO: Validation that all flags are valid, and that there's at least one flag?
```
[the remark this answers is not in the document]
jskeet: Not sure on either front, really - it feels odd to create a GrpcAdapter that can never actually be used for anything.

What I *could* potentially get behind is a "test" transport type, if we want to expose a TestGrpcAdapter... probably in the Testing package, although maybe even in this package if it means we can keep more things internal. Let's chat more about all of this when we're closer to merging. (I think there's a lot to do before that.)

Google.Api.Gax.Grpc/GrpcAdapter.cs
```diff
@@ -21,30 +22,49 @@ public abstract class GrpcAdapter
...
         /// <summary>
         /// Creates a channel for the given endpoint, using the given credentials and options.
         /// </summary>
+        /// <param name="apiDescriptor"></param>
         /// <param name="endpoint">The endpoint to connect to. Must not be null.</param>
         /// <param name="credentials">The channel credentials to use. Must not be null.</param>
         /// <param name="options">The channel options to use. Must not be null.</param>
         /// <returns>A channel for the specified settings.</returns>
-        public ChannelBase CreateChannel(string endpoint, ChannelCredentials credentials, GrpcChannelOptions options)
+        public ChannelBase CreateChannel(GrpcApiDescriptor apiDescriptor, string endpoint, ChannelCredentials credentials, GrpcChannelOptions options)
```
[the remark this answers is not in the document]
jskeet: I'd rather have an `IsSupported(GrpcApiDescriptor)` method - my expectation is that the `Configure` method for DI (which is the place we might have multiple configured adapters) can work out which one to use. (We might have a static convenience method in `GrpcAdapter` to do that: given an API descriptor and multiple adapters, return the first which supports the API, or null if none of them do.) At that point, we don't actually want to create the channel - we're just getting ready for a later `Build` call.

Google.Api.Gax.Grpc/GrpcAdapter.cs
jskeet: Yup. I suspect it's fine too. Let's leave this TODO here for now just in case.

Google.Api.Gax.Grpc/GrpcAdapter.cs
jskeet: Changed to `DefaultAdapter`, but I'm expecting all of this will change anyway, and we won't expose a single default adapter - this method will effectively replace it. But I think we can get to the details of that slightly later in the evolution of this PR :)

Google.Api.Gax.Grpc/GrpcApiDescriptor.cs
jskeet: Hmm... not sure about overriding just to return the base values. Let's talk about that more.

In terms of calling some operations with one and some with another: if you want to do that, create two different clients :) I don't think this is a degree of flexibility we should be trying to incorporate, really.

Google.Api.Gax.Grpc/GrpcChannelFactory.cs
jskeet: Whoops, yes. Was considering separating the two, but then didn't.

Google.Api.Gax.Grpc/Rest/RestGrpcAdapter.cs
jskeet: Yes - was expecting to need more, but given that it's *just* a service collection, I figured I could put that into the dictionary directly.

jskeet:
Closing in favour of #553.

PR #549, opened 2022-03-23
feat: Make UseJwtAccessWithScopes public

### LINKED ISSUE #546: Simplify UseJwtAccessWithScopes
We shouldn't bother generating complicated code in builders for UseJwtAccessWithScopes - we never "override" it anyway, as all APIs support it.

We can leave it as a protected property which defaults to being true (and stop generating the explicit parameterless constructors), and potentially make it public in the future. (We could just remove it entirely, but it feels useful to be able to potentially intervene.)

Filing this as a feature request in terms of simplifying the client code.

jskeet:
This is an escape hatch, effectively - e.g. for scenarios where
custom endpoints that don't accept self-signed JWTs are being used.
With this in place, we can remove the generator code that sets the
property in the constructor.

Fixes #546.


amanda-tarafa (approved):
LGTM

jskeet:
Closing and reopening to try to prod the build.

PR #550, opened 2022-03-29
Taking #548 further...

jskeet:
This starts with commits from #548, but then takes it further, with a `ServiceMetadata` class to contain "all the default aspects of a service".

I was hoping we'd be able to only do the REST method parsing on a per-service basis, but that doesn't quite work at the moment, due to the need for a single channel to handle calls to multiple services (e.g. for LRO and mixins).


Google.Api.Gax.Grpc/ServiceMetadata.cs
```diff
@@ -0,0 +1,77 @@
...
+using Google.Protobuf.Reflection;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Provides metadata about a single service within an API.
+    /// Often most of these aspects will be the same across multiple services,
+    /// but they can be specified with different values in the original proto, so
```
amanda-tarafa, line 17: nit: Just flagging that this comment is incomplete

Google.Api.Gax.Grpc/Rest/RestServiceCollection.cs
```diff
@@ -18,6 +18,7 @@ namespace Google.Api.Gax.Grpc.Rest
     /// Represents a set of methods all expected to be accessible via the same host.
     /// (The host itself is not part of the state of this class.)
     /// TODO: Do we need this class, or could we keep the dictionary directly in RestChannel?
+    /// TODO: Perhaps change this to be on a per-service basis? (Rename to RestMethodCollection...)
```
amanda-tarafa, line 21: I think if you do this then the channel cannot be shared across services in an API? Unless you change the implementation of the REST adapter.

Google.Api.Gax.Grpc/ServiceMetadata.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="apiMetadata"></param>
+        public ServiceMetadata(ServiceDescriptor serviceDescriptor, string defaultEndpoint, IEnumerable<string> defaultScopes, bool supportsScopedJwts, GrpcTransports transports, ApiMetadata apiMetadata)
+        {
+            ServiceDescriptor = GaxPreconditions.CheckNotNull(serviceDescriptor, nameof(serviceDescriptor));
+            Name = serviceDescriptor.Name;
+            DefaultEndpoint = defaultEndpoint;
+            DefaultScopes = GaxPreconditions.CheckNotNull(defaultScopes, nameof(defaultScopes)).ToList().AsReadOnly();
+            SupportsScopedJwts = supportsScopedJwts;
+            Transports = transports;
+            ApiMetadata = apiMetadata;
```
amanda-tarafa, line 74: This one cannot be null I think as it's the one we use to fetch the channel for rest.

Should some validations be in place to check that this service is really part of the API? Just because all of this is public.

Google.Api.Gax.Grpc/ServiceMetadata.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Linq;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Provides metadata about a single service within an API.
+    /// Often most of these aspects will be the same across multiple services,
+    /// but they can be specified with different values in the original proto, so 
+    /// </summary>
+    public sealed class ServiceMetadata
```
amanda-tarafa, line 19: I'm wondering what value does this add over ApiMetadata? (I know I mentioned per method granularity for specifying transport and this is the closest we can get)

Google.Api.Gax.Grpc/GrpcTransports.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    [Flags]
+    public enum GrpcTransports
```
amanda-tarafa, line 16: This is a nit, but should this be ApiTransports or just Transports?

Google.Api.Gax.Grpc/Gcp/GcpCallInvokerPool.cs
```diff
@@ -26,28 +26,23 @@ public sealed class GcpCallInvokerPool
 
         private readonly DefaultChannelCredentialsCache _credentialsCache;
 
+        private readonly ServiceMetadata _serviceMetadata;
         private readonly Dictionary<Key, GcpCallInvoker> _callInvokers = new Dictionary<Key, GcpCallInvoker>();
         private readonly object _lock = new object();
 
+        // TODO:
```
amanda-tarafa, line 33: nit: Empty TODO

Google.Api.Gax.Grpc/Gcp/GcpCallInvokerPool.cs
```diff
@@ -26,28 +26,23 @@ public sealed class GcpCallInvokerPool
...
         /// <summary>
-        /// Creates a call invoker pool which will apply the specified scopes to the default application credentials
-        /// if they require any.
+        /// Creates a call invoker pool which will use the given service metadata to determine scopes
+        /// and self-signed JWT support.
         /// </summary>
-        /// <param name="scopes">The scopes to apply. Must not be null, and must not contain null references. May be empty.</param>
-        public GcpCallInvokerPool(IEnumerable<string> scopes) : this(scopes, false)
+        /// <param name="serviceMetadata"></param>
+        public GcpCallInvokerPool(ServiceMetadata serviceMetadata)
```
amanda-tarafa, line 40: This means that invokers cannot be shared across services in the same API, even for gRPC? Or am I missing something? Or were these per service before?

I think it would be OK to allow the per service granularity, but we probably want to keep the per API pools?

amanda-tarafa: OK, I see that at least the normal ChannelPool is per-service right now.

amanda-tarafa (commented):
Some comments to chat about but looks good.

jskeet:
Closing in favour of #553.

PR #558, opened 2022-04-08
feat: Add logging for gRPC clients

jskeet:
We might well want to add some more logging into the streaming calls
in the future.

If we go ahead with this, we'll need to make corresponding generator
changes - they're likely to be relatively straightforward, but
affect a lot of generated test code.

At the moment there's no testing for logging actually being
performed. To do that, we really want an in-memory logger like the
one from the Functions Framework - I could copy that code,
potentially.

### DIFF
--- a/Google.Api.Gax.Grpc/ApiCall.cs
+++ b/Google.Api.Gax.Grpc/ApiCall.cs
@@ -7,6 +7,7 @@
 
 using Google.Protobuf;
 using Grpc.Core;
+using Microsoft.Extensions.Logging;
 using System;
 using System.Threading.Tasks;
 
@@ -151,11 +152,18 @@ namespace Google.Api.Gax.Grpc
         internal ApiCall<TRequest, TResponse> WithMergedBaseCallSettings(CallSettings callSettings) =>
             new ApiCall<TRequest, TResponse>(_asyncCall, _syncCall, callSettings.MergedWith(BaseCallSettings));
 
-        internal ApiCall<TRequest, TResponse> WithRetry(IClock clock, IScheduler scheduler) =>
+        // Note: it seems unfortunate that we can't reuse whatever logger is already configured here, but we don't
+        // have access to it. This is the logger for logging "I'm about to back off / retry" rather than the actual calls.
+        internal ApiCall<TRequest, TResponse> WithRetry(IClock clock, IScheduler scheduler, ILogger retryLogger) =>
             new ApiCall<TRequest, TResponse>(
-                _asyncCall.WithRetry(clock, scheduler),
-                _syncCall.WithRetry(clock, scheduler),
+                _asyncCall.WithRetry(clock, scheduler, retryLogger),
+                _syncCall.WithRetry(clock, scheduler, retryLogger),
                 BaseCallSettings);
+
+        internal ApiCall<TRequest, TResponse> WithLogging(ILogger logger) =>
+            logger is null
+                ? this
+                : new ApiCall<TRequest, TResponse>(_asyncCall.WithLogging(logger), _syncCall.WithLogging(logger), BaseCallSettings);
 
         /// <summary>
         /// Constructs a new <see cref="ApiCall{TRequest, TResponse}"/> that applies an overlay to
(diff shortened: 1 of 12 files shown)


jskeet:
I'm *reasonably* convinced we want this. We could potentially make it a lot less of a breaking change by keeping the ClientHelper constructor as it is, and adding a "WithLogging" call - but I'm not convinced that would be particularly beneficial.

amanda-tarafa:
I'll review first thing Monday.

Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -18,15 +19,21 @@ public class ClientHelper
...
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly string _baseCategoryName;
 
         /// <summary>
         /// Constructs a helper from the given settings.
         /// Behavior is undefined if settings are changed after construction.
         /// </summary>
         /// <param name="settings">The service settings.</param>
-        public ClientHelper(ServiceSettingsBase settings)
+        /// <param name="loggerFactory">The logger factory to use for API calls</param>
```
amanda-tarafa, line 30: I'm not sure about the logger per method. It's more usual to have a logger per class. Could we then do some structured logging to add method info etc.?

 I'm ultimately fine either way.

Google.Api.Gax.Grpc/ApiCallLoggingExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        // By design, the code is mostly duplicated between the async and sync calls.
+
+        #region Unary calls
+        // Async call
+        internal static Func<TRequest, CallSettings, Task<TResponse>> WithLogging<TRequest, TResponse>(
+            this Func<TRequest, CallSettings, Task<TResponse>> fn, ILogger logger) =>
+            async (request, callSettings) =>
+            {
+                logger?.LogTrace("Starting asynchronous API call.");
```
amanda-tarafa, line 25: Not sure if we should add an EventID here or if we should let loggers handle that?

Google.Api.Gax.Grpc/ApiCall.cs
```diff
@@ -151,12 +152,19 @@ public sealed class ApiCall<TRequest, TResponse>
         internal ApiCall<TRequest, TResponse> WithMergedBaseCallSettings(CallSettings callSettings) =>
             new ApiCall<TRequest, TResponse>(_asyncCall, _syncCall, callSettings.MergedWith(BaseCallSettings));
 
-        internal ApiCall<TRequest, TResponse> WithRetry(IClock clock, IScheduler scheduler) =>
+        // Note: it seems unfortunate that we can't reuse whatever logger is already configured here, but we don't
```
amanda-tarafa, line 155: But when we configure  them we are effectively using the same one, right?

jskeet: Yes, that's right.

Google.Api.Gax.Grpc/ApiCallLoggingExtensions.cs
jskeet: Loggers will add event IDs, I believe.

Google.Api.Gax.Grpc/ClientBuilderBase.cs
```diff
@@ -537,6 +545,7 @@ protected virtual void Configure(IServiceProvider provider)
         {
             GaxPreconditions.CheckNotNull(provider, nameof(provider));
 
+            LoggerFactory ??= provider.GetService<ILoggerFactory>();
```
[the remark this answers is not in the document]
jskeet: Hmm. Not sure - let's have a think about that separately. (We could make the Configure method public instead of protected. I can't immediately remember why it's protected anyway, as we'd need to generate code to call it... let's discuss.)

Google.Api.Gax.Grpc/ClientHelper.cs
jskeet: Hmm. The trouble is the code that's doing the logging doesn't know which method it's logging. We could potentially put that in ApiCall though.

If we accept an ILogger rather than an ILoggerFactory, that opens the option of clients using different loggers for different use-cases, too. We could either have just an ILogger property on ClientBuilderBase, or have both ILogger and ILoggerFactory and use the factory if we don't have a logger. Let's chat about it.

jskeet:
Closing in favour of #559.

PR #559, opened 2022-04-12
Introduce logging into GAX gRPC clients

jskeet:
Review one commit at a time.


jskeet:
This is now ready to review. The test loggers come from the Functions Framework, so should be fine as they are.

amanda-tarafa (approved):
LGTM. I like this.

PR #574, opened 2022-05-25
Update dependencies (and remove workaround reflection)

jskeet:
We should do a common protos 2.6.0 release before the GAX v4 release.


Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Google.Protobuf" Version="[3.18.0, 4.0.0)" />
+    <PackageReference Include="Google.Protobuf" Version="[3.20.0, 4.0.0)" />
```
amanda-tarafa: Should this one be 3.21.0 ?? On https://github.com/googleapis/gapic-generator-csharp/pull/487 you said we should move to 3.21.0 before GaxV4

generateprotos.sh
```diff
@@ -10,8 +10,8 @@ OS=windows
 declare -r ROOT=$(realpath $(dirname $0))
 cd $ROOT
 
-PROTOBUF_VERSION=3.18.0
-GRPC_VERSION=2.39.1
+PROTOBUF_VERSION=3.20.0
```
amanda-tarafa: And here as well.

amanda-tarafa (approved):
LGTM. With a question

Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
jskeet: Hmm... had thought I'd done that. (It all got a bit confusing due to timing.) Will have a look.

PR #577, opened 2022-06-24
First pass at gRPC transcoding improvements

jskeet:
This isn't yet absolutely complete, but:

- The refactoring work is about right now, I think
- It introduces the data-file-driven tests which I think we'll want to adopt across languages eventually

I'd like to get it in as the basis for other REGAPIC work.


jskeet:
Definitely best to review one commit at a time.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -9,198 +9,65 @@
...
-            {
-                throw new RpcException(status, httpResponse.GetTrailers());
-            }
-            return (TResponse) _parser.Parse(httpResponse.Content, _protoMethod.OutputType);
-        }
+    internal HttpRequestMessage CreateRequest(IMessage request, string host)
+    {
+        // TODO: Use multiple rules, iterating over them until a match is found.
+        var output = _transcoder.Transcode(request);
+        return output.CreateRequest(host);
```
amanda-tarafa: nit: Could these be squashed into one method call `_transcoder.CreateTranscodedRequest(request, how)`? Even if `TranscodingOutput` remains a class (maybe a child class of `HttpRuleTranscoder`).

Or is the separetion useful for testing or something else?

Google.Api.Gax.Grpc/Rest/TranscodingOutput.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Net.Http;
+using System.Text;
+
+namespace Google.Api.Gax.Grpc.Rest;
+
+/// <summary>
+/// The result of transcoding a protobuf request using an HttpRule.
+/// This is produced by <see cref="HttpRuleTranscoder"/>.
+/// </summary>
+internal sealed class TranscodingOutput
```
amanda-tarafa, line 20: nit: Maybe `TranscodedRequest` and then `CreateRequest` => `CreateHttpRequest`

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -9,158 +9,262 @@
...
 
-        internal string Format(IMessage message) => string.Join("", _segments.Select(segment => segment.Format(message)));
+    /// <summary>
+    /// Attempts to format the path with respect to the given request,
+    /// returning the formatted segment or null if formatting failed due to the request data.
+    /// </summary>
+    internal string TryFormat(IMessage message)
+    {
+        StringBuilder builder = new StringBuilder();
+        var segments = string.Join("", _segments.Select(segment => segment.TryFormat(message)));
```
amanda-tarafa, line 81: I don't think you are using this anymore.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -9,89 +9,140 @@
...
+
+            _pathPattern = HttpRulePathPattern.Parse(pattern, requestMessage);
+
+            (bool allFieldsInBody, string bodyName, _bodyTranscoder) = rule.Body switch
+            {
+                // TODO: The body shouldn't contain any fields that are present in the URI.
+                "*" => (true, null, new Func<IMessage, string>(protoRequest => protoRequest.ToString())),
+                "" => (false, null, new Func<IMessage, string>(protoRequest => null)),
+                string name when requestMessage.FindFieldByName(name) is FieldDescriptor field => (false, name, new Func<IMessage, string>(protoRequest => field.Accessor.GetValue(protoRequest).ToString())),
+                _ => throw new ArgumentException($"Method {methodName} has a body parameter {rule.Body} in the 'google.api.http' annotation which is not a field in {requestMessage.Name}")
```
amanda-tarafa, line 83: Just out of curiosity, we have a presubmit check for this?

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -9,89 +9,140 @@
...
+            };
+
+            _pathPattern = HttpRulePathPattern.Parse(pattern, requestMessage);
+
+            (bool allFieldsInBody, string bodyName, _bodyTranscoder) = rule.Body switch
+            {
+                // TODO: The body shouldn't contain any fields that are present in the URI.
+                "*" => (true, null, new Func<IMessage, string>(protoRequest => protoRequest.ToString())),
+                "" => (false, null, new Func<IMessage, string>(protoRequest => null)),
+                string name when requestMessage.FindFieldByName(name) is FieldDescriptor field => (false, name, new Func<IMessage, string>(protoRequest => field.Accessor.GetValue(protoRequest).ToString())),
```
amanda-tarafa, line 82: Can this ever be null?

`field.Accessor.GetValue(protoRequest)?.ToString() ?? ""`

amanda-tarafa: Or should that be a reason for failure? It's not clear on the AIP.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -9,89 +9,140 @@
...
+                HttpRule.PatternOneofCase.Patch => (rule.Patch, s_patchMethod),
+                HttpRule.PatternOneofCase.Put => (rule.Put, HttpMethod.Put),
+                HttpRule.PatternOneofCase.Delete => (rule.Delete, HttpMethod.Delete),
+                HttpRule.PatternOneofCase.Custom => (rule.Custom.Path, new HttpMethod(rule.Custom.Kind)),
+                _ => throw new ArgumentException("HTTP rule has no pattern")
+            };
+
+            _pathPattern = HttpRulePathPattern.Parse(pattern, requestMessage);
+
+            (bool allFieldsInBody, string bodyName, _bodyTranscoder) = rule.Body switch
```
amanda-tarafa, line 77: I think that Body being null is valid, seems like it actually cannot be specified for GET and DELETE.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -9,89 +9,140 @@
...
+            {
+                // TODO: The body shouldn't contain any fields that are present in the URI.
```
amanda-tarafa, line 79: What I understand is that the body must not be the same as a URI parameters vs. contain.

I also wonder if it's us that need to check for this. We don't check for many of the other constraints specified in the AIP (as far as I can see, but I might be missing something). For instance, we don't check for:

- The body clause must be identical in the top-level annotation and each additional binding.
- RPCs must not define a body at all for RPCs that use the GET or DELETE HTTP verbs.

If anything these should be presubmit checks on protos?

I also understand that, if we don't have those presubmit checks, and we end up executing requests with bad transcoding, errors that end users get will be confusing. In turn, if we check these constraints ourselves we can at least give meaningful errors. But I guess we either check all or none?

Again, I might be missing something here.

amanda-tarafa (approved):
LGTM with a few comments/questions possibly because I don't have all the context.

Google.Api.Gax.Grpc/Rest/TranscodingOutput.cs
jskeet: Maybe later - we tend to only really think about transcoding as the proto->HTTP side of things, so I'm not worried about being ambiguous there. CreateRequest => CreateHttpRequest (or ToHttpRequestMessage) makes sense.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: Thanks - will fix in an extra tidy-up commit.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
jskeet: It's useful for #579 :)

jskeet:
Added an extra commit to address review comments. PTAL.

amanda-tarafa (approved):
LGTM

PR #579, opened 2022-07-21
Support for requesting numeric enums

jskeet:
This won't actually have any effect until we use it in the generator.


Google.Api.Gax.Grpc/ApiMetadata.cs
```diff
@@ -54,22 +62,46 @@ private ApiMetadata(string name)
...
         }
 
+        /// <summary>
+        /// Method used in the above constructor to create the lazy provider.
+        /// (Unfortunately this can't be a local method.)
+        /// </summary>
+        private static Lazy<IReadOnlyList<FileDescriptor>> BuildDescriptorsProviderFromDescriptors(IEnumerable<FileDescriptor> descriptors)
+        {
+            var actualDescriptors = descriptors.ToList().AsReadOnly();
+            return new Lazy<IReadOnlyList<FileDescriptor>>(() => actualDescriptors);
```
amanda-tarafa, line 76: Shouldn't this be as follows for it to really be lazy? (Can't suggest here)

```csharp
return new Lazy<IReadOnlyList<FieldDescriptor>>(() => descriptors.ToList().AsReadOnly());
```

amanda-tarafa: And also, I think you can just put this line of code when calling the other constructor above and remove the static method.

Google.Api.Gax.Grpc/ApiMetadata.cs
```diff
@@ -54,22 +62,46 @@ private ApiMetadata(string name)
...
         /// <param name="descriptorsProvider">A provider function for the protobuf descriptors of the API. Must not be null, and must not
         /// return a null value. This will only be called once by this API descriptor, when first requested.</param>
-        public ApiMetadata(string name, Func<IEnumerable<FileDescriptor>> descriptorsProvider) : this(name)
+        public ApiMetadata(string name, Func<IEnumerable<FileDescriptor>> descriptorsProvider) : this(name, BuildDescriptorsProviderFromOtherProvider(descriptorsProvider), false)
+        {
+        }
+
+        /// <summary>
+        /// Method used in the above constructor to create the lazy provider.
+        /// (Unfortunately this can't be a local method.)
```
amanda-tarafa, line 91: Or just write the whole instruction when calling the constructor, a comment there "Create a lazy provider" should be good enough?

Google.Api.Gax.Grpc/ApiMetadata.cs
```diff
@@ -39,10 +40,17 @@ public sealed partial class ApiMetadata
...
         public string Name { get; }
 
-        private ApiMetadata(string name)
+        /// <summary>
+        /// When true, <see cref="RestGrpcAdapter"/> will request that enums are encoded as numbers in JSON
+        /// rather than as strings, preserving unknown values.
+        /// </summary>
+        public bool RequestNumericEnumJsonEncoding { get; }
+
+        private ApiMetadata(string name, Lazy<IReadOnlyList<FileDescriptor>> fileDescriptorsProvider, bool requestNumericEnumJsonEncoding)
```
amanda-tarafa, line 49: Or make this one recieve the `Func<IEnumerable<FileDescriptor>>` and build the lazy here.
You then only need to convert the `IEnumerable<FileDescriptor> descriptors` in `() => descriptors` before calling this constructor.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -32,25 +31,26 @@ internal class RestMethod
...
     /// <returns>A representation of the method that can be used to handle HTTP requests/responses.</returns>
-    internal static RestMethod Create(MethodDescriptor method, JsonParser parser)
+    internal static RestMethod Create(ApiMetadata apiMetadata, MethodDescriptor method, JsonParser parser)
     {
         var rule = method.GetOptions().GetExtension(AnnotationsExtensions.Http);
         if (rule is null)
         {
             throw new ArgumentException($"Method {method.Name} in service {method.Service.Name} has no HTTP rule");
         }
         var transcoder = new HttpRuleTranscoder(method.FullName, method.InputType, rule);
```
amanda-tarafa, line 52: Any reason why you just don't pass the ApiMetadata to the HttpRuleTranscoder and perform the check for int enums while transcoding. (And you can then hide TranscodingOutput :))
I think it'll be cleaner but I'm fine either way.

amanda-tarafa (approved):
LGTM with a few comments.

Google.Api.Gax.Grpc/ApiMetadata.cs
jskeet: The intention isn't for it be lazy in this case - indeed, we want to eagerly evaluate the argument so that any later changes aren't reflected. But we're wrapping it in a Lazy for consistency with the other overload.

Google.Api.Gax.Grpc/ApiMetadata.cs
jskeet: Not really sure what you mean - but we can't change the constructor parameter now, as this is already used in code.

Google.Api.Gax.Grpc/ApiMetadata.cs
jskeet: That would change the laziness for that case though.
See https://github.com/googleapis/gax-dotnet/blob/main/Google.Api.Gax.Grpc.Tests/ApiMetadataTest.cs#L16 for tests that would fail - which admittedly look like they're named the wrong way round...

amanda-tarafa: Ah yes, I though that the non-laziness there was unintended.

amanda-tarafa (approved):
LGTM

PR #581, opened 2022-07-22
fix: Improve query parameter transcoding

jskeet:
This still isn't entirely complete, but:

- Required fields are included even if they're set to the default value
- Repeated fields are included appropriately

I'll refactor this before handling nested messages appearing as query parameters.

This fixes the Compute integration tests.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
@@ -8,6 +8,7 @@
 using Google.Protobuf;
 using Google.Protobuf.Reflection;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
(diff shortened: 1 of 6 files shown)


Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -131,16 +132,45 @@ static IEnumerable<string> GetValues(FieldDescriptor field, IMessage request)
...
+                if (value is null)
                 {
                     yield break;
                 }
-                yield return value is IFormattable formattable
-                    ? formattable.ToString(format: null, CultureInfo.InvariantCulture)
-                    : value?.ToString();
 
-                // TODO: Handle repeated fields
+                if (!field.HasPresence && !IsRequired(field) && IsDefaultValue(value))
```
amanda-tarafa, line 140: nit: I think that I would understand this better like: 

```csharp
if (!IsRequired(field) && !field.HasPresence && IsDefaultValue(value))
```

And also, for when `field.HasPresence == true` shouldn't you later check that the field was actually set?

amanda-tarafa (approved):
LGTM, with one nit/question.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
amanda-tarafa: Thanks!

And, yes, didn't spot that, sorry!

PR #582, opened 2022-07-25
Finish transcoding (feature complete)

jskeet:


Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -81,30 +80,77 @@ internal SingleRuleTranscoder(string methodName, MessageDescriptor requestMessag
...
+            var result = new List<QueryParameterField>();
+
+            AccumulateMessages(message => message, currentPath: null, requestDescriptor);
+            return result;
+
+            void AccumulateMessages(Func<IMessage, IMessage> currentSelector, string currentPath, MessageDescriptor messageDescriptor)
+            {
+                if (descriptorStack.Contains(messageDescriptor))
+                {
+                    throw new ArgumentException($"Message descriptor {messageDescriptor.FullName} contains itself recursively.");
```
amanda-tarafa, line 115: Could it be that maybe the third level path in the recursion is in excluded paths? And you need to accept up to two levels of the recursion?

amanda-tarafa (approved):
LGTM, with one question, that might very well be a missunderstanding on my part.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
jskeet: Hmm... I suppose that's possible, but it would be relatively fiddly to find out - and I think it's very unlikely we'd actually run into that situation. (Whereas running into a situation which would otherwise bork the code with a stack overflow is entirely feasible, albeit accidentally. It's relatively hard to *accidentally* include several levels in the URL binding.)

PR #586, opened 2022-08-04
fix: More transcoding tests, and a couple of bug fixes for transcoding that they exposed

jskeet:


Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -78,9 +78,8 @@ internal SingleRuleTranscoder(string methodName, MessageDescriptor requestMessag
             {
                 "*" => (true, null, new Func<IMessage, string>(protoRequest => protoRequest.ToString())),
                 "" => (false, null, new Func<IMessage, string>(protoRequest => null)),
-                // TODO: If a field is specified, but then isn't present in the request, should the request
-                // fail or should it just have no body?
-                string name when requestMessage.FindFieldByName(name) is FieldDescriptor field => (false, field.JsonName, new Func<IMessage, string>(protoRequest => field.Accessor.GetValue(protoRequest).ToString())),
+                // TODO: If a field is specified, but it's not a message field, should we fail at this point?
```
amanda-tarafa, line 81: I seem to remember that the original TODO was more about whether the field was present but null.
Although this TODO is valid as well, I'm not sure if the body could be just a string, for instance.

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
jskeet: The original Todo is now handled. The new Todo is about validation of the config.

PR #588, opened 2022-08-30
fix: Allow slashes in single-star wildcard fields (gRPC transcoding)

jskeet:
I'm still trying to get more clarification about this, but it seems to me that both `*` and `**` should be allowed to match all characters; the difference is that "*" should only end up with a single URL segment after encoding, so it changes how we format the result.

As both `*` and `**` can be included in a single pattern, this commit also changes how escaping is controlled. Each group within the final regex is now escaped separately, either escaping slashes or not based on whether the wildcard was `*` or `**`.

Additionally, this change handles patterns with trailing literals (e.g. `abc/*/def`).

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
@@ -126,8 +126,8 @@ internal sealed class HttpRulePathPattern
         internal string JsonFieldPath { get; }
 
         private readonly Regex _validationRegex;
+        private readonly Func<Match, string> _formatter;
         private readonly Func<IMessage, string> _propertyAccessor;
-        private readonly bool _multiSegmentEscaping;
 
         /// <summary>
         /// Creates a segment representing the given field text, with respect to
@@ -144,8 +144,7 @@ internal sealed class HttpRulePathPattern
             string[] bits = fieldText.Split(s_fieldPathPatternSeparator, 2);
             string fieldPath = bits[0];
             string pattern = bits.Length == 2 ? bits[1] : "*";
-            _multiSegmentEscaping = pattern.Contains("/") || pattern.Contains("**");
-            _validationRegex = ParsePattern(pattern);
+            (_validationRegex, _formatter) = ParsePattern(pattern);
 
             string[] fieldNames = fieldPath.Split(s_fieldPathSeparator);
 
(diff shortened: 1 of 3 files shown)


Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -218,53 +217,78 @@ static FieldDescriptor GetField(MessageDescriptor descriptor, string name)
...
                     }
-                    builder.Append(Regex.Escape(pattern.Substring(currentIndex, starStart - currentIndex)));
+
                     int starEnd = starStart + 1;
                     while (starEnd < pattern.Length && pattern[starEnd] == '*')
                     {
                         starEnd++;
                     }
                     int starCount = starEnd - starStart;
+                    // Deliberately local to have a different variable per
```
amanda-tarafa, line 248: nit: per... iteration?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -218,53 +217,78 @@ static FieldDescriptor GetField(MessageDescriptor descriptor, string name)
...
+                        formatBuilder += (match, sb) => sb.Append(literal);
+                    }
                     if (starStart < 0)
                     {
                         break;
                     }
-                    builder.Append(Regex.Escape(pattern.Substring(currentIndex, starStart - currentIndex)));
+
                     int starEnd = starStart + 1;
                     while (starEnd < pattern.Length && pattern[starEnd] == '*')
```
amanda-tarafa, line 243: And could there be actually more than 2 starts? Why the while if not?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -218,53 +217,78 @@ static FieldDescriptor GetField(MessageDescriptor descriptor, string name)
...
                         case 1:
-                            builder.Append("[^/]+");
+                            formatBuilder += (match, sb) => sb.Append(Uri.EscapeDataString(match.Groups[matchGroupIndex].Value));
                             break;
                         case 2:
-                            builder.Append(".+");
+                            formatBuilder += (match, sb) => sb.Append(string.Join("/", match.Groups[matchGroupIndex].Value.Split('/').Select(segment => Uri.EscapeDataString(segment))));
                             break;
                         default:
                             throw new ArgumentException($"Resource pattern '{pattern}' is invalid");
```
amanda-tarafa, line 264: Ah, OK, I see this fails if there are more than 2 starts, but, would it be better if we failed sooner?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -218,53 +217,78 @@ static FieldDescriptor GetField(MessageDescriptor descriptor, string name)
...
                     {
                         starEnd++;
                     }
                     int starCount = starEnd - starStart;
+                    // Deliberately local to have a different variable per 
+                    int matchGroupIndex = groupIndex;
+                    groupIndex++;
+                    // We match any characters within the regex, regardless of number of stars. It just affects
+                    // how the value is escaped. This uses a reluctant matcher, which means x/*/** (which is more likely
+                    // than x/**/*) will match x/a/b/c as "*=a" and then "**=b/c" which is probably the best result.
```
amanda-tarafa, line 253: Could /**/* mean that there needs to be at least one segment at the end? Definetely something that should be in the spec. Or maybe even something like /**/*/** meaning there should be at least one segment?

amanda-tarafa (approved):
LGTM, with a few nits.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
amanda-tarafa: (I've updated the comment as the stars with the slashes were being used for bold formatting, so I've now put them inside code blocks)

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: Whoops, thanks. Done.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: Those would be valid patterns, yes - but `*/**` would accomplish the same thing and be more "usual". Fundamentally it's a bit ambiguous here.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
jskeet: Could do... this way is simpler if we need significantly different code based on one or two stars. It's not clear to me yet whether we do.

I'll merge the code as-is now, and we can easily refactor later.

PR #592, opened 2022-09-06
fix: Fix transcoding and release as 4.1.1

jskeet:
It's definitely worth reviewing one commit at a time.
Most commits are small in terms of production code, but the generated test code can be large - skip the generated code.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
@@ -165,7 +165,7 @@ internal sealed class HttpRulePathPattern
                 messageSelector = message =>
                 {
                     var parent = previousSelector(message);
-                    return parent is null ? null : (IMessage) field.Accessor.GetValue(previousSelector(parent));
+                    return parent is null ? null : (IMessage) field.Accessor.GetValue(parent);
                 };
                 descriptor = field.MessageType;
             }
--- a/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
@@ -170,27 +170,19 @@ internal sealed class HttpRulePathPattern
                 descriptor = field.MessageType;
             }
 
-            // Now handle the last field, which must be a singular integer or string field.
+            // Now handle the last field, which must be a singular scalar field.
+            // (It's extremely unusual for it to be anything other than a string field, but it's easy enough to handle.)
             var lastFieldName = fieldNames.Last();
             var lastField = GetField(descriptor, lastFieldName);
-            AppendFieldNameToJsonFieldPath(lastField);
-            _propertyAccessor = lastField.FieldType switch
+            if (lastField.IsRepeated || lastField.FieldType == FieldType.Message)
             {
-                FieldType.String =>
-                    message =>
-                    {
-                        var parent = messageSelector(message);
-                        return parent is null ? null : (string) lastField.Accessor.GetValue(parent);
-                    },
-                FieldType.Int32 or FieldType.SInt32 or FieldType.UInt32 or FieldType.Fixed32 or
-                FieldType.Int64 or FieldType.SInt64 or FieldType.UInt64 or FieldType.Fixed64 =>
-                    message =>
-                    {
-                        var parent = messageSelector(message);
-                        var value = parent is null ? null : ((IFormattable) lastField.Accessor.GetValue(parent));
-                        return value?.ToString(format: null, CultureInfo.InvariantCulture);
-                    },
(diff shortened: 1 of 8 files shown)


Google.Api.Gax.Grpc/Rest/RestServiceCollection.cs
```diff
@@ -37,7 +37,8 @@ internal static RestServiceCollection Create(ApiMetadata metadata)
             var typeRegistry = TypeRegistry.FromFiles(fileDescriptors.ToArray());
             var parser = new JsonParser(JsonParser.Settings.Default.WithIgnoreUnknownFields(true).WithTypeRegistry(typeRegistry));
             var methodsByName = services.SelectMany(service => service.Methods)
-                .Where(x => true) // TODO: filter out streaming methods.
+                // We don't yet support streaming methods.
+                .Where(x => !x.IsClientStreaming && !x.IsServerStreaming)
```
amanda-tarafa, line 41: Both of these will be true for bidi I think, else we need a specific check.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -268,12 +270,44 @@ internal IEnumerable<string> GetValues(IMessage request)
...
 
             static string FormatValue(object value) => 
                 value is bool b ? (b ? "true" : "false")
+                : value is Enum enumValue ? OriginalEnumValueHelper.GetOriginalName(enumValue)
                 : value is IFormattable formattable ? formattable.ToString(format: null, CultureInfo.InvariantCulture)
                 : value.ToString();
         }
+
+        // Effectively a cache of mapping from enum values to the original name as specified in the proto file,
+        // fetched by reflection. This code is taken from the protobuf JsonFormatter.
```
amanda-tarafa: I'm guessing the protobuf JsonFormatter is private then, or this code is at least?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -170,27 +170,19 @@ internal FieldSegment(string fieldText, MessageDescriptor descriptor)
                 descriptor = field.MessageType;
             }
 
-            // Now handle the last field, which must be a singular integer or string field.
+            // Now handle the last field, which must be a singular scalar field.
+            // (It's extremely unusual for it to be anything other than a string field, but it's easy enough to handle.)
             var lastFieldName = fieldNames.Last();
             var lastField = GetField(descriptor, lastFieldName);
+            if (lastField.IsRepeated || lastField.FieldType == FieldType.Message)
```
amanda-tarafa, line 177: Is `bytes` allowed as well?

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
```diff
@@ -268,46 +268,46 @@ internal IEnumerable<string> GetValues(IMessage request)
...
-        // Effectively a cache of mapping from enum values to the original name as specified in the proto file,
-        // fetched by reflection. This code is taken from the protobuf JsonFormatter.
-        private static class OriginalEnumValueHelper
-        {
-            private static readonly ConcurrentDictionary<System.Type, Dictionary<object, string>> _dictionaries
-                = new ConcurrentDictionary<System.Type, Dictionary<object, string>>();
+    internal static string FormatPathOrQueryValue(object value) =>
+        value is bool b ? (b ? "true" : "false")
+        : value is Enum enumValue ? OriginalEnumValueHelper.GetOriginalName(enumValue)
+        : value is IFormattable formattable ? formattable.ToString(format: null, CultureInfo.InvariantCulture)
```
amanda-tarafa, line 282: For the floating point numbers, is everyone else using invariant culture, or server side at least? Is that one of the things that's unclear?

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
jskeet: Everyone should definitely be using the invariant culture or equivalent (no "1,5" for example) but there are other problems around precision. Happy to talk about it on a VC if that would be useful.

Google.Api.Gax.Grpc/Rest/RestServiceCollection.cs
jskeet: Yup, I'm pretty sure they're both true for bidi.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscoder.cs
jskeet: This code is private, yes. It might be good to expose it at some point, but that can wait.

PR #597, opened 2022-09-06
feat: Expose a formatting method for routing header values

jskeet:


Google.Api.Gax.Grpc/ProtobufUtilities.cs
```diff
@@ -0,0 +1,91 @@
...
+        value is 0 || value is 0L ||
+        // uint32, fixed32, uint64, fixed64
+        value is UnsignedInt32Zero || value is 0UL ||
+        // bytes
+        (value is ByteString bs && bs.IsEmpty) ||
+        // Any enum (all protobuf enums have an underlying type of Int32)
+        (value is Enum enumValue && Convert.ToInt32(enumValue) == 0);
+
+    /// <summary>
+    /// Formats a value in a way that is suitable for a header, URL path segment (after URL-encoding), or query parameter. The value
```
amanda-tarafa, line 50: can't comment here but, where it says `after URL-encoding` did you mean `before URL-encoding`?

These are not encoded, right?

amanda-tarafa: Ah, I think you meant it **will** be valid after URL-encoding.

Google.Api.Gax.Grpc/ProtobufUtilities.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// is effectively formatted the same way it is in the JSON representation. The return value of this
+    /// method is only guaranteed for single primitive values - not repeated fields, maps, or messages.
```
amanda-tarafa, line 52: Maybe if this is made public at some point, it might be worth it to fail for repeated, maps, etc.

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Grpc/ProtobufUtilities.cs
jskeet: Yes, that's right :)

Google.Api.Gax.Grpc/ProtobufUtilities.cs
jskeet: Yup, potentially. I think we'd want to guarantee more stuff if we use it from anything that's expected to be used from hand-written code. I don't mind too much while it's just generated code that calls "things that call this".

PR #599, opened 2022-09-09
fix: Parse error details more robustly

### LINKED ISSUE #572: Unknown error types will cause *known* error types to be ignored in REGAPIC
We can't parse error details that contain unknown "Any" types in REGAPIC. We'll still return a reasonable RpcStatus of an error, but any other (known) error details would be lost.

If we parsed the Any as a Struct instead, we could ignore ones with unknown types.

jskeet:
This avoids details being lost if new messages are added to error_details.proto and then used in responses.

Fixes #572.


amanda-tarafa (approved):
LGTM

PR #614, opened 2022-09-15
chore: Regenerate protos

jskeet:
This is really just so that I don't have extraneous changes to ignore every time I regenerate the transcoding test protos.


amanda-tarafa (approved):
LGTM

PR #618, opened 2022-10-10
feat: Supports environment variable quota project ID.

amanda-tarafa:
- Env var has effect only for credentials used to create GAPIC and Veneer clients.


amanda-tarafa:
@jskeet FYI.

This is an alternative to googleapis/google-api-dotnet-client/pull/2228.

Not my prefer approach, but it is consistent. The env var would be used as a default value for quota project client options.

jskeet:
Nope, I definitely prefer the other approach.

amanda-tarafa:
Closing in favour of https://github.com/googleapis/google-api-dotnet-client/pull/2228

PR #627, opened 2022-11-04
feat: rest server streaming support

viacheslav-rostovtsev:
TODO: add docs, look more at the headers retrieving code.


jskeet (commented):
Initial very brief look at PartialDecodingStreamReader:

- I don't think this will cope with characters that are split between byte arrays. I think we should create a StreamReader and read text from that instead.
- I don't see how this handles } in string values
- We need to work out how to handle failure without just swallowing the exception. (I don't think that's the exception that will be thrown, either.)

I suspect we need a fuller JSON tokenizer for the middle part of this. It doesn't need to keep values etc, but needs to know what state it's in. We can probably start with https://github.com/protocolbuffers/protobuf/blob/main/csharp/src/Google.Protobuf/JsonTokenizer.cs for that - I'm happy to work on that side if you'd like. (We can agree an interface for it, then I can do the implementation while you modify the rest of the code around it.)

viacheslav-rostovtsev:
> I don't think this will cope with characters that are split between byte arrays. I think we should create a StreamReader and read text from that instead.
Good point. Switched to using StreamReader.

> I don't see how this handles } in string values
It tries to parse them and fails. Added comments to clarify.

>We need to work out how to handle failure without just swallowing the exception. (I don't think that's the exception that will be thrown, either.)
This failure is checking for complete object in buffer. I've refactored to better explain. (This was the exception that was being thrown. I agree that a more obvious one is better. Switched.)

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,146 @@
...
+                    continue;
+                }
+                
+                _currentBuffer.Append(c);
+                if (c != '}')
+                {
+                    continue;
+                }
+
+                try
```
jskeet, line 111: I guess this will work for now... it feels like quite an expensive way of detecting the end of a message, where a JSON tokenizer could do this without having to try to parse and catch the exception. (We could have quite a few of these in a single message due to nesting, and it's a shame if a string of "}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}}" causes tens of exceptions to be thrown...

jskeet: (I would also like to remove our Json.NET dependency eventually - we currently only have it transitively due to Google.Apis.Auth; in a new Auth library we'd want to remove it. But that can be handled later.)

viacheslav-rostovtsev:
I've tested it with Firestore (which uses their server-streaming queries for a lot of things) and all tests pass, except for BIDI ones. 
Those are not supposed to pass, but WILL require additional work (by their authors) since the NotImplemented error gets swallowed in an unobvious way.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
```
jskeet, line 1: Add copyright comment.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Google.Api.Gax.Grpc.Rest;
+
+/// <summary>
+/// An IAsyncStreamReader implementation that reads an array of messages
+/// from HTTP stream as they arrive in (partial) JSON chunks. 
+/// </summary>
+/// <typeparam name="TResponse">Type of proto messages in the stream</typeparam>
+public class PartialDecodingStreamReader<TResponse> : IAsyncStreamReader<TResponse>
```
jskeet, line 19: Does this need to be public? If we make it internal we'll have more flexibility later.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+{
+    private readonly Func<string, TResponse> _parsing;
```
jskeet, line 21: "parsing" feels like an odd name for a variable to me. Converter? Parser? ResponseConverter?

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        if (_readyResults.Count > 0)
+        {
+            Current = _readyResults.Dequeue();
+            return true;
+        }
+
+        while (_readyResults.Count == 0)
+        {
+            var buffer = new char[1000];
```
jskeet, line 55: We don't need to create a new buffer on every iteration. I'd move it above the while loop.

(I'd also change it to 8K characters.)

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+        
+        _readyResults = new Queue<TResponse>();
+        _currentBuffer = new StringBuilder();
+        _arrayClosed = false;
+    }
+
+    /// <inheritdoc />
+    public async Task<bool> MoveNext(CancellationToken cancellationToken)
+    {
+        if (_readyResults.Count > 0)
```
jskeet, line 47: I'd expect to see a "completed" check here - if we've already detected the end of the stream previously, we shouldn't keep trying to read from it.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: I'm okay with this staying in for now, but:

- Please call `ToString` before this try/catch, so we can use the same string below.
- Please add a TODO to put a JSON tokenizer in here instead

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                if (!_arrayClosed)
+                {
+                    var errorText = "Closing `]` bracket not received after iterating through the stream. " +
+                                    "This means that streaming ended without all objects transmitted. " +
+                                    "It is likely a result of server or network error.";
+                    throw new InvalidOperationException(errorText);
+                }
+
+                break;
```
jskeet, line 77: This "break" is effectively "return false" isn't it? I think it would be clearer to do that directly and remove the `if (!_readyResults.Any())` code later.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+            foreach (var c in readChars)
+            {
+                // Closing bracket for the top-level array
+                if (_currentBuffer.Length == 0 && c == ']')
+                {
+                    _arrayClosed = true;
+                    continue;
+                }
+
+                // Between-objects commas and spaces, as well as an opening bracket
```
jskeet, line 90: This makes me slightly nervous in terms of accepting garbage... but it'll do for now.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,136 @@
...
+                    JObject.Parse(_currentBuffer.ToString());
+                }
+                catch (Newtonsoft.Json.JsonReaderException)
+                {
+                    // Tried to parse a partial json because the `}` was a part of
+                    // a string or a child inner object.
+                    continue;
+                }
+
+                TResponse obj = _parsing(_currentBuffer.ToString());
```
jskeet, line 119: Can we rename this to `responseElement` or something similar? That's a bit more descriptive than `obj`.

Google.Api.Gax.Grpc/Rest/ReadHttpResponseMessage.cs
```diff
@@ -64,10 +65,15 @@ internal string Content
             (OriginalResponseMessage, _readException) = (response, readException);
 
         internal Metadata GetHeaders()
+        {
```
jskeet, line 68: Please use an expression-bodied method:

```csharp
internal Metadata GetHeaders() => ReadHeaders(OriginalResponseMessage.Headers);
```

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -146,7 +189,7 @@ private async Task AddAuthHeadersAsync(HttpRequestMessage request, RestMethod re
             // Finally, if the channelTask completes, the await does nothing.
             await resultTask.ConfigureAwait(false);
             // If we're here, the channelTask has completed successfully.
-           
+
```
jskeet, line 192: I suspect this change was unintentional.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -75,7 +75,7 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
                 }
                 cancellationTokenSource = new CancellationTokenSource(delayInterval);
             }
-            var httpResponseTask = SendAsync(restMethod, host, options, request, cancellationTokenSource.Token);
+            var httpResponseTask = ReadResponseAsync(SendAsync(restMethod, host, options, request, cancellationTokenSource.Token, HttpCompletionOption.ResponseContentRead));
```
jskeet, line 78: I think it would be clearer to introduce a new local variable for the sending task.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -84,32 +84,40 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
             return new AsyncUnaryCall<TResponse>(responseTask, responseHeadersTask, statusFunc, trailersFunc, cancellationTokenSource.Cancel);
         }
 
-        private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request, CancellationToken deadlineToken)
+
```
jskeet, line 87: Remove extra blank line?

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -124,6 +132,41 @@ private async Task<ReadHttpResponseMessage> SendAsync<TRequest>(RestMethod restM
             }
         }
 
+        private async Task<HttpResponseMessage> SendAsync<TRequest>(RestMethod restMethod, string host, CallOptions options, TRequest request,
```
jskeet, line 135: Could you move this to before AsyncServerStreamingCall to make the change easier to understand? We can then have a purely-refactoring PR that moves it.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -85,4 +85,16 @@ internal async Task<TResponse> ReadResponseAsync<TResponse>(Task<ReadHttpRespons
...
         return (TResponse) _parser.Parse(httpResponse.Content, _protoMethod.OutputType);
     }
+
+    // TODO: Handle cancellation?
+    /// <summary>
+    /// Converts the http response into a IAsyncStreamReader.
+    /// </summary>
+    internal IAsyncStreamReader<TResponse> ResponseStreamAsync<TResponse>(Task<HttpResponseMessage> httpResponseTask)
+    {
+        var httpResponse = httpResponseTask.ConfigureAwait(false).GetAwaiter().GetResult();
```
jskeet, line 95: This makes me nervous: it's going to block until the task has completed. Why isn't this an async method using await?

jskeet: (Or if we can't make this async, we should move the "open the stream" part into PartialDecodingStreamReader, so that *that* can await it appropriately.)

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/ReadHttpResponseMessage.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
viacheslav-rostovtsev: 👍

viacheslav-rostovtsev:
Addressed the comments, please take another look.

Google.Api.Gax.Grpc/Rest/RestMethod.cs
viacheslav-rostovtsev: Refactored to address this and push the first wait to the MoveAsync.
Also added some unit tests.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+            DataStr = JsonConvert.SerializeObject( new [] { foo, bar }, Formatting.Indented);
+        }
+
+        /// <summary>
+        /// Test coarse split data.
+        /// </summary>
+        [Fact]
+        public async void DecodingStreamReaderTestByLine()
+        {
+
```
jskeet, line 44: Nit: remove blank line.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace Google.Api.Gax.Grpc.Rest.Tests
+{
+    public class PartialDecodingStreamReaderTest
+    {
+        private static readonly string DataStr;
```
jskeet, line 22: Please rename to something more descriptive - and personally I'd suggest using a verbatim string literal to show exactly what you expect it to look like.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// Test coarse split data.
+        /// </summary>
+        [Fact]
+        public async void DecodingStreamReaderTestByLine()
+        {
+            
+            StreamReader reader = new MyStreamReader(DataStr.Split('\n'));
+            var decodingReader = new PartialDecodingStreamReader<JObject>(Task.FromResult(reader), JObject.Parse);
+
+            var result = await decodingReader.MoveNext(CancellationToken.None).ConfigureAwait(false);
```
jskeet, line 48: There's no need for `ConfigureAwait` in unit tests.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            result = await decodingReader.MoveNext(CancellationToken.None).ConfigureAwait(false);
+            Assert.False(result);
+        }
+    }
+
+    /// <summary>
+    /// A fake of a StreamReader emitting given strings
+    /// </summary>
+    internal class MyStreamReader : StreamReader
```
jskeet, line 132: Please rename this to something more descriptive - `ReplayingStreamReader` would probably be reasonable.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        public override Task<int> ReadAsync(char[] buffer, int index, int count)
+        {
+            if (_queue.Count <= 0)
+            {
+                base.ReadToEnd(); // EndOfStream starts to return true
+                return Task.FromResult(0);
+            }
+
+            var nextString = _queue.Dequeue();
+            for (int i = 0; i < nextString.Length; i++)
```
jskeet, line 157: This is ignoring both `index` and `count`.
If we're confident that `count` will always be big enough, let's assert that and then just use `nextString.Length`, but we should definitely observe `index`. I'm somewhat surprised this is working without that.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,150 @@
...
+    
+    private StreamReader _streamReader;
+    private bool _arrayClosed;
+
+    /// <summary>
+    /// Creates the StreamReader
+    /// </summary>
+    /// <param name="streamReaderTask">A stream reader returning partial JSON chunks</param>
+    /// <param name="responseConverter">A function to transform a well-formed JSON object into the proto message.</param>
+    public PartialDecodingStreamReader(Task<StreamReader> streamReaderTask, Func<string, TResponse> responseConverter)
```
jskeet, line 42: I think it would be better to use a `TextReader` than a `StreamReader` here. We'd use `StreamReader` for the prod implementation, but logically all we care about is being able to read text data.

That means we can't use `EndOfStream`, but that's okay - I think it makes sense to have a field that knows we've finished reading the data.

jskeet (commented):
Mostly fine now. I think my biggest suggested change would be to use TextReader instead of StreamReader.

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc.Tests/Rest/PartialDecodingStreamReaderTest.cs
```diff
@@ -0,0 +1,165 @@
...
+            Assert.False(result);
+        }
+
+        /// <summary>
+        /// Test when data breaks off unexpectedly.
+        /// </summary>
+        [Fact]
+        public async void DecodingStreamReaderTestIncomplete()
+        {
+            var s = DataStr.Split(',')[0]; // right after the first element ends
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -0,0 +1,150 @@
...
+                }
+
+                return false;
+            }
+
+            var readChars = buffer.Take(readLen);
+            foreach (var c in readChars)
+            {
+                // Closing bracket for the top-level array
+                if (_currentBuffer.Length == 0 && c == ']')
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -75,41 +75,85 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
+            CancellationToken deadlineToken, HttpCompletionOption httpCompletionOption)
         {
             // Ideally, add the header in the client builder instead of in the ServiceSettingsBase...
-            var httpRequest = restMethod.CreateRequest((IMessage) request, host);
+            var httpRequest = restMethod.CreateRequest((IMessage)request, host);
             foreach (var headerKeyValue in options.Headers
-                .Where(mh => !mh.IsBinary)
-                .Where(mh=> mh.Key != VersionHeaderBuilder.HeaderName))
+                         .Where(mh => !mh.IsBinary)
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -75,41 +75,85 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
+                         .Where(mh => mh.Key != VersionHeaderBuilder.HeaderName))
             {
                 httpRequest.Headers.Add(headerKeyValue.Key, headerKeyValue.Value);
             }
+
             httpRequest.Headers.Add(VersionHeaderBuilder.HeaderName, RestVersion);
 
             HttpResponseMessage httpResponseMessage;
-            using (CancellationTokenSource linkedCts = CancellationTokenSource.CreateLinkedTokenSource(options.CancellationToken, deadlineToken))
+            using (CancellationTokenSource linkedCts =
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -75,41 +75,85 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
+
+            return httpResponseMessage;
+        }
+
+        /// <summary>
+        /// Equivalent to <see cref="CallInvoker.AsyncServerStreamingCall{TRequest, TResponse}(Method{TRequest, TResponse}, string, CallOptions, TRequest)"/>.
+        /// </summary>
+        public AsyncServerStreamingCall<TResponse> AsyncServerStreamingCall<TResponse, TRequest>(Method<TRequest, TResponse> method, string host, CallOptions options, TRequest request)
+        {
+            var restMethod = _serviceCollection.GetRestMethod(method);
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -85,4 +86,20 @@ internal async Task<TResponse> ReadResponseAsync<TResponse>(Task<ReadHttpRespons
         }
         return (TResponse) _parser.Parse(httpResponse.Content, _protoMethod.OutputType);
     }
+
+    internal IAsyncStreamReader<TResponse> ResponseStreamAsync<TResponse>(Task<HttpResponseMessage> httpResponseTask)
+    {
+        var streamReaderTask = GetStreamReader(httpResponseTask);
+        Func<string, TResponse> responseConverter = s =>  (TResponse) _parser.Parse(s, _protoMethod.OutputType);
```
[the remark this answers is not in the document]
viacheslav-rostovtsev: 👍

jskeet (approved):
This looks good, thanks - I do prefer #627 though :)

jskeet:
Merging now, and then will investigate the TextReader change which has some test failures at the moment.

PR #632, opened 2022-11-10
feat: Add scoped self-signed JWT support for REST

jskeet:


Google.Api.Gax.Rest/ClientBuilderBase.cs
```diff
@@ -99,6 +99,15 @@ public abstract class ClientBuilderBase<TClient>
         /// </remarks>
         public IHttpClientFactory HttpClientFactory { get; set; }
 
+        /// <summary>
+        /// Returns whether or not self-signed JWTs will be used over OAuth tokens when OAuth scopes are explicitly set.
```
amanda-tarafa, line 103: Do we have a note about this only applying to SAs in GAPICs? If no, then that's fine.

Google.Api.Gax.Rest.Tests/ClientBuilderBaseTest.GetHttpClientInitializerTest.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                var path = Path.GetTempFileName();
+                File.WriteAllText(path, TestServiceAccountJson);
+
+                var builder = new FakeBuilder(s_scopes, true)
+                {
+                    CredentialsPath = path
+                };
+                await AssertCredentialAsync(builder, AssertBasicProperties);
+                File.Delete(path);
```
amanda-tarafa, line 52: Do you have to delete it if it's a temp file?
I never do that, as I think it deletes on its own when it reaches a threshold.
Fine by me if this is a precaution though.

amanda-tarafa (approved):
LGTM

Google.Api.Gax.Rest/ClientBuilderBase.cs
jskeet: This comment is directly copy/pasted from the gRPC one :)

Google.Api.Gax.Rest.Tests/ClientBuilderBaseTest.GetHttpClientInitializerTest.cs
jskeet: We probably don't *have* to, but we might as well. This whole test is copy/pasted from gRPC and changed just enough to make it pass (with the slightly different API we expose).

PR #634, opened 2022-11-14
chore: Add a simple JSON state tracker

jskeet:
This will allow a small change after #627 has been merged, to allow us to detect the end of a single response, and the end of all responses, more efficiently and easily.

(The aim is for PartialDecodingStreamReader to become simpler, so that it really doesn't need to know about JSON at all.)

The tests look brief, but have 100% coverage other than the "we don't know what this state means" situation (which can't be tested without poking via reflection, of course).

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
+++ b/Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
@@ -1,0 +1,240 @@
+﻿/*
+ * Copyright 2022 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+
+namespace Google.Api.Gax.Grpc.Rest;
+
+/// <summary>
+/// Simple state tracker to indicate when a server-streamed response is "done".
+/// </summary>
+/// <remarks>
+/// We expect that:
+/// - The overall response is an array
+/// - The later JSON parser will perform fuller validation (e.g. this code
+///   won't spot that [{[{]}}] is broken.
+///
+/// Note that this is mutable rather than us creating a new instance on each "push"
+/// of a character, because:
+/// - It's not *just* a simple state machine: we count open/close object/arrays
+/// - We could make it a struct containing those counters and a detailed state enum,
+///   it's not obvious that would be easier to use.
+/// </remarks>
+internal sealed class JsonStateTracker
+{
+    private int arrayDepth;
+    private int objectDepth;
+
+    private State state = State.BeforeTopArray;
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
```diff
@@ -0,0 +1,240 @@
...
+                return NextAction.IgnoreAndContinue;
+            case '{':
+                objectDepth++;
+                state = State.WithinObject;
+                return NextAction.BufferAndContinue;
+            case ']':
+                arrayDepth--;
+                state = State.AfterTopArray;
+                return NextAction.SignalEndOfResponses;
+            default:
```
amanda-tarafa, line 91: An array of literals is never expected?

Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
```diff
@@ -0,0 +1,240 @@
...
+    /// <summary>
+    /// The action that should be taken by the caller immediately after a call to <see cref="Push(char)"/>.
+    /// </summary>
+    internal enum NextAction
+    {
+        /// <summary>
+        /// We've received the closing ] at the end of the top level array.
+        /// There should be no further non-whitespace characters.
+        /// </summary>
+        SignalEndOfResponses,
```
amanda-tarafa, line 213: There's one little nit here re usage. If you signal end of response, then the caller is not "encouraged" to continue pushing characters, so that last piece of validation may not happen. But it's us using it, so maybe not a big deal.

The alternative is for this tokenizer to return here IgnoreAndContinue. We don't need to signal end of response as that's when the stream finishes, I think.

amanda-tarafa (approved):
LGTM, with a couple of comments.

Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
jskeet: I'm okay either way, to be honest. Let's see what @viacheslav-rostovtsev thinks.

Google.Api.Gax.Grpc/Rest/JsonStateTracker.cs
jskeet: As the top level array? No, we always expect to get a response *message*.

I guess *strictly* speaking the message could be a FieldMask, Timestamp or Duration which has a direct string representation in JSON instead of a JSON object, but I don't think we need to handle that case. (In reality, all RPCs have Request/Response objects.)

jskeet:
@viacheslav-rostovtsev: PTAL as well.

jskeet:
Spoken with @viacheslav-rostovtsev offline, and he's happy for this to be merged. Will do so, then integrate.

PR #637, opened 2022-11-15
fix: Drop trailing hex characters from version header values

### LINKED ISSUE #636: Investigate hex after version numbers in x-goog-api-client headers
Our `VersionHeaderBuilder.AppendAssemblyVersion` includes hex, at least in some cases. 

So for example, in a test, we get `foo/1.0.0+01ad4caf515a21bb5af11e227e1c9b172cf26bbd` ... when I'd expect `foo/1.0.0`.

This should be fairly easy to fix, but we'll need to check that it's not going to cause any problems internally.

jskeet:
Fixes #636

### DIFF
--- a/Google.Api.Gax/VersionHeaderBuilder.cs
+++ b/Google.Api.Gax/VersionHeaderBuilder.cs
@@ -102,7 +102,7 @@ namespace Google.Api.Gax
             var info = assembly.GetCustomAttributes<AssemblyInformationalVersionAttribute>().FirstOrDefault()?.InformationalVersion;
             if (info != null && IsHeaderNameValueValid(info)) // Skip informational version if it's not a valid header value
             {
-                return info;
+                return FormatInformationalVersion(info);
             }
             var file = assembly.GetCustomAttributes<AssemblyFileVersionAttribute>().FirstOrDefault()?.Version;
             if (file != null)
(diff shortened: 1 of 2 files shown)


jskeet:
Marked as "do not merge" while we wait for confirmation that this won't break anything.

Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -112,6 +112,35 @@ private static string FormatAssemblyVersion(System.Type type)
...
 
+        // Visible for testing
+
+        /// <summary>
+        /// Formats an AssemblyInformationalVersionAttribute value to avoid losing useful information,
+        /// but also avoid including unnecessary hex that is appended automatically.
+        /// </summary>
+        internal static string FormatInformationalVersion(string info)
+        {
+            // In some case, the runtime includes 40 hex digits after a + or period in InformationalVersion.
```
amanda-tarafa, line 123: tiny nit:
```suggestion
            // In some cases, the runtime includes 40 hex digits after a + or period in InformationalVersion.
```

Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -112,6 +112,35 @@ private static string FormatAssemblyVersion(System.Type type)
...
+            {
+                return info;
+            }
+            for (int i = signIndex + 1; i < info.Length; i++)
+            {
+                char c = info[i];
+                bool isHex = (c >= '0' && c <= '9')
+                    || (c >= 'a' && c <= 'f')
+                    || (c >= 'A' && c <= 'F');
+                if (!isHex)
```
amanda-tarafa, line 136: Well, it could also all be hex characters and "valid" info. I guess what am saying is that the fact that `preview` and `build` use characters outside of hex is just a coincidence, there might be other similar qualifiers that use all hex characters.

If this is used by us alone, we know which qualifiers we use, and we can skip stripping those. Else we may be stripping things that someone intended to have as part of the version?

I know this is a nit-pick though, so it's fine as is.

amanda-tarafa (approved):
LGTM

Google.Api.Gax/VersionHeaderBuilder.cs
jskeet: It would have to be exactly 40 characters, all hex digits, following a period or `+`. I've been deliberately as picky as I can be, basically :)

amanda-tarafa: Oh, didn't notice the length checking before the for. Yep that probably makes it way better. Thanks!

PR #638, opened 2022-11-16
draft: server streaming with TextReader

viacheslav-rostovtsev:
https://github.com/googleapis/gax-dotnet/pull/627 but with an additional commit changing StreamReader to TextReader.


jskeet (approved):
This looks cleaner to me. Do you have any concerns?

jskeet:
(Except that the test now fails... I haven't checked out why...)

PR #643, opened 2022-11-22
chore: switch to textreader

viacheslav-rostovtsev:


jskeet:
@amanda-tarafa Given that both Slava and I have commits in this PR, please could you review it?

There are probably neater refactors of this, but as we're planning on revamping it anyway, I'll do that in a different PR with the state reader.

amanda-tarafa (approved):
Yep, LGTM

PR #644, opened 2022-11-23
chore: Integrate JSON state tracker into PartialDecodingStreamReader

jskeet:
This effectively reimplements PartialDecodingStreamReader - but without changing its API.

Note that the tests are mostly the same as before, but with a file-scoped namespace for clarity, and some new tests.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
+++ b/Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
@@ -6,16 +6,25 @@
  */
 
 using Grpc.Core;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Google.Api.Gax.Grpc.Rest;
+
+using static Google.Api.Gax.Grpc.Rest.JsonStateTracker.NextAction;
+
+// TODO: Disposal of the TextReader
+// TODO: Evaluate what exception should be thrown, in different cases:
+//       - Errors encountered that haven't naturally caused an exception
+//       - Regular exceptions (e.g. IOException, InvalidProtobufException)
+//       - Cancellation
+//       We should retain an ExceptionDispatchInfo so we can throw the
+//       right exception consistently (on follow-up calls) rather than
+//       just a message.
 
 /// <summary>
 /// An IAsyncStreamReader implementation that reads an array of messages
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -43,117 +101,147 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
...
-                    // it's possible to receive more data after the closing `]`
-                    // We currently just close the reader, which means we won't detect a response which
-                    // is bad due to trailing non-whitespace after the ]... it's unclear whether it's
-                    // worth continuing to read just to catch that though.
-                    _arrayClosed = true;
-                    _textReader.Close();
-                    // We can't just return false here, as we might have queued responses.
-                    break;
-                }
+                // Cancellation
```
amanda-tarafa, line 163: Or the task result is a null TextReader, in which case, you'll be here forever, or not, you'll be here until actual cancellation where the problem really was that the TextReader was null.

This will possibly never happen depending on how we construct this.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -43,117 +101,147 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
...
+
+    /// <summary>
+    /// Called when we don't have a queued response and haven't reached a terminal state (error or completed).
+    /// </summary>
+    /// <returns></returns>
+    private async Task ReadAndProcessData(CancellationToken cancellationToken)
+    {
+        try
+        {
+            _textReader ??= await AwaitWithCancellation(_textReaderTask, cancellationToken, defaultValue: null).ConfigureAwait(false);
```
amanda-tarafa, line 160: I think that if you get rid of `_textReader` and always call `AwaitWithCancellation` here you can move the cancellation check (and the dispose call) from `MoveNext` to `AwaitWithCancellation`. You can throw the cancelled exception in AwaitWithCancellation.

The result is slightly different in that we won't throw cancelled until we have to read again, but that seems OK, we usually don't check for cancellation in between syncrhonous operations.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -43,117 +101,147 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
...
         }
-
-        if (_readyResults.Count > 0)
-        {
-            Current = _readyResults.Dequeue();
-            return true;
-        }
-        else
+        // If *anything* fails, this signals an overall failure.
+        // Let the calling code effectively rethrow. (We lose the stack trace, which isn't ideal, admittedly.)
```
amanda-tarafa, line 222: I think with my comments above, we can just dispose here, and rethrow without loosing the stack trace.
Basically, if we dispose on error, we don't need to save the _errorMessage.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -6,32 +6,90 @@
...
+    private readonly Queue<TResponse> _queuedResponses;
+
+    /// <summary>
+    /// The current response that we're building up.
+    /// </summary>
+    private readonly StringBuilder _currentResponseBuffer;
     
+    /// <summary>
+    /// The reader containing the stream of responses.
+    /// This will initially be null; the first
```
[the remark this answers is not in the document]
jskeet: Fixed, thanks.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -6,32 +6,90 @@
...
+    /// recognize "we've reached the end of one response," or "we've reached
+    /// the end of all responses," or "something's gone wrong". (It doesn't
+    /// try to perform complete validation, but spots unexpected data between
+    /// elements etc.)
+    /// </summary>
+    private JsonStateTracker _jsonState;
+
+    /// <summary>
+    /// Will be set to true when all responses have been read from the stream.
+    /// (Any responses queue in <see cref="_queuedResponses"/> should still be returned.)
```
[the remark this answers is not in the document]
jskeet: Done.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: I don't think I'd want AwaitWithCancellation to dispose, but using a try/finally for that seems okay - and I do like the idea of not having to explicitly call ThrowIfCanceled.

One issue there is in terms of the exception we want to be thrown on cancellation. Currently we catch the exception, retain the message, and always throw InvalidOperationException. That may not be ideal, and needs more discussion, but probably in a later iteration. (I'll add a TODO.)

jskeet: Ooh, thinking about it, I'm not sure I like still returning more results (due to them being buffered) after the cancellation token has been cancelled. Will see where I get to.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: Right - changing to throw in AwaitWithCancellation will somewhat fix this; we'll get an NRE if we ever create a task that returns null, but that can't actually happen.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -43,117 +101,147 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
...
+                if (!_completed)
                 {
-                    // This will throw unless the characters in the _currentBuffer
-                    // add up to a correct JSON and since the _currentBuffer always
-                    // starts with an opening `{` bracket from one of the
-                    // top-level array's element's,
-                    // this will throw unless _currentBuffer contains one message.
-                    // TODO[virost, jskeet, 11/2022] Use a JSON tokenizer instead
-                    JObject.Parse(currentStr);
+                    _errorMessage = "Response stream completed without a closing array";
```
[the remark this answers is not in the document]
jskeet: I'm actually removing disposal entirely for now. I think we need to consider it more carefully in general.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: We need to save *something*, because if the caller calls MoveNext again, they should get the exception again - potentially *unless* the previous failure was due to cancellation, and they use a different cancellation token the next time. (That's a really tricky case, as we may have lost data. Ouch.)

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -83,6 +83,7 @@ internal class PartialDecodingStreamReader<TResponse> : IAsyncStreamReader<TResp
     /// </summary>
     private bool _endOfData;
 
+    // TODO: retain an exception instead of a message?
```
amanda-tarafa, line 86: Probably an ExceptionDispatchInfo would be better. Fine to do that in the next iteration.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
amanda-tarafa: For completeness we discussed this offline and we agreed that it's OK that the enumerator does not recover after a failed MoveNext, even if because cancellation.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -220,28 +203,26 @@ private async Task ReadAndProcessData(CancellationToken cancellationToken)
...
         var cancellationTask = Task.Delay(-1, cancellationToken);
         var resultTask = await Task.WhenAny(task, cancellationTask).ConfigureAwait(false);
-        return resultTask == cancellationTask ? defaultValue : task.Result;
+        if (resultTask == cancellationTask)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            // This is just a backstop. We really shouldn't get here.
+            throw new InvalidOperationException("Cancelled token did not throw");
+        }
+        return task.Result;
```
amanda-tarafa, line 226:
```suggestion
        cancellationToken.ThrowIfCancellationRequested();
        return task.Result;
```

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -92,9 +93,11 @@ internal class PartialDecodingStreamReader<TResponse> : IAsyncStreamReader<TResp
...
     /// <summary>
-    /// Creates the StreamReader
+    /// Creates a new instance which will read data from the TextReader provided by
+    /// <paramref name="textReaderTask"/>, and convert each object within a top-level JSON array
+    /// into a response using <paramref name="responseConverter"/>.
     /// </summary>
-    /// <param name="textReaderTask">A stream reader returning partial JSON chunks</param>
+    /// <param name="textReaderTask">A task to provide text reader returning partial JSON chunks</param>
     /// <param name="responseConverter">A function to transform a well-formed JSON object into the proto message.</param>
     public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string, TResponse> responseConverter)
```
amanda-tarafa, line 102: Instead of receiving the TextReader here, could we receive something else from which to build the TextReader from, in case that makes disposing easier?

amanda-tarafa (approved):
LGTM. Fine if all comments are addressed (or not) on later PRs.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: That code is certainly simpler, yes. I'm probably being paranoid about the cancelled task being done without the cancellation token being cancelled. Will do this, but with an additional comment.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: I don't think so - I think this is the right level of abstraction here, and disposing should be okay.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: Removed this comment entirely, but added more detail to the earlier TODO.

PR #645, opened 2022-11-24
chore: Fix disposal in server streaming for REGAPIC

jskeet:


Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -115,32 +117,44 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
     /// <inheritdoc />
     public async Task<bool> MoveNext(CancellationToken cancellationToken)
     {
-        while (true)
+        bool disposeOnExit = true;
+        try
```
amanda-tarafa, line 121: I think that if you dispose in the catch and rethrow, you don't need this variable. _endOfData would have already disposed on ReadAndProcessData.

Or, stop disposing on ReadAndProcessData and only dispose on MoveNext...

Or, if you check for cancellation on ReadAndProcessData only (I know you prefer checking always on MoveNext) then you can dispose and throw on ReadAndProcessData only.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -222,7 +237,40 @@ private static async Task<T> AwaitWithCancellation<T>(Task<T> task, Cancellation
...
+    public void Dispose()
+    {
+        if (_disposed)
+        {
+            return;
+        }
+        _disposed = true;
+
+        // Note: the returned task is deliberately ignored. It's fire-and-forget.
+        _textReaderTask.ContinueWith(DisposeReader, TaskContinuationOptions.OnlyOnRanToCompletion | TaskContinuationOptions.ExecuteSynchronously);
```
amanda-tarafa, line 260: Shouldn't this be awaited on DisposeAsync instead, and then have Dispose call DisposeAsync fire and forgetting?
DisposeAsync callers should be able to await for disposal if they wish to do so?

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -115,32 +117,44 @@ public PartialDecodingStreamReader(Task<TextReader> textReaderTask, Func<string,
...
+                // satisfies one of the above conditions. (We may need to loop multiple times.)
+                await ReadAndProcessData(cancellationToken).ConfigureAwait(false);
+            }
+        }
+        finally
+        {
+            if (disposeOnExit)
             {
-                return false;
+                Dispose();
```
amanda-tarafa, line 156: maybe 
```suggestion
               await DisposeAsync();
```

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -160,6 +174,7 @@ private async Task ReadAndProcessData(CancellationToken cancellationToken)
             if (charsRead == 0)
             {
                 _endOfData = true;
+                Dispose();
```
amanda-tarafa, line 177:
```suggestion
                await DisposeAsync();
```

amanda-tarafa (approved):
LGTM, with a few comments. But fine as is as well.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: I'm trying to avoid catching and rethrowing, basically - that always feels a bit weird.
Also, in the new PR I expect to separate out disposal part from everything else (at the same time as changing the exceptions being thrown). I think it'll be simpler at that point, when it's obvious that the "dispose on the way out" doesn't affect the result itself at all.

amanda-tarafa (approved):
LGTM

PR #646, opened 2022-11-24
feat: Improve server streaming exception behavior

jskeet:
Invalid JSON is now translated into an RpcException with a status code of DataLoss; any other exceptions are propagated simply. If any call to MoveNext fails with an exception, all future calls fail with the same exception.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
+++ b/Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
@@ -9,6 +9,7 @@ using Grpc.Core;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
```diff
@@ -158,72 +137,82 @@ public async Task<bool> MoveNext(CancellationToken cancellationToken)
...
-            var textReader = await AwaitWithCancellation(_textReaderTask, cancellationToken).ConfigureAwait(false);
+        var textReader = await AwaitWithCancellation(_textReaderTask, cancellationToken).ConfigureAwait(false);
 
-            var readTask = textReader.ReadAsync(_readBuffer, 0, _readBuffer.Length);
-            int charsRead = await AwaitWithCancellation(readTask, cancellationToken).ConfigureAwait(false);
+        var readTask = textReader.ReadAsync(_readBuffer, 0, _readBuffer.Length);
+        int charsRead = await AwaitWithCancellation(readTask, cancellationToken).ConfigureAwait(false);
 
-            if (charsRead == 0)
+        if (charsRead == 0)
```
amanda-tarafa, line 174: Just noticing this now, but I think this can be `charsRead < _readBuffer.Length`. We'd have to process the buffer before checking for this then though.
But it does save one extra read attempt.

amanda-tarafa (approved):
LGTM, and yes, this is way cleaner!
Just one more comment.

Google.Api.Gax.Grpc/Rest/PartialDecodingStreamReader.cs
jskeet: Nope - it's entirely valid for a reader to return fewer characters than the max, and still have more later. (Indeed, our tests would do this.)

PR #647, opened 2022-11-25
feat: Improve handling of bad/missing/unsupported methods in REGAPIC

jskeet:


Google.Api.Gax.Grpc.Tests/Rest/RestMethodTest.cs
```diff
@@ -35,9 +35,7 @@ public void CreateRequest_WithRequestNumericEnumJson(bool value, string expected
     public void CreateRequest_WithHttpOverrides()
     {
         var rule = new HttpRule { Get = "/v2/def/{name}" };
-        var methodDescriptor = TestServiceReflection.Descriptor.Services
-            .Single(svc => svc.Name == "Sample")
-            .FindMethodByName("SimpleMethod");
+        var methodDescriptor = GetMethod("Sample", "SimpleMethod");
```
amanda-tarafa, line 38: Is it worth it to check that the override overrides even when there were no rules? Just so that we don't change it back to what it was?

amanda-tarafa (approved):
LGTM, this is nice!

Google.Api.Gax.Grpc.Tests/Rest/RestMethodTest.cs
jskeet: In terms of adding a test? Yes, I can do that.

amanda-tarafa (approved):
LGTM!

amanda-tarafa (approved):

PR #653, opened 2022-11-29
fix: Fix cancellation (including deadlines) for REGAPIC streaming

### LINKED ISSUE #648: Indicate cancellation in REGAPIC with an RpcException
In both Grpc.Core and Grpc.Net.Client, for both streaming calls and unary calls, if a cancellation token is cancelled the result is an RpcException with a StatusCode of Cancelled.

We should do the same in REGAPIC.

jskeet:
Fixes #648

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestChannel.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestChannel.cs
@@ -64,30 +64,36 @@ namespace Google.Api.Gax.Grpc.Rest
         {
             var restMethod = _serviceCollection.GetRestMethod(method);
 
-            var cancellationTokenSource = new CancellationTokenSource();
-            if (options.Deadline.HasValue)
-            {
-                // TODO: [virost, 2021-12] Use IClock.
-                var delayInterval = options.Deadline.Value - DateTime.UtcNow;
-                if (delayInterval.TotalMilliseconds <= 0)
-                {
-                    throw new RpcException(new Status(StatusCode.DeadlineExceeded, $"The timeout was reached when calling a method `{restMethod.FullName}`"));
-                }
-                cancellationTokenSource = new CancellationTokenSource(delayInterval);
-            }
+            var deadlineToken = CreateDeadlineCancellationToken(options, method.FullName);
+            var callCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(options.CancellationToken, deadlineToken);
 
-            var httpResponseTask = SendAsync(restMethod, host, options, request, cancellationTokenSource.Token, HttpCompletionOption.ResponseContentRead);
+            var httpResponseTask = SendAsync(restMethod, host, options, request, callCancellationTokenSource.Token, deadlineToken, HttpCompletionOption.ResponseContentRead);
             var readResponseTask = ReadResponseAsync(httpResponseTask);
 
             var responseTask = restMethod.ReadResponseAsync<TResponse>(readResponseTask);
             var responseHeadersTask = ReadHeadersAsync(readResponseTask);
             Func<Status> statusFunc = () => GetStatus(readResponseTask);
             Func<Metadata> trailersFunc = () => GetTrailers(readResponseTask);
-            return new AsyncUnaryCall<TResponse>(responseTask, responseHeadersTask, statusFunc, trailersFunc, cancellationTokenSource.Cancel);
+            return new AsyncUnaryCall<TResponse>(responseTask, responseHeadersTask, statusFunc, trailersFunc, callCancellationTokenSource.Cancel);
         }
 
+        /// <summary>
+        /// Creates an HTTP request, adds headers from CallOptions, and sends the request.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -64,26 +64,17 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
-                    throw new RpcException(new Status(StatusCode.DeadlineExceeded, $"The timeout was reached when calling a method `{restMethod.FullName}`"));
-                }
-                cancellationTokenSource = new CancellationTokenSource(delayInterval);
-            }
+            var deadlineToken = CreateDeadlineCancellationToken(options, method.FullName);
+            var methodDisposalCts = new CancellationTokenSource();
 
-            var httpResponseTask = SendAsync(restMethod, host, options, request, cancellationTokenSource.Token, HttpCompletionOption.ResponseContentRead);
+            var httpResponseTask = SendAsync(restMethod, host, options, request, deadlineToken, HttpCompletionOption.ResponseContentRead);
             var readResponseTask = ReadResponseAsync(httpResponseTask);
```
amanda-tarafa, line 71: I think the linked token built by SendAsync should be passed to ReadResponseAsync as well and used to wrap the content reading call with something similar to AwaitWithCancellation.

amanda-tarafa (commented):
I've left some comments, but I might be missing something.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -125,33 +120,41 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
...
 
-            Task<HttpResponseMessage> httpResponseTask = SendAsync(restMethod, host, options, request, cancellationTokenSource.Token, HttpCompletionOption.ResponseHeadersRead);
+            Task<HttpResponseMessage> httpResponseTask = SendAsync(restMethod, host, options, request, deadlineCancellationToken, HttpCompletionOption.ResponseHeadersRead);
 
             Task<Metadata> responseHeadersTask = ReadHeadersAsync(httpResponseTask);
             Func<Status> statusFunc = () => GetStatus(ReadResponseAsync(httpResponseTask));
             Func<Metadata> trailersFunc = () => GetTrailers(ReadResponseAsync(httpResponseTask));
 
-            PartialDecodingStreamReader<TResponse> responseStream = restMethod.ResponseStreamAsync<TResponse>(httpResponseTask);
+            CancellationTokenSource callCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(options.CancellationToken);
```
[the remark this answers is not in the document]
jskeet: I think my change (new commit) should fix this.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
amanda-tarafa: OK, but I still think that methodDisposalCts.Token should be passed to ReadResponseAsync (which currently expects no token), so the task that actually reads the content is cancallable as well.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
```diff
@@ -65,9 +65,9 @@ public RestChannel(RestServiceCollection serviceCollection, string endpoint, Cha
             var restMethod = _serviceCollection.GetRestMethod(method);
 
             var deadlineToken = CreateDeadlineCancellationToken(options, method.FullName);
-            var methodDisposalCts = new CancellationTokenSource();
+            var methodDisposalCts = CancellationTokenSource.CreateLinkedTokenSource(options.CancellationToken, deadlineToken);
```
amanda-tarafa, line 68: nit: methodDisposalCts => callCancellationTokenSource (just so that is the same as in the streaming call)

amanda-tarafa (commented):
One comment left, otherwise looks good.

Google.Api.Gax.Grpc/Rest/RestChannel.cs
jskeet: Done.

jskeet:
> OK, but I still think that methodDisposalCts.Token should be passed to ReadResponseAsync (which currently expects no token), so the task that actually reads the content is cancallable as well.

I don't think that's required:

- ReadResponseAsync awaits the task it's passed, which is returned from SendAsync (which sets up cancellation)
- The SendAsync call uses HttpCompletionOption.ResponseContentRead, which means the returned task shouldn't return until the response is already present
- This means that ReadAsStringAsync in ReadResponseAsync shouldn't actually block

It's worth noting that none of the calls within ReadResponseAsync actually takes a cancellation token anyway.

I'm going to merge without that change, and if the above doesn't convince you, we can revisit it :)

jskeet:
Looks like I can't merge without another review after all :)

amanda-tarafa (approved):
LGTM

Convinced: I hadn't noticed the HttpCompletionOption.ResponseContentRead.

PR #660, opened 2022-12-07
fix: Throw RpcException if transcoding fails

### LINKED ISSUE #659: REGAPIC: which exception should be thrown if transcoding fails?
If a request is invalid in a way that makes transcoding fail (e.g. it's missing a required field, or that field's value is inappropriate) we currently throw an ArgumentException. If the request can be transcoded, but is otherwise invalid, the server will fail instead and we'll get an RpcException. Should we create an "artificial" RpcException instead in this case?

jskeet:
Fixes #659


amanda-tarafa (approved):
LGTM

PR #668, opened 2023-01-10
chore: Update dependencies in GAX prior to a release

jskeet:
Note that Grpc.Core is currently essentially frozen at 2.46.5. However, it appears to work fine with Grpc.Core.Api 2.51.0 - and copious integration tests running on Windows will prove that before libraries are released.


amanda-tarafa (approved):
LGTM

PR #677, opened 2023-02-02
chore: Remove testing with .NET Core 3.1

jskeet:
While we still support .NET Core 3.1 on a best-effort basis, it isn't supported by MS any more, and I'd rather not install an unsupported runtime where possible.

(This is the first step; we can do the same for google-cloud-dotnet afterwards, but this is much smaller.)


amanda-tarafa (approved):
LGTM

jskeet:
Not sure why I didn't merge this before - will rebase it etc then merge.

PR #688, opened 2023-03-28
feat: Regenerate common API protos

jskeet:
If this looks okay, I'll do a release tomorrow, so we can start using the generator options.


amanda-tarafa (approved):
LGTM. Yep, this looks good.

PR #704, opened 2023-07-20
feat: adds support for periodic channel refreshing

hemanshv:
This adds support for channel priming and refreshing. It will  be used in bigTable.
Its  a prototype PR so tests and comments are missing. We update them once the approach is finalized.

### DIFF
--- a/Google.Api.Gax.Grpc/Gcp/ChannelRef.cs
+++ b/Google.Api.Gax.Grpc/Gcp/ChannelRef.cs
@@ -6,6 +6,7 @@
  */
 
 using Grpc.Core;
+using System;
 using System.Threading;
 
 namespace Google.Api.Gax.Grpc.Gcp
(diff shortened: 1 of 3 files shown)


hemanshv:
Hi @jskeet and @amanda-tarafa,
Please have a look at it and share your feedback.

Thanks,

Google.Api.Gax.Grpc/Gcp/ChannelRef.cs
```diff
@@ -18,15 +19,29 @@ internal sealed class ChannelRef
     {
         private int _affinityCount;
         private int _activeStreamCount;
-        private int _id;
+        internal int _id;
+        internal Guid _channelId;
+        public  DateTime ChannelStartTime { get; set; }
```
jskeet, line 24: I'm not sure why these are public when the class is internal anyway.

Google.Api.Gax.Grpc/Gcp/ChannelRef.cs
```diff
@@ -18,15 +19,29 @@ internal sealed class ChannelRef
     {
         private int _affinityCount;
         private int _activeStreamCount;
-        private int _id;
+        internal int _id;
```
jskeet, line 22: All fields should be private (other than constants).

If you need these to be accessed internally, they should be exposed as properties (and should probably be read only)

Google.Api.Gax.Grpc/Gcp/GcpCallInvokerWithRefresh.cs
```diff
@@ -0,0 +1,539 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc.Gcp
+{
+    /// <summary>
+    /// Call invoker which can fan calls out to multiple underlying channels
+    /// based on request properties.
+    /// </summary>
+    public sealed class GcpCallInvokerWithRefresh : CallInvoker
```
jskeet, line 25: I'd really prefer us to avoid the duplication here. I was hoping we could use the existing GcpCallInvokerPool and GcpCallInvoker classes, either making `ChannelRef` itself handle the reference or (more likely) making GcpCallInvoker do so.

Google.Api.Gax.Grpc/Gcp/ChannelRef.cs
```diff
@@ -18,15 +19,29 @@ internal sealed class ChannelRef
...
+        internal Guid _channelId;
+        public  DateTime ChannelStartTime { get; set; }
+        public Action<CallInvoker> ChannelPrimer { get; set; }
+        public bool Grace { get; set; }
```
jskeet, line 26: Do these three properties really need to be read-write?

And at the very least I'd want "Grace" to have some documentation - the other properties are reasonably self-documenting, but Grace could mean almost anything.

jskeet (commented):
I haven't reviewed the details of the implementation, as I'd prefer a different design. Will write more in the design doc.

Google.Api.Gax.Grpc/Gcp/GcpCallInvokerWithRefresh.cs
hemanshv: Noted Jon, I will raise another draft pr for that option. 
Thanks for providing your feedback,

PR #706, opened 2023-07-25
chore: Add one more x-goog-request-param test

jskeet:
(This is relevant due to the way that Grpc.Net.Client concatenates headers.)


hemanshv (approved):
LGTM

PR #717, opened 2023-08-31
fix: Encode unknown enums as numeric values for REGAPIC paths and query parameters, and routing headers

### LINKED ISSUE #716: Encode unknown enum values as numbers in URLs, query parameters, and routing headers
ProtobufUtilities.FormatValueAsJsonPrimitive is used for URL segments, query parameters, and routing headers.

Current behavior is to return the original proto enum value name, or null if the value is unknown. We should *at least* encode unknown values as the numeric value (which is what JSON formatter does).

jskeet:
To explain how routing headers are affected by this: ProtobufUtilities.FormatValueAsJsonPrimitive is called by RoutingHeaderExtractor.FormatRoutingHeaderValue, which is called by the generated code.

Fixes #716.

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/Rest/transcoder_tests.json
+++ b/Google.Api.Gax.Grpc.Tests/Rest/transcoder_tests.json
@@ -318,6 +318,22 @@
       }
     },
     {
+      "name": "ComplexQuery_UnknownEnumValues",
+      "details": "Unknown enum values are represented numerically",
+      "inlineRule": {
+        "get": "/abc/{in_path}"
+      },
+      "requestMessageName": "QueryTestMessage",
+      "request": {
+        "enumField": 123,
+        "inPath": "def"
+      },
+      "success": {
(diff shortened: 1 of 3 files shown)


Google.Api.Gax.Grpc.Tests/Rest/transcoder_tests.json
```diff
@@ -317,6 +317,22 @@
         "uri": "/abc/def?enumField=FIRST_VALUE&requiredInteger=0"
       }
     },
+    {
+      "name": "ComplexQuery_UnknownEnumValues",
+      "details": "Enum values are represented with their original names",
```
amanda-tarafa, line 322:
```suggestion
      "details": "Enum values are represented with their numeric values",
```

jskeet: Whoops - fixed (with slightly different wording, but basically it's there)

PR #724, opened 2023-09-15
feat: Regenerate common protos

jskeet:
- New policy-related messages (and reference in Control)
- New field behavior of IDENTIFIER


jskeet:
(This is primarily to make some versioning-related experimentation simpler)

amanda-tarafa (approved):

Issue #725, opened 2023-09-25
Check IndexOf/StartsWith/EndsWith operations


jskeet:
We should almost certainly be using ordinal comparisons all the time.

jskeet:
Fixed in #726

PR #726, opened 2023-09-26
fix: Use ordinal string operations

jskeet:
Checked:

- IndexOf
- StartsWith
- EndsWith
- OrderBy
- Sort
- Compare

(Annoyingly, for StartsWith and EndsWith there are various single-character cases where a character literal would do the right thing - but those overloads aren't supported on .NET Framework.)


jskeet:
Fixes #725.

amanda-tarafa (approved):

PR #729, opened 2023-10-05
feat: Include protos in the Google.Api.CommonProtos package

### LINKED ISSUE #715: Add proto files to `Google.Api.CommonProtos`
I'm always frustrated when I have to use C# with Protobuf to generate code from proto definitions.

The package `Google.Protobuf.Tools` contains the proto definitions for well known types.

These can be used in my own proto definitions.
```xml
<PackageReference Include="Google.Protobuf.Tools" Version="3.24.2" GeneratePathProperty="true"/>

<Protobuf AdditionalImportDirs="$(PkgGoogle_Protobuf_Tools)\tools" />
```

But the `Google.Api.CommonProtos` package does not contain its proto definitions.

Whenever someone needs to use some definition from here:
[googleapis/googleapis](https://github.com/googleapis/googleapis/tree/master/google/)

these files need to be provided manually somehow.

For java there exists a package that contains those protos
[com.google.api.grpc:proto-google-common-protos](https://mvnrepository.com/artifact/com.google.api.grpc/proto-google-common-protos)
and which are picked up by the different build tool plugins for proto code generation.

Can you add the proto definitions for googleapis to the `Google.Api.CommonProtos` package?

jskeet:
(And make them easy to consume with Grpc.Tools.)

Fixes #715.

cc @JamesNK @jtattermusch @tonydnewell


updateprotos.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
amanda-tarafa, line 1: This is to be called manually?

amanda-tarafa (approved):

updateprotos.sh
jskeet: Yup. (Like generateprotos.sh is today. We do this on an ad-hoc basis.)

JamesNK:
A couple of questions:
* Have you tested this end-to-end?
* The PR adds proto files to the repo. Are these a copy? Wouldn't the proto files already exist somewhere in the repo for tooling to generate code from?

This change might remove the need for me to include `http.proto` in Microsoft.AspNetCore.Grpc.JsonTranscoding. Someone can set the flag to get Google.Api.CommonProtos to include proto files instead.

jskeet:
@JamesNK responses inline:

> * Have you tested this end-to-end?

My end-to-end test was:

- Generate the package in a local repository
- Create a new console app, and add a dependency on Grpc.Tools and the test version of the package
- Create a proto that uses ClientLibrarySettings within a message
- Add a Protobuf msbuild element for the proto
- Observe that it fails to build
- Add the msbuild property (IncludeGoogleApiCommonProtos) and observe that this fixes the build failure (and logs the location)
- Write code to populate create an instance of the message and populate the ClientLibrarySettings
- Build the code: all is well

(I did run into some problems within that, when trying to write code using the message without having done a successful build beforehand... but I think that's just an ordering aspect, and not specific to this change.)

> * The PR adds proto files to the repo. Are these a copy? Wouldn't the proto files already exist somewhere in the repo for tooling to generate code from? If so, I wonder if `<None Include="protos\**\*.proto" Pack="true" PackagePath="content/protos" />` could be modified to use LinkBase to reference existing files and avoid duplicate.

The repo doesn't currently contain the protos. The previous procedure would basically fetch them to a temporary directory and generate the code, but not add the protos themselves to the repo. (This is what we do for all the client libraries, too.) So the presence of the protos is new. I hope that makes sense - please let me know if it doesn't.

> This change might remove my need to include `http.proto` in Microsoft.AspNetCore.Grpc.JsonTranscoding. Someone can set the flag to get Google.Api.CommonProtos to include proto files instead.

Right. Presumably it'll be fine to set *either* IncludeGoogleApiCommonProtos *or* IncludeHttpRuleProtos to true... we might want to recommend against doing both, as I suspect it's non-deterministic (or at least "hard to work out") which version would actually be used in that case.

tonydnewell:
LGTM

jskeet:
Hearing no further issues, I'll merge and do a release.

PR #730, opened 2023-10-11
Release Google.Api.CommonProtos version 2.12.0

jskeet:
Changes since 2.11.0:

- The package contains the original protos, in contents/protos. Build targets in the package set the `GoogleApiCommonProtos_ProtosPath` msbuild property, and if `IncludeGoogleApiCommonProtos` is set to the value `true` in the project, the `Protobuf_StandardImportsPath` property is updated to include it before the `Protobuf_BeforeCompile` target is executed. (This should make the package easier to work with from Grpc.Tools.)

### DIFF
--- a/Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
+++ b/Google.Api.CommonProtos/Google.Api.CommonProtos.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\CommonProperties.xml" />
 
   <PropertyGroup>
-    <Version>2.11.0</Version>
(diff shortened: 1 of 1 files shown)


amanda-tarafa (approved):

jskeet:
Will look into this tomorrow.

amanda-tarafa:
This is the same as always with releases from Gax. The release triggered twice, one succeeded that other failed.

Here's the updated package: https://www.nuget.org/packages/Google.Api.CommonProtos

jskeet:
Ah, phew :)

Issue #712, opened 2023-08-04
Evaluate Grpc.Net.Client changes for GAX impact


jskeet:
See:

- https://github.com/grpc/grpc-dotnet/issues/2226
- https://github.com/grpc/grpc-dotnet/pull/2220
- https://github.com/grpc/grpc-dotnet/pull/2229

Changing from Grpc.Core to Grpc.Net.Client seamlessly is great if it works (although it could cause some surprises in terms of logging etc). It sounds like the only cases we need to worry about are ".NET Framework, on Windows Server 2022, with an API that uses client/bidi-streaming." We could potentially detect whether any APIs use client/bidi-streaming and still fall back to Grpc.Core when on .NET Framework by default for those - or do the more complex OS check if necessary. We should work out how many APIs actually have client/bidi-streaming methods.

PR #735, opened 2023-11-07
fix: Update gRPC adapter detection to always use Grpc.Core on .NET Framework

jskeet:
Currently, users who have a recent (extra) Grpc.Net.Client dependency and are using .NET Framework will default to Grpc.Net.Client, even if it won't work due to client/bidi-streaming calls.

Eventually we'll detect this properly, but for now we can just use Grpc.Core unconditionally on .NET Framework. (This can still be overridden with the environment variable or explicit code, of course.)

This is part of #712, but doesn't fully address it.

### DIFF
--- a/Google.Api.Gax.Grpc/GrpcAdapter.cs
+++ b/Google.Api.Gax.Grpc/GrpcAdapter.cs
@@ -115,19 +115,42 @@ namespace Google.Api.Gax.Grpc
             };
         }
 
-        // TODO: Is this really what we want? Definitely simple, but not great in other ways...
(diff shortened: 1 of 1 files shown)


amanda-tarafa (approved):
LGTM, in case you want to merge this. But if the sample in https://github.com/grpc/grpc-dotnet/issues/2242 could be turned into an API as Tony is proposing, that would be ideal.

jskeet:
Still waiting for a review on that snippet - I think we should go with this for now.

amanda-tarafa (approved):
Yep, I think this is good also as is.
If the sample is reviewd, then we can use it, and if it get turns into an API then we can call that API.

PR #747, opened 2024-01-04
Update StatusPartial.cs

JamesNK:
Mismatch between code sample and the type's name.


jskeet (approved):
Whoops, sorry for not spotting this before.

jskeet:
(I'm not worried by the conventional commit warning here - it's not something we need in release notes. Any future commits you create for this repo would be best to have a conventional commit message, but that's all.)

PR #758, opened 2024-02-08
feat: Add a FieldFormats static class for auto-populated fields

jskeet:
Fixes b/324254628

### DIFF
--- a/Google.Api.Gax.Tests/FieldFormatsTest.cs
+++ b/Google.Api.Gax.Tests/FieldFormatsTest.cs
@@ -1,0 +1,28 @@
+﻿/*
+ * Copyright 2024 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Xunit;
+
+namespace Google.Api.Gax.Tests;
+
+public class FieldFormatsTest
+{
+    [Fact]
+    public void GenerateUuid4()
+    {
+        // We generate lots of values to check both uniqueness and formatting.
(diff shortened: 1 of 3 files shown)


Google.Api.Gax/FieldGenerators.cs
```diff
@@ -0,0 +1,27 @@
...
+// TBD:
+// - Class name
+// - Method names
+// - Package
+// Basically everything!
+
+/// <summary>
+/// Generator methods for auto-populated fields.
+/// </summary>
+public static class FieldGenerators
```
amanda-tarafa, line 21: `FieldValueGenerators` or `ValueGenerators` ? I think just Field is confusing, specially when quickly seen in generator code.

Is something other than value generation that we expect to be doing here? Validations are unlikely I guess. But maybe canonicalization? Just thinking whether we want *Generators and the hypotethetical *Canonicalizations or simply *Helpers. (* is Field or FieldValue or Value).

Google.Api.Gax/FieldGenerators.cs
```diff
@@ -0,0 +1,27 @@
+﻿/*
+ * Copyright 2024 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using System;
+
+namespace Google.Api.Gax;
```
amanda-tarafa, line 10: Yep, this is fine.

(I'd really like if we had a base package that both Apiaries and Gax/GAPICs could use, we'd then could put this kind of thing there (and WithCancellationToken :)). But that's something for another day.)

Google.Api.Gax/FieldGenerators.cs
jskeet: Changed to FieldFormats.

amanda-tarafa (approved):

PR #763, opened 2024-03-11
feat: Implement API versioning in GAX

jskeet:
Once this is in, we'll need to:

- Release the common protos (can be a GA)
- Release GAX (probably as a beta)
- Update the generator
- At some point, release both GAX (as GA) and the generator

### DIFF
--- a/Google.Api.Gax.Grpc/ClientHelper.cs
+++ b/Google.Api.Gax.Grpc/ClientHelper.cs
@@ -17,7 +17,15 @@ namespace Google.Api.Gax.Grpc
     /// </summary>
     public class ClientHelper
     {
+        // Visible for testing
+        internal const string ApiVersionHeaderName = "x-goog-api-version";
+
         private readonly CallSettings _clientCallSettings;
+
+        /// <summary>
+        /// Call settings specifying headers for the client version (x-goog-api-client) and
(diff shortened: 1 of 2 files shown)


jskeet:
Right, this is ready for review again - one commit at a time, as ever.

This PR now *doesn't* include common protos, but *does* include a refactoring for ClientHelper construction, which makes things significantly simpler. I haven't tried making the corresponding changes in the generator, but I can't imagine it'll be *terribly* hard. (There's just a single line of code which should need changing, in ServiceImplClientClassGenerator - `clientHelper.WithInitializer(New(_ctx.Type<ClientHelper>())(effectiveSettings, logger)),`).

Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -24,12 +24,27 @@ public class ClientHelper
         /// Constructs a helper from the given settings.
         /// Behavior is undefined if settings are changed after construction.
         /// </summary>
+        /// <remarks>
+        /// This constructor will be removed in the next major version of GAX.
```
amanda-tarafa, line 36: Mark as obsolete?

amanda-tarafa (approved):
LGTM with a tiny comment.

PR #764, opened 2024-03-12
feat: Update common protos

jskeet:
I'm separating this from #763 as that's likely to take a little longer with the change to the ClientHelper approach.


jskeet:
(As it happens, the GAX changes for API versioning don't even need the common proto changes... it's the generator that needs both.)

amanda-tarafa (approved):

jskeet:
BEGIN_COMMIT_OVERRIDE
chore: update common protos
END_COMMIT_OVERRIDE

(This comment is to fix the release notes in #765 which shouldn't include common protos)

PR #766, opened 2024-03-12
chore: updating blunderbuss config for default assignee

shivgautam:


amanda-tarafa:
Closing, as we'll do this using a synced group. Thanks @shivgautam

PR #769, opened 2024-03-13
Open Telemetry prototype

jskeet:
The intention of this PR is that we merge *some* parts of it, but not all of it - ideally whole commits should be in or out.

### DIFF
--- a/Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
+++ b/Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
@@ -1,0 +1,51 @@
+﻿/*  
+ * Copyright 2024 Google LLC All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc;
+
+/// <summary>
+/// Extension methods to provide tracing via <see cref="ActivitySource"/>.
+/// </summary>
+internal static class ApiCallTracingExtensions
+{
+    // Unary async
+    internal static Func<TRequest, CallSettings, Task<TResponse>> WithTracing<TRequest, TResponse>(
+        this Func<TRequest, CallSettings, Task<TResponse>> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+
+    // Unary sync
+    internal static Func<TRequest, CallSettings, TResponse> WithTracing<TRequest, TResponse>(
+        this Func<TRequest, CallSettings, TResponse> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+
+    // Server-streaming async
+    internal static Func<TRequest, CallSettings, Task<AsyncServerStreamingCall<TResponse>>> WithTracing<TRequest, TResponse>(
+        this Func<TRequest, CallSettings, Task<AsyncServerStreamingCall<TResponse>>> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+
+    // Server-streaming sync
+    internal static Func<TRequest, CallSettings, AsyncServerStreamingCall<TResponse>> WithTracing<TRequest, TResponse>(
+        this Func<TRequest, CallSettings, AsyncServerStreamingCall<TResponse>> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+
+    // Client-streaming
+    internal static Func<CallSettings, AsyncClientStreamingCall<TRequest, TResponse>> WithTracing<TRequest, TResponse>(
+        this Func<CallSettings, AsyncClientStreamingCall<TRequest, TResponse>> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+
+    // Bidi-streaming
+    internal static Func<CallSettings, AsyncDuplexStreamingCall<TRequest, TResponse>> WithTracing<TRequest, TResponse>(
+        this Func<CallSettings, AsyncDuplexStreamingCall<TRequest, TResponse>> fn, ActivitySource activitySource, string methodName) =>
+        fn;
+}
--- a/Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
+++ b/Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
@@ -7,9 +7,7 @@
 
 using Grpc.Core;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Text;
 using System.Threading.Tasks;
 
 namespace Google.Api.Gax.Grpc;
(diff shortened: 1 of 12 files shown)


Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
```diff
@@ -19,15 +17,64 @@ namespace Google.Api.Gax.Grpc;
...
-        this Func<TRequest, CallSettings, Task<TResponse>> fn, ActivitySource activitySource, string methodName) =>
-        fn;
+        this Func<TRequest, CallSettings, Task<TResponse>> fn, ActivitySource activitySource, string methodName)
+    {
+        GaxPreconditions.CheckNotNull(activitySource, nameof(activitySource));
+        var activityName = FormatActivityName(fn, methodName);
+        return async (request, callSettings) =>
+        {
+            using var activity = activitySource.StartActivity(activityName, ActivityKind.Client);
+            activity?.SetTag(GrpcCallTypeTag, UnaryCallType);
```
amanda-tarafa, line 40: Can the activity be null?

amanda-tarafa, line 40: I think we should have a tag with the name of the client, in case users pass in their own activity source, we may want to still be able to identify this activity as produced by us.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
```diff
@@ -48,4 +95,36 @@ internal static class ApiCallTracingExtensions
     internal static Func<CallSettings, AsyncDuplexStreamingCall<TRequest, TResponse>> WithTracing<TRequest, TResponse>(
         this Func<CallSettings, AsyncDuplexStreamingCall<TRequest, TResponse>> fn, ActivitySource activitySource, string methodName) =>
         fn;
+
+    // This is still very much up in the air, and may even require changes to the parameters, so that we get more information
+    // (e.g. the full RPC name, the client name etc).
+    private static string FormatActivityName(Delegate fn, string methodName) => $"{fn.Method.Name}/{methodName}";
+
+    // TODO: See if there's a standard way of doing this. It seems odd to have to do it ourselves.
```
amanda-tarafa, line 103: I was thinking the same.
This actually looks more like a logging concern than a tracing one? But yes, let's see what others do or if there are best practices etc.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
```diff
@@ -19,15 +17,64 @@ namespace Google.Api.Gax.Grpc;
...
+            activity?.SetTag(GrpcCallTypeTag, UnaryCallType);
+            try
+            {
+                var response = await fn(request, callSettings).ConfigureAwait(false);
+                activity?.SetStatus(ActivityStatusCode.Ok);
+                return response;
+            }
+            catch (Exception ex) when (SetActivityException(activity, ex))
+            {
+                // We'll never actually get here, because SetActivityException always returns false.
```
amanda-tarafa, line 49: I think it'll be way clearer to have the SetActivityException method be void, and just call it here, and then thrwo. And that's not just because that's actually what happens, but also because the `bool SetActivityException` combination makes me thingkthat this method will return true if it set the info, and false otherwise, when what happens is, it always sets the info but always returns false.

amanda-tarafa (commented):
I've left a few comments, but of the things we knew we had to chat about anyway.

Google.Api.Gax/VersionHeaderBuilder.cs
```diff
@@ -93,7 +93,10 @@ private static string GetEntryAssemblyVersionOrNull()
             }
         }
         
-        private static string FormatAssemblyVersion(System.Type type)
+        /// <summary>
+        /// Formats the version of the assembly containing the specified type, in a "generally informative" way.
+        /// </summary>
+        internal static string FormatAssemblyVersion(System.Type type)
```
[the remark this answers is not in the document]
jskeet: I agree, but I don't have a better place for now. Will add a TODO for now. One thing that's worth noting is that the "is it a valid header value" may not be relevant in a more general case (and *may* not be reasonable for OTel, even).

Google.Api.Gax/ActivitySources.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        GaxPreconditions.CheckNotNull(type, nameof(type));
+        return s_activitySources.GetOrAdd(type, CreateActivitySource);
+    }
+
+    /// <summary>
+    /// Returns an <see cref="ActivitySource"/> named after the given client type.
+    /// </summary>
+    /// <remarks>
+    /// Multiple calls to this method (or <see cref="FromClientType{T}()"/>) for the same
```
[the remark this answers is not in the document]
jskeet: Done.

Google.Api.Gax/ActivitySources.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// type will return the same cached activity source.
+    /// </remarks>
+    /// <param name="type">The client type to obtain an activity source for. Must not be null.</param>
+    /// <returns>An <see cref="ActivitySource"/> named after the given client type.</returns>
+    public static ActivitySource FromClientType(System.Type type)
```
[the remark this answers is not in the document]
jskeet: I was wondering about that. They *could* be used for any type, but we don't want to give the impression that they *would* be used by any other system by default. Let's talk about it more - I know you're not 100% convinced by having this at all.

Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -214,7 +214,8 @@ public sealed class Options
             public string ApiVersion { get; set; }
 
             /// <summary>
-            /// The activity source to use for tracing, if any. This may be null.
+            /// The activity source to use for tracing, if any. This may be null. This is ignored
+            /// if <see cref="Settings"/> specifies an activity source.
```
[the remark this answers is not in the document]
jskeet: That wouldn't be a valid cref. Basically we get to *either* refer to the Settings property, or ServiceSettingsBase.ActivitySource. We could do both as: `if <see cref="ServiceSettingsBase.ActivitySource" /> specified via <see cref="Settings" /> is not null"?

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Google.Api.Gax.Grpc;
+
+/// <summary>
+/// Extension methods to provide tracing via <see cref="ActivitySource"/>.
+/// </summary>
+internal static class ApiCallTracingExtensions
+{
+    // Unary async
+    internal static Func<TRequest, CallSettings, Task<TResponse>> WithRetry<TRequest, TResponse>(
```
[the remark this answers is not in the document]
jskeet: Doh, thanks. Done.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
jskeet: Yes - StartActivity returns null if nothing is listening to it.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
jskeet: I was worried about `throw` losing context in some async situations.

We could use `ExceptionDispatchInfo.Capture(e.InnerExceptions.FirstOrDefault() ?? e).Throw();` instead, perhaps.
(I'm aware that using an exception filter here is a bit odd.)

Will add a comment so we don't forget to come back to this.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
jskeet: Maybe. I've added a TODO so that we remember to look at it - basically I think we'll need a whole design doc just about what names/tags we want to expose.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
jskeet: I think setting the status to be an error is probably important for tracing though.

Google.Api.Gax.Grpc/ClientHelper.cs
```diff
@@ -104,6 +108,7 @@ public ClientHelper(Options options)
             // These operations are applied in reverse order.
             // I.e. Version header is added first, then retry is performed.
             return ApiCall.Create(methodName, asyncGrpcCall, syncGrpcCall, baseCallSettings, Clock)
+                .WithTracing(_activitySource)
```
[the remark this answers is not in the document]
jskeet: Thereby hangs a tail... currently we're using the method name of the delegate in WithTracing to get the activity name, which would be messed up by WithLogging. We should almost certainly stop doing that, basically - it makes all of this code really brittle. But we can decide all of that when we revisit the activity (and activity source) name.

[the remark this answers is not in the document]
amanda-tarafa: Ah yes! Definetely a strong reason to stop using the name of the delegate in my view. As in I think it's more important (an expected) that the logs are contained in the trace. The methods themselves can access the current activity and set their own name as a tag, so we can still make that information available easily.

Google.Api.Gax.Grpc/ClientHelper.cs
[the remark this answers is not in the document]
amanda-tarafa: Ah yes, this always trips me, even when I'm writing docs myself.

I think as is is fine then.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
amanda-tarafa: Yes, agreed we need to desing this elsewhere: naming, tags, and events and spans, and what corresponds to what, for instance, is a retry attempt represented in it's own span or is it an event? And what name it has, etc.

Google.Api.Gax.Grpc/ApiCallTracingExtensions.cs
amanda-tarafa: Yep, the status is relevant, but the exception message and the stack trace seem odd in tracing, at least to me.

amanda-tarafa (commented):
This looks good, I think we mostly agree on everything including in what needs further discussion. And thanks for the TODOs!

jskeet:
Great, thanks. We can talk in a VC about how we want to make progress. (We could potentially make everything internal and start merging the bits we're happier with, for example.)

krugre999 (approved):

krugre999:
Duck u

amanda-tarafa:
I'm closing this as it's been sitting here for a long time.

PR #771, opened 2024-03-20
feat: Target netstandard2.0 in GAX

jskeet:
This will allow us to target netstandard2.0 in client libraries as well. This is experimental at this stage - we're likely to release a beta with these targets, and then release a beta client library targeting netstandard2.0 with the GAX beta as a dependency, but a GA GAX release targeting netstandard2.0 is further out.

Note: the [compatibility tester](https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator) currently complains about various members, but I *believe* that's a bug in the tester - see https://github.com/dotnet/sdk/issues/39629 for details. We should at least confirm that it's a bug before doing a GA release, but I think an alpha or beta would be okay before then.

### DIFF
--- a/Google.Api.Gax.Grpc.Testing/Google.Api.Gax.Grpc.Testing.csproj
+++ b/Google.Api.Gax.Grpc.Testing/Google.Api.Gax.Grpc.Testing.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\CommonProperties.xml" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
   </PropertyGroup>
 
   <!-- Packaging information -->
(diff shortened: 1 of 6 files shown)


amanda-tarafa (approved):
LGTM. And yes, this should be OK for a beta.

One thing to try is a project using using a library targeting the stable GAX (netstandard2.0) and a library targetting the beta GAX (netstandard2.1) and see if there are any issues (there shouldn't be any I think).

Also, once we get the stable GAX released, we'd probably want to release all libraries in one go?

jskeet:
Yes, it would be good to try a combination of both, and shouldn't be hard to do.

As for releasing everything - *maybe*. It might be simplest to do a "release as normal for 3 months, and then do a catch-up set of releases for things which haven't been released in a while". We'll need to change the breaking change detector to look for netstandard2.0...

jskeet:
Release-As: 4.8.0-alpha01

jskeet:
BEGIN_COMMIT_OVERRIDE
feat: Target netstandard2.0 in GAX

Release-As: 4.8.0-alpha01
END_COMMIT_OVERRIDE

PR #773, opened 2024-03-22
feat: Start implementing OpenTelemetry support

jskeet:
This is all internal at the moment - the only public-facing change is the dependency on System.Diagnostics.DiagnosticSource. Still, that *is* public so we may not want to merge this. Happy to chat.


amanda-tarafa (approved):
LGTM

I'm happy to merge this as is, including with the new dependency.

jskeet:
Right. Will wait until we've released the stable 4.8.0 before merging.

PR #779, opened 2024-04-25
tests: Remove duplicate test case

jskeet:
(I'm happy to change it instead of removing it if something else was intended, but I *think* we still cover everything.)

Fixes b/336944350


amanda-tarafa (approved):
Yep, I think that was just accidental duplicate.

PR #786, opened 2024-07-18
feat: Make the gRPC ClientBuilderBase<T>.ApiKey property public

jskeet:
This adds an imposition on concrete client builders to call GetEffectiveSettings, where previously it was only required when API keys were exposed in the concrete builder. However:

- We don't expect any customers to derive from ClientBuilderBase themselves
- We've been generating calls to GetEffectiveSettings unconditionally since generator version 1.4.24 (adopted on January 29th 2024)

In theory, a user could use a new version of GAX with an old client library (more than 6 months old), set an API key and find that the API key isn't sent (and no other auth is performed, including ADC). This feels very unlikely to be a problem.

We should, however, check any hand-written client builders in google-cloud-dotnet.

Fixes b/353517066.

### DIFF
--- a/Google.Api.Gax.Grpc/ClientBuilderBase.cs
+++ b/Google.Api.Gax.Grpc/ClientBuilderBase.cs
@@ -175,7 +175,7 @@ namespace Google.Api.Gax.Grpc
         /// and ensure they call <see cref="GetEffectiveSettings{T}(T)"/> to potentially specify the API key header
         /// via CallSettings.
         /// </remarks>
-        protected string ApiKey { get; set; }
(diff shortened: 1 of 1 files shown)


jskeet:
We probably don't want to release this just yet - not until we actually want to start using API keys more. We could merge and just not do a release though.

amanda-tarafa (approved):

jskeet:
@amanda-tarafa: Are you happy for me to just merge right away, or should we chat first?

amanda-tarafa:
> @amanda-tarafa: Are you happy for me to just merge right away, or should we chat first?

Yes, that's fine.

PR #823, opened 2025-05-15
chore: Modify release scripts and remove kokoro configs

purva9413:


buildrelease.sh
```diff
@@ -12,27 +12,19 @@
 # Google.Api.Gax.Grpc.Testing.<version>.nupkg
 # Sample: buildrelease.sh 847118821d8a8b4ac2ea3bbe692acea29c14c4da
 
-# The repo is cloned into a fresh "releasebuild" directory.
-
 set -e
 
-if [ -z "$1" ]
+if [[ $# -gt 0 ]];
```
amanda-tarafa, line 17: Remove this validation as we don't have it anywhere else.

buildrelease.sh
```diff
@@ -12,27 +12,19 @@
 # Google.Api.Gax.Grpc.Testing.<version>.nupkg
 # Sample: buildrelease.sh 847118821d8a8b4ac2ea3bbe692acea29c14c4da
```
amanda-tarafa, line 13: Update this description to mention that we'll use the tag at HEAD, not from a given commit.

And also, the sample here on this line is outdated, this does not take a parameter anymore.

amanda-tarafa (changes requested):
Tiny changes only, otherwise, LGTM

buildrelease.sh
purva9413: Makes sense, removed this.

buildrelease.sh
purva9413: Removing the sample example altogether. Since the script does not take in any arg anymore, we do not need this sample example.

purva9413: Also fixed the description to mention we use the HEAD commit.

amanda-tarafa (approved):

amanda-tarafa:
@purva9413 please rebase this PR on main before merging.

PR #846, opened 2025-09-04
fix: Return default response if http content is empty

purva9413:
Fixes https://github.com/googleapis/google-cloud-dotnet/issues/15025

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/RestMethod.cs
+++ b/Google.Api.Gax.Grpc/Rest/RestMethod.cs
@@ -96,7 +96,12 @@ internal class RestMethod
         {
             throw new RpcException(status, httpResponse.GetTrailers());
         }
-        return ParseJson<TResponse>(httpResponse.Content);
+
+        string jsonToParse = httpResponse.Content;
+
+        // See b/436913122#comment13. Returns the default value on an empty response
+        // instead of converting to an empty JSON element string ("{}") to prevent JSON parsing issues.
+        return string.IsNullOrEmpty(jsonToParse) ? default : ParseJson<TResponse>(jsonToParse);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


Google.Api.Gax.Grpc/Rest/RestMethod.cs
```diff
@@ -96,7 +96,10 @@ internal async Task<TResponse> ReadResponseAsync<TResponse>(Task<ReadHttpRespons
         {
             throw new RpcException(status, httpResponse.GetTrailers());
         }
-        return ParseJson<TResponse>(httpResponse.Content);
+
+        string jsonToParse = httpResponse.Content;
+
+        return string.IsNullOrEmpty(jsonToParse) ? default : ParseJson<TResponse>(jsonToParse);
```
jskeet, line 102: That will return a null reference, which isn't ideal. I'd probably convert a null/empty response to `"{}"` instead, to be an *empty* response instead. @amanda-tarafa thoughts?

purva9413: I had a similar suggestion in b/436913122#comment12 but observation was the converting the empty response to `"{}"` for the JsonParser eventually returns the body as an empty json element `{ }` which is not ideal. This way the customer can do a null check before proceeding (and looks like that's what they are expecting from their experience with some other API calls in this [issue](https://github.com/googleapis/google-cloud-dotnet/issues/15025)). So after discussion with Amanda we decided to go with this default/null return. Lmk if that clarifies this PR a bit, happy to discuss.

jskeet: Right, have now had a conversation with Amanda as well, and I see. It's horrible, but yes, that may be right.
I think it would be worth adding a comment here to make it clear that's deliberate.

purva9413: Added in the comment.

PR #877, opened 2026-04-10
feat: add new CloudRunJob platform

PierreBart:
# Context

I use the https://github.com/manigandham/serilog-sinks-googlecloudlogging lib to push logs to Cloud Logging. The lib is able to infer the runtime context to enrich logs with attributes about the execution environment. It does so using the `MonitoredResourceBuilder`. 

It appears the class has no platform for Cloud Run Job, but only Cloud Run services.

I would like to add a `CloudRunJob` platform to enrich logs with the job name, execution name, and resource type.

### DIFF
--- a/Google.Api.Gax/CloudRunJobPlatformDetails.cs
+++ b/Google.Api.Gax/CloudRunJobPlatformDetails.cs
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2026 Google Inc. All Rights Reserved.
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Newtonsoft.Json.Linq;
+using System;
+using System.Linq;
+
+namespace Google.Api.Gax;
+
+/// <summary>
+/// Google Cloud Run Job details.
+/// </summary>
+public sealed class CloudRunJobPlatformDetails
+{
+    /// <summary>
+    /// Builds a <see cref="CloudRunJobPlatformDetails"/> from the given metadata
+    /// and Cloud Run Job environment variables.
+    /// The metadata is normally retrieved from the GCE metadata server.
+    /// </summary>
+    /// <param name="metadataJson">JSON metadata, normally retrieved from the GCE metadata server.
+    /// Must not be <c>null</c>.</param>
+    /// <returns>A populated <see cref="CloudRunJobPlatformDetails"/> if the metadata represents a Cloud Run Job;
+    /// <c>null</c> otherwise.</returns>
+    public static CloudRunJobPlatformDetails TryLoad(string metadataJson)
+    {
+        // Check environment variables first, to avoid more expensive JSON parsing.
+        string jobName = Environment.GetEnvironmentVariable("CLOUD_RUN_JOB");
+        if (jobName is null)
+        {
+            return null;
+        }
+
+        GaxPreconditions.CheckNotNull(metadataJson, nameof(metadataJson));
+        JObject metadata;
+        try
+        {
+            metadata = JObject.Parse(metadataJson);
+        }
+        catch
+        {
+            return null;
+        }
+
+        var projectId = metadata["project"]?["projectId"]?.ToString();
+        var zoneName = metadata["instance"]?["zone"]?.ToString();
+        if (projectId == null || zoneName == null || !Platform.s_zoneTemplate.TryParseName(zoneName, out var zoneResourceName))
+        {
+            return null;
+        }
+        string zone = zoneResourceName[1];
+        return new CloudRunJobPlatformDetails(metadataJson, projectId, zone, jobName);
+    }
+
+    /// <summary>
+    /// Constructs details of a Google Cloud Run job execution.
+    /// </summary>
+    /// <param name="metadataJson">JSON metadata, normally retrieved from the GCE metadata server.
+    /// Must not be <c>null</c>.</param>
+    /// <param name="projectId">The project ID. Must not be null.</param>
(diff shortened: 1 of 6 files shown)


Google.Api.Gax/CloudRunJobPlatformDetails.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// The metadata is normally retrieved from the GCE metadata server.
+        /// </summary>
+        /// <param name="metadataJson">JSON metadata, normally retrieved from the GCE metadata server.
+        /// Must not be <c>null</c>.</param>
+        /// <returns>A populated <see cref="CloudRunJobPlatformDetails"/> if the metadata represents and GCE instance;
+        /// <c>null</c> otherwise.</returns>
+        public static CloudRunJobPlatformDetails TryLoad(string metadataJson)
+        {
+            // Check environment variables first, to avoid more expensive JSON parsing.
+            string jobName = Environment.GetEnvironmentVariable("CLOUD_RUN_JOB");
```
PierreBart: Environment variables defined in Cloud Run job: https://docs.cloud.google.com/run/docs/container-contract#jobs-env-vars

amanda-tarafa:
We've been shying away from doing this kind of thing, because it's flimsy. We can't possibly cover all environments and can at most guarantee that the happy path works. Most of GCP running environments now support logging agents that will enrich logs you write to stdoutput with the right environmental information. See for instance [Cloud Run Logging - Write Container Logs](https://docs.cloud.google.com/run/docs/logging#container-logs) documentation.

In addition we don't support ourselves https://github.com/manigandham/serilog-sinks-googlecloudlogging.

I'll discuss internally with the team if this is something we'd like to do, but for expectations, it's unlikely. The general advice at the moment is to use the GCP environments logging agents/capabilities to enrich your log entries with environment specific information. As an alternative you can request this as a feature in  https://github.com/manigandham/serilog-sinks-googlecloudlogging and they might be willing to support it as a fallback to `MonitoredResourceBuilder`.

PierreBart:
Thank you @amanda-tarafa for the detailed feedback.

I'll wait for your follow-up and proceed based on your guidance.

Google.Api.Gax/CloudRunJobPlatformDetails.cs
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
amanda-tarafa: nit: 2026

Google.Api.Gax/CloudRunJobPlatformDetails.cs
```diff
@@ -0,0 +1,118 @@
...
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Newtonsoft.Json.Linq;
+using System;
+using System.Linq;
+
+namespace Google.Api.Gax
```
amanda-tarafa: nit: For new types, we use file scoped namespaces.

Google.Api.Gax/CloudRunJobPlatformDetails.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// </summary>
+        public string ExecutionName { get; }
+
+        /// <inheritdoc/>
+        public override string ToString() =>
+            $"[Cloud Run Job: ProjectId='{ProjectId}', Zone='{Zone}', " +
+            $"JobName='{JobName}', ExecutionName='{ExecutionName}']";
+    }
+
+
```
amanda-tarafa, line 100: nit: extra blank lines

Google.Api.Gax/Platform.cs
```diff
@@ -217,6 +222,16 @@ public Platform(CloudRunPlatformDetails cloudRunDetails)
...
 
+        /// <summary>
+        /// Construct with details of Google Cloud Run Job.
+        /// </summary>
+        /// <param name="cloudRunJobDetails">Details of Google Cloud Run Job.</param>
+        public Platform(CloudRunJobPlatformDetails cloudRunJobDetails)
+        {
+            CloudRunJobDetails = GaxPreconditions.CheckNotNull(cloudRunJobDetails, nameof(cloudRunJobDetails));
+        }
+
```
amanda-tarafa, line 233: nit: Remove extra blank line

Google.Api.Gax/CloudRunJobPlatformDetails.cs
amanda-tarafa: Thanks!

amanda-tarafa (changes requested):
There are a few nits, but other than that this looks good.

One thing to note is that our intention is to eventually have this logic move to [OTEL contribution packages](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Resources.Gcp) and leave GAX geared towards our own internal needs. But for now, and given that OTEL contrib is also depending on GAX, we'll accept this change.

Google.Api.Gax/CloudRunJobPlatformDetails.cs
PierreBart: Switched to file-scoped namespace.

PierreBart:
I've applied your suggestions.

Additionally, I have updated the platform description to match the `MonitoredResourceDescriptor` as defined in https://docs.cloud.google.com/logging/docs/api/v2/resource-list#resource-types, i.e., I removed the `execution` attribute.

amanda-tarafa (approved):

robertvoinescu-work (approved):

amanda-tarafa:
Thanks for this!

PR #903, opened 2026-08-14
feat: Support resumable upload

amanda-tarafa:
Please review one commit at a time.

### DIFF
--- a/Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * Copyright 2022 Google LLC
  * Use of this source code is governed by a BSD-style
  * license that can be found in the LICENSE file or at
@@ -21,14 +21,20 @@ internal sealed class HttpRuleTranscodingOutput : ITranscodingOutput
 {
     private const string ApplicationJsonMediaType = "application/json";
 
-    private IEnumerable<KeyValuePair<string, string>> _queryStringParameters;
-    private string _uriPath;
+    private readonly IEnumerable<KeyValuePair<string, string>> _queryStringParameters;
+    private readonly string _uriPath;
+    private readonly string _prefix;
     internal string Body { get; }
     internal HttpMethod Method { get; }
 
-    internal HttpRuleTranscodingOutput(HttpMethod method, string uriPath, IEnumerable<KeyValuePair<string, string>> queryStringParameters, string body) =>
-        (Method, _uriPath, _queryStringParameters, Body) =
-        (method, uriPath, queryStringParameters, body);
+    internal HttpRuleTranscodingOutput(HttpMethod method, string uriPath, IEnumerable<KeyValuePair<string, string>> queryStringParameters, string body)
+        : this(method, uriPath, queryStringParameters, body, prefix: null)
+    {
+    }
+
+    private HttpRuleTranscodingOutput(HttpMethod method, string uriPath, IEnumerable<KeyValuePair<string, string>> queryStringParameters, string body, string prefix) =>
+        (Method, _uriPath, _queryStringParameters, Body, _prefix) =
+        (method, uriPath, queryStringParameters, body, prefix);
 
     HttpRequestMessage ITranscodingOutput.ToHttpRequestMessage(string host)
     {
@@ -46,7 +52,10 @@ internal sealed class HttpRuleTranscodingOutput : ITranscodingOutput
     }
 
     internal HttpRuleTranscodingOutput WithAdditionalQueryParameter(string name, string value) =>
-        new HttpRuleTranscodingOutput(Method, _uriPath, _queryStringParameters.Concat(new[] { new KeyValuePair<string, string>(name, value) }), Body);
+        new HttpRuleTranscodingOutput(Method, _uriPath, _queryStringParameters.Concat(new[] { new KeyValuePair<string, string>(name, value) }), Body, _prefix);
+
+    internal HttpRuleTranscodingOutput WithPrefix(string prefix) =>
+        new HttpRuleTranscodingOutput(Method, _uriPath, _queryStringParameters, Body, prefix);
 
     /// <summary>
     /// Merges the uri path and the query string parameters, escaping them.
@@ -58,6 +67,15 @@ internal sealed class HttpRuleTranscodingOutput : ITranscodingOutput
     internal string GetRelativeUri()
     {
         var sb = new StringBuilder();
+        if (!string.IsNullOrEmpty(_prefix))
+        {
+            // We trim the trailing '/' but it may have not been present at all.
+            sb.Append(_prefix.TrimEnd('/'));
+        }
+        if (!_uriPath.StartsWith("/"))
+        {
+            sb.Append("/");
+        }
         sb.Append(_uriPath);
         bool sbHasParameters = false;
 
--- a/Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
+++ b/Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
@@ -17,7 +17,7 @@ namespace Google.Api.Gax.Grpc.Rest;
 /// The result of transcoding a protobuf request using an HttpRule.
 /// This is produced by <see cref="HttpRuleTranscoder"/>.
 /// </summary>
-internal sealed class TranscodingOutput
+internal sealed class HttpRuleTranscodingOutput : ITranscodingOutput
 {
     private const string ApplicationJsonMediaType = "application/json";
 
@@ -26,12 +26,11 @@ internal sealed class TranscodingOutput
     internal string Body { get; }
     internal HttpMethod Method { get; }
 
-    internal TranscodingOutput(HttpMethod method, string uriPath, IEnumerable<KeyValuePair<string, string>> queryStringParameters, string body) =>
+    internal HttpRuleTranscodingOutput(HttpMethod method, string uriPath, IEnumerable<KeyValuePair<string, string>> queryStringParameters, string body) =>
         (Method, _uriPath, _queryStringParameters, Body) =
         (method, uriPath, queryStringParameters, body);
 
-    // TODO: Rename to ToHttpRequestMessage?
-    internal HttpRequestMessage CreateRequest(string host)
+    HttpRequestMessage ITranscodingOutput.ToHttpRequestMessage(string host)
     {
         var relativeUri = GetRelativeUri();
         var uri = host is null ? new Uri(relativeUri, UriKind.Relative) : new UriBuilder { Host = host, Path = relativeUri }.Uri;
@@ -46,13 +45,13 @@ internal sealed class TranscodingOutput
         };
     }
 
-    internal TranscodingOutput WithAdditionalQueryParameter(string name, string value) =>
-        new TranscodingOutput(Method, _uriPath, _queryStringParameters.Concat(new[] { new KeyValuePair<string, string>(name, value) }), Body);
+    internal HttpRuleTranscodingOutput WithAdditionalQueryParameter(string name, string value) =>
+        new HttpRuleTranscodingOutput(Method, _uriPath, _queryStringParameters.Concat(new[] { new KeyValuePair<string, string>(name, value) }), Body);
 
     /// <summary>
     /// Merges the uri path and the query string parameters, escaping them.
     /// Ignores the possibility that the path can already have parameters or contain an anchor (`#`).
-    /// This method is visible for testing; production code should generally call <see cref="CreateRequest(string)"/>
+    /// This method is visible for testing; production code should generally call <see cref="ITranscodingOutput.ToHttpRequestMessage(string)"/>
     /// instead.
     /// </summary>
     /// <returns>The URI path merged with the encoded query string parameters</returns>
--- a/Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
+++ b/Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
@@ -24,18 +24,24 @@ internal sealed class ResumableUploadChunk
     /// <summary>
     /// The offset in <see cref="Buffer"/> where the chunk data begins.
     /// </summary>
-    internal long Offset { get; }
+    internal long BufferOffset { get; }
 
     /// <summary>
     /// The number of bytes in the chunk.
     /// </summary>
     internal int Count { get; }
 
-    internal ResumableUploadChunk(byte[] buffer, long offset, int count)
+    /// <summary>
+    /// The offset in the overall upload stream where this chunk begins.
+    /// </summary>
+    internal long UploadOffset { get; }
+
+    internal ResumableUploadChunk(byte[] buffer, long bufferOffset, int count, long uploadOffset)
     {
         Buffer = GaxPreconditions.CheckNotNull(buffer, nameof(buffer));
-        Offset = offset;
+        BufferOffset = bufferOffset;
         Count = count;
+        UploadOffset = uploadOffset;
     }
 }
 
--- a/Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
+++ b/Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2026 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+
+using Google.Protobuf;
+using Google.Protobuf.Reflection;
+using System;
+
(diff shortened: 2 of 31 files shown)


Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,363 @@
...
+        _currentOffset = chunkResponse.CommittedOffset ?? (_currentOffset + chunkRequest.UploadChunk.Count);
+
+        if (chunkResponse.IsFinal)
+        {
+            ReportProgress(ResumableUploadState.Finalized, _currentOffset);
+            return chunkResponse.ResponseBody;
+        }
+
+        ReportProgress(ResumableUploadState.Uploading, _currentOffset);
+        return await ReadNextChunkAsync().ConfigureAwait(false);
```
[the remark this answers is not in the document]
amanda-tarafa: Upload does not return committed offset, we just make the assumption that if we got a 200, we can move forward.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
```diff
@@ -58,6 +67,15 @@ internal HttpRuleTranscodingOutput WithAdditionalQueryParameter(string name, str
     internal string GetRelativeUri()
     {
         var sb = new StringBuilder();
+        if (!string.IsNullOrEmpty(_prefix))
+        {
+            // We trime the trailing '/' but it may have not been present at all.
```
robertvoinescu-work: NIT: *trim

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
```diff
@@ -0,0 +1,343 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc.Rest;
+
+/// <summary>
+/// Defines the contract for streams used in resumable upload operations,
+/// supporting reading, position rewinding, and optional length metadata.
+/// </summary>
+internal interface IResumableUploadStream : IDisposable
```
robertvoinescu-work, line 19: Since the underlying Stream implements `IAsyncDisposable`, should we also implement it here? It would help us handle cases such as when the underlying stream is a network stream.

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
```diff
@@ -0,0 +1,343 @@
...
+
+        if (count == 0)
+        {
+            return 0;
+        }
+
+        int totalRead = 0;
+
+        // Step 1: Satisfy read request from ring buffer if current position is behind underlying stream position
+        if (count > 0 && _currentPosition < _underlyingPosition)
```
robertvoinescu-work, line 222: NIT: We can technically remove this count > 0 since we return early if 0 or throw an exception if < 0, but I'm fine if we want to keep it as is for consistency with the rest. Your call.

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
```diff
@@ -0,0 +1,343 @@
...
+
+        _bufferCapacity = (int) Math.Min(bufferCapacity, int.MaxValue);
+        _leaveOpen = leaveOpen;
+
+        _ringBuffer = new byte[_bufferCapacity];
+    }
+
+    public long? Length => _isEofReached ? _underlyingPosition : null;
+
+    public bool TryRewind(long targetOffset)
```
robertvoinescu-work, line 196: NIT: The name makes me think the passed parameter will be how much we step back our _currentposition, but this is setting the current position. Maybe `TrySetToEarlierOffset`? Your call.

Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
```diff
@@ -0,0 +1,67 @@
...
+    internal long Offset { get; }
+
+    /// <summary>
+    /// The number of bytes in the chunk.
+    /// </summary>
+    internal int Count { get; }
+
+    internal ResumableUploadChunk(byte[] buffer, long offset, int count)
+    {
+        Buffer = GaxPreconditions.CheckNotNull(buffer, nameof(buffer));
```
robertvoinescu-work, line 41: NIT: I'm thinking we should fail early and add preconditions checks on the offset and count, that way we can be sure a chunk is always a valid object and we fail before sending an actual http request if not.
```
GaxPreconditions.CheckArgument(offset > 0 "The offset must be greater than 0")
GaxPreconditions.CheckArgument(count > 0 "The count must be greater than 0")
GaxPreconditions.CheckArgument(offset < buffer.length, nameof(offset), "Offset must not exceed bounds of the buffer.")
GaxPreconditions.CheckArgument(count + offset < buffer.length, "The count from the offset must not exceed bound of the buffer")

**Comment by amanda-tarafa** [comment]
(Test failure is I wrote an integration test as a unit test inadvertently, I'll move it to integration)

File: Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs [hunk 7]
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    /// <summary>
+    /// Begins a new resumable upload session.
+    /// </summary>
+    /// <param name="request">The initial request payload to initiate the upload.</param>
+    /// <param name="stream">The content stream to upload.</param>
+    /// <param name="uploadSettings">Optional settings overriding default upload configuration.</param>
+    /// <param name="callSettings">Optional call settings to apply to initial start RPC.</param>
+    /// <returns>A task returning the completed response object upon successful upload.</returns>
```
robertvoinescu-work, line 71: Can we add a note saying this will throw an exception if called twice:
```xml
/// <exception cref="System.InvalidOperationException">
/// Thrown if this method is called more than once, or if the session has already been started or resumed.
/// </exception>
```

robertvoinescu-work: Do we need to dispose after return?

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    /// <summary>
+    /// Resumes an existing resumable upload session using a known upload URI.
+    /// </summary>
+    /// <param name="uploadUri">The upload URI for the existing resumable session.</param>
+    /// <param name="stream">The content stream to upload.</param>
+    /// <param name="uploadSettings">Optional settings overriding default upload configuration.</param>
+    /// <returns>A task returning the completed response object upon successful upload.</returns>
+    public async Task<TResponse> ResumeUploadAsync(
```
robertvoinescu-work, line 102: I'm missing some context - what's the use case for `ResumeUploadAsync`, it seems like it is it to allow hooking into a resumable upload session after it's been explicitly established. What advantage does this provide over starting the call as a resumable upload session? I'm asking because currently it seems like we could remove this and just require anyone who wants to use these features to do so from the initial request.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,332 @@
...
+    /// <returns>A task returning the completed response object upon successful upload.</returns>
+    public async Task<TResponse> BeginUploadAsync(
+        TRequest request,
+        Stream stream,
+        ResumableUploadSettings uploadSettings = null,
+        CallSettings callSettings = null)
+    {
+        GaxPreconditions.CheckNotNull(request, nameof(request));
+        GaxPreconditions.CheckNotNull(stream, nameof(stream));
+        GaxPreconditions.CheckState(UploadUri is null, "Session has already been started or resumed.");
```
robertvoinescu-work, line 80: If we create another session off the same `UploadCall` the `BeginUploadAsync` could effectively be called twice. This could be avoided if `CreateSession` returns a cached value on subsequent calls.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,332 @@
...
+            lock (_lock)
+            {
+                return _chunkGranularity;
+            }
+        }
+    }
+
+    internal ResumableUploadSession(ApiResumableUploadCall<TRequest, TResponse> uploadCall)
+    {
+        _uploadCall = GaxPreconditions.CheckNotNull(uploadCall, nameof(uploadCall));
```
robertvoinescu-work, line 61: The session seems very short lived as is - `BeginUploadAsync` can only be called once and then this object becomes unusable. Why not take uploadCall as a method parameter instead? It wouldn't really be a session then, but we could avoid creating a new object for each upload.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,332 @@
...
+            return false;
+        }
+
+        // 2. Recoverable conditions:
+        // - Category 2 status codes (HTTP 400 InvalidArgument, HTTP 412 FailedPrecondition, HTTP 416 OutOfRange)
+        // - Category 1 status codes (transient errors: Unavailable, DeadlineExceeded, ResourceExhausted, Internal)
+        return ex.StatusCode == StatusCode.InvalidArgument
+            || ex.StatusCode == StatusCode.FailedPrecondition
+            || ex.StatusCode == StatusCode.OutOfRange
+            || ResumableUploadSettings.TransientErrorCodes.Contains(ex.StatusCode);
```
robertvoinescu-work: NOTE TO SELF: When do we get transient error codes up in this layer?

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,363 @@
...
+        ReportProgress(ResumableUploadState.OffsetReceived, _currentOffset);
+        return await ReadNextChunkAsync().ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Reading / Chunk Preparation State: Reads next chunk from stream and transitions to UploadChunkAsync.
+    /// </summary>
+    private async Task<TResponse> ReadNextChunkAsync()
+    {
+        int bytesRead = await _stream.ReadAsync(_chunkBuffer, 0, _chunkBuffer.Length, CancellationToken.None).ConfigureAwait(false);
```
robertvoinescu-work, line 253: Our chunk size is limited to 2.14 GB with the current logic. If a user specifies a chunk size greater then we default to max `int`. Is there a reason we don't support larger chunk sizes? - 2GB does seem large to be fair. If we go with this limit, can we be explicit and treat chunk size as `int` throughout? This would change  the proposed public surface slightly (`ResumableUploadSettings.ChunkSize`).

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
```diff
@@ -0,0 +1,363 @@
...
+            return false;
+        }
+
+        // 2. Recoverable conditions:
+        // - Category 2 status codes (HTTP 400 InvalidArgument, HTTP 412 FailedPrecondition, HTTP 416 OutOfRange)
+        // - Category 1 status codes (transient errors: Unavailable, DeadlineExceeded, ResourceExhausted, Internal)
+        return ex.StatusCode == StatusCode.InvalidArgument
+            || ex.StatusCode == StatusCode.FailedPrecondition
+            || ex.StatusCode == StatusCode.OutOfRange
+            || ResumableUploadSettings.TransientErrorCodes.Contains(ex.StatusCode);
```
robertvoinescu-work: Should we treat `TransientErrorCodes` as unrecoverable since we would have hit max attempts if they show up here (`ResumableUploadSettings.DefaultRetry`)?

robertvoinescu-work, line 361: Should transient errors be treated as non-recoverable at this point? It seems like if a transient error occurs here, it means we have already exhausted the maximum retry attempts configured in `ResumableUploadSettings`.

robertvoinescu-work (approved):
Looks mostly good to me. I'm approving for now so we can start getting this tested.

Google.Api.Gax.Grpc/Rest/ResumableUploadRequest.cs
amanda-tarafa: Acked, I'll add these later but the way this is used from ResumableUploadSession guarantees that all the indices are withing bounds.

Google.Api.Gax.Grpc/Rest/HttpRuleTranscodingOutput.cs
amanda-tarafa: Done

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
amanda-tarafa: In GAX we target .NET Standard 2.0 and .NET Framework, `Stream` does not implement `IAsyncDisposable` in neither.

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
amanda-tarafa: I like the name of the method, but I will change `targetOffset` to `targetPosition`. I'll do that on a later PR though.

Google.Api.Gax.Grpc/Rest/ResumableUploadStream.cs
amanda-tarafa: I had it out, and while reading the code, I "missed it" so, I added it back in, it reads nicely. But I'm happy to remove it it you eel strongly about it.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: Exposing the session is a language neutral requirement, in subsequent milestones, we'll add pause and cancel methods to the session.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: Will do in upcoming PR.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: But also the upload call is not like a request, it's more like a low level client that we are wrapping here. And upload call does not contain any request-specific information, etc. it's just the plumbing to make the calls. Note how the BeginUploadAsync or ResumeUploadAsync are the ones taking the request and settings, etc.

So, you create a session, but nothing happens until you call either begin or resume, at which time a session is bound to a specific upload.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: We will create other sessions with the same upload call, the upload call is just like the underlying client, there's no state-per-request in the upload call.

When a user calls BeginUploadAsync or ResumeUploadAsync, that's when the user passes the specific request, stream, etc. and the session is bounded to that from then on.

They are free to call BeginUploadAsync on two different sessions passing the same request and the same stream etc. and that's equivalent to sending the same request twice in any other method. If the API does not allow such a behavior, it will fail.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: All of this is for uploading really big files (you can also upload small ones, but hey). If your upload fails in the middle because you lost your connection, you can then create a session and  pass the upload URI (which you get when you called BegingUploadAsync) and the system will start uploading from the last data the server received. That's why it's called resumable upload, because you can resume it after it's failed for loss of connection. A session lives on the service for a handful of days, so even if you loose your connection for 24hrs or more, you can still pick up where you left off.

And also, later on we will add a Pause method (and a cancel one that you can't resume from).

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: Yes, chunk size should really be no larger than 2GiB, the stream size and the offset within that stream are the ones that may be bigger. I've updated the chunk size and granularity values to be `int`, all on the last commt.

Google.Api.Gax.Grpc/Rest/ResumableUploadSession.cs
amanda-tarafa: I'm on the fence about this one and will discuss internally when I'm back. I'm proposing, on the language neutral design, to keep the retry deadline for the transient errors quite short (although this is a hard to come-by-number), but then jump to recovery (or potentially we don't distinguish between transient and recoverable and always go to recovery). I'll leave it as is for now, but we'll revisit.

amanda-tarafa (commented):
Thanks, I've answered all your questions and fixed the main bit re: the chunk size being int. The rest are nits that I will address later.

PR #914, opened 2026-08-26
fix: Validate path parameters against path traversal dot segments

quartzmo:
Validates and encodes path parameters in gRPC-REST path transcoding.

* Rejects '.' and '..' path segments in both single-segment and multi-segment parameters to prevent path traversal (CWE-22).
* Percent-encodes each individual path segment to prevent parameter injection (CWE-88) while preserving '/' separators for multi-segment parameters.
* Unescapes input before segment evaluation to prevent percent-encoded delimiter bypasses.
* Added unit tests for path traversal validation and segment-level escaping.

b/529889305

### DIFF
--- a/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
@@ -95,7 +95,23 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
             {
                 request = new RuleTestRequest { X = xValue };
             }
-            Assert.Throws<ArgumentException>(() => rulePathPattern.TryFormat(request));
+            var exception = Assert.Throws<ArgumentException>(() => rulePathPattern.TryFormat(request));
+            string unescaped = Uri.UnescapeDataString(xValue);
+            bool hasDoubleDot = false;
+            bool hasSingleDot = false;
+            foreach (var segment in unescaped.Split('/'))
+            {
+                if (segment == "..") hasDoubleDot = true;
+                if (segment == ".") hasSingleDot = true;
+            }
+            if (hasDoubleDot)
+            {
+                Assert.Contains("contains invalid segment '..'", exception.Message);
+            }
+            else if (hasSingleDot)
+            {
+                Assert.Contains("contains invalid segment '.'", exception.Message);
+            }
         }
 
         [Theory]
--- a/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
@@ -70,8 +70,6 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
         // Dialogflow session (standard single-wildcard segment)
         [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/..")]
         [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/.")]
-        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1?key=val")]
-        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1#frag")]
         // Firestore documents (reserved double-wildcard path)
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/../../default")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/../../../../../../../escape-db")]
@@ -98,12 +96,6 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
             var exception = Assert.Throws<ArgumentException>(() => rulePathPattern.TryFormat(request));
             string unescaped = Uri.UnescapeDataString(xValue);
 
-            if (unescaped.Contains("?") || unescaped.Contains("#"))
-            {
-                Assert.StartsWith($"Path parameter 'x' contains invalid character", exception.Message);
-                return;
-            }
-
             bool isReserved = pattern.Contains("**");
             bool hasDoubleDot = false;
             bool hasSingleDot = false;
@@ -127,9 +119,13 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
 
         [Theory]
         [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1", "v3/projects/p/locations/l/agents/a/sessions/s1:detectIntent")]
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1?key=val", "v3/projects/p/locations/l/agents/a/sessions/s1%3Fkey%3Dval:detectIntent")]
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1#frag", "v3/projects/p/locations/l/agents/a/sessions/s1%23frag:detectIntent")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1", "v1/projects/sys-prod-123/databases/default/documents/doc-1/indexes")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/my-file.txt", "v1/projects/sys-prod-123/databases/default/documents/my-file.txt/indexes")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/my-file..txt", "v1/projects/sys-prod-123/databases/default/documents/my-file..txt/indexes")]
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc?key=val", "v1/projects/sys-prod-123/databases/default/documents/doc%3Fkey%3Dval/indexes")]
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc#frag", "v1/projects/sys-prod-123/databases/default/documents/doc%23frag/indexes")]
         public void ValidRealisticPatterns_Succeed(string pattern, string xValue, string expectedFormatResult)
         {
             var rulePathPattern = ParsePattern(pattern);
--- a/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
@@ -76,6 +76,8 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/../../default")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/../../../../../../../escape-db")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/%2e%2e/escape-db")]
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/..%2f..%2fescape-db")]
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/%2e%2e%2f%2e%2e%2fescape-db")]
         [InlineData("v1/{x=**}/indexes", "../escape-db")]
         [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/./child")]
         // Webhooks (multiple standard wildcards)
--- a/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
+++ b/Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
@@ -66,6 +66,49 @@ namespace Google.Api.Gax.Grpc.Rest.Tests
             Assert.Throws<ArgumentException>(() => HttpRulePathPattern.Parse(pattern, RuleTestRequest.Descriptor));
         }        
 
+        [Theory]
+        // Dialogflow session (standard single-wildcard segment)
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/..")]
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/.")]
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1?key=val")]
+        [InlineData("v3/{x=projects/*/locations/*/agents/*/sessions/*}:detectIntent", "projects/p/locations/l/agents/a/sessions/s1#frag")]
+        // Firestore documents (reserved double-wildcard path)
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/../../default")]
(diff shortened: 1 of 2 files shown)


Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,6 +256,19 @@ public string TryFormat(IMessage request)
...
+                throw new ArgumentException($"Path parameter '{JsonFieldPath}' contains invalid characters '?' or '#': '{result}'");
+            }
+            // Split path parameter value by '/' and reject any standalone dot (.) or double-dot (..) segments
+            // (either literal or URL-encoded) to prevent path traversal exploits.
+            if (result.Split('/').Select(segment => Uri.UnescapeDataString(segment)).Any(s => s == "." || s == ".."))
+            {
+                throw new ArgumentException($"Path parameter '{JsonFieldPath}' contains invalid segment '.' or '..': '{result}'");
+            }
             // Escape everything except slashes
             return string.Join("/", result.Split('/').Select(segment => Uri.EscapeDataString(segment)));
```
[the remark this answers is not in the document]
quartzmo: Fixed, PTAL.

efevans:
can you amend the PR with a `b/#` styled link to the issue

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,6 +256,23 @@ public string TryFormat(IMessage request)
...
+            // Reject query (?) or fragment (#) injections as defense-in-depth.
+            // Although segment-level Uri.EscapeDataString escapes these characters, this check
+            // guarantees explicit validation if the escaping behavior is modified in the future.
+            if (result.Contains("?") || result.Contains("#"))
+            {
+                throw new ArgumentException($"Path parameter '{JsonFieldPath}' contains invalid characters '?' or '#': '{result}'");
+            }
+            // Unescape the entire value first and split by '/' to prevent bypasses using URL-encoded slashes (e.g. %2f).
+            string unescapedVal = Uri.UnescapeDataString(result);
+            string[] segments = unescapedVal.Split('/');
```
efevans: For a hot path like this, I'd prefer something more performant than calling `Split` here which allocates each new string object on the heap (looks like this will be called once per field segment).

Better performance in .NET could be either iterating through the string by each char, or converting result of `UnescapeDataString` to a `Span` which can allow some string manipulation while keeping everything on the stack.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,6 +256,23 @@ public string TryFormat(IMessage request)
             {
                 return null;
             }
+            // Reject query (?) or fragment (#) injections as defense-in-depth.
+            // Although segment-level Uri.EscapeDataString escapes these characters, this check
+            // guarantees explicit validation if the escaping behavior is modified in the future.
+            if (result.Contains("?") || result.Contains("#"))
+            {
+                throw new ArgumentException($"Path parameter '{JsonFieldPath}' contains invalid characters '?' or '#': '{result}'");
```
efevans: In the design doc I see mention of  `*` and `**` characters in the value being disallowed but not these two characters. Are these checks something run by Amanda?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,6 +256,23 @@ public string TryFormat(IMessage request)
...
+            }
+            // Unescape the entire value first and split by '/' to prevent bypasses using URL-encoded slashes (e.g. %2f).
+            string unescapedVal = Uri.UnescapeDataString(result);
+            string[] segments = unescapedVal.Split('/');
+            foreach (var segment in segments)
+            {
+                if (segment == "." || segment == "..")
+                {
+                    throw new ArgumentException($"Path parameter '{JsonFieldPath}' contains invalid segment '.' or '..': '{result}'");
```
efevans: The doc includes separate error messages for `*` and `**`:

For `*`:
If a value matching * is exactly dot or two-dots then fail with an error message of `Invalid value {. or ..} for {resource ID field name}`.

For `**`:
If a value matching ** contains one or more segments that are exactly dot and/or two-dots then fail with an error message of `Value for { property name } must not contain segments that are exactly . or ..` .

efevans (changes requested):
Some questions about the requirements and performance improvements

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Yes, the ? and # checks are from the spec at [go/client-libraries:rest-special-uri-chars](https://goto.google.com/client-libraries:rest-special-uri-chars).

While standard path escaping percent-encodes ? and # to neutralize query injection, some templates use reserved expansion (which bypasses escaping to preserve slashes). Explicitly rejecting them here acts as a defense-in-depth check to prevent method override attacks or injection if the escaping behavior ever changes.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Fixed, thanks!

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Great catch, fixed!

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
efevans: This is my bad, and I didn't notice that the original implementation was already using `Split`. Could you revert back to a `Split` implementation, and then we'll check from there if there's any low hanging fruit?

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
efevans: I'm not seeing it, just the notes about disallowing `*` and `**`, and about encoding all special characters outside of `[-_.~/0-9a-zA-Z]`. Can you point me to the section? You can ping me over chat to avoid a direct link to the doc here.

efevans (changes requested):

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,75 @@ public string TryFormat(IMessage request)
             {
                 return null;
             }
-            // Escape everything except slashes
-            return string.Join("/", result.Split('/').Select(segment => Uri.EscapeDataString(segment)));
+            // Unescape the entire value first to prevent bypasses using URL-encoded slashes (e.g. %2f).
```
amanda-tarafa, line 261: We can't do this. URL-encoded slashes are how slashes are included as part of the value of a path segment. There's nothing wrong about it, and no URL parsers that follows spec will decode these values and interpret them as path separators.

(And if we needed to do this, all this code is way too big for that).

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,75 @@ public string TryFormat(IMessage request)
             {
                 return null;
             }
-            // Escape everything except slashes
-            return string.Join("/", result.Split('/').Select(segment => Uri.EscapeDataString(segment)));
```
amanda-tarafa, line 260: Simply check `segment` for equality with `.` and `..`. There's nothing more we need to do here.

amanda-tarafa (changes requested):

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
amanda-tarafa: > While standard path escaping percent-encodes ? and # to neutralize query injection, some templates use reserved expansion (which bypasses escaping to preserve slashes). Explicitly rejecting them here acts as a defense-in-depth check to prevent method override attacks or injection if the escaping behavior ever changes.

I'm not sure I understand this paragraph, but on this implementation there's no bypassing escaping to preserve slashes. Slashes are preserved because only the path segments are enconded. The way to protect against possible behavior changes in how we encode is by adding the right tests.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Updated, PTAL!

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Updated, PTAL!

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
quartzmo: Updated, PTAL!

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,28 @@ public string TryFormat(IMessage request)
...
+
+            // Escape each path segment individually to preserve slashes, while rejecting path traversal segments ('.' or '..').
+            // We unescape first to prevent path traversal bypasses via URL-encoded slashes (e.g. %2f).
```
amanda-tarafa: I had left this comment earlier:

"We can't do this. URL-encoded slashes are how slashes are included as part of the value of a path segment. There's nothing wrong about it, and no URL parsers that follows spec will decode these values and interpret them as path separators."

amanda-tarafa (changes requested):

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,27 @@ public string TryFormat(IMessage request)
             {
                 return null;
             }
-            // Escape everything except slashes
-            return string.Join("/", result.Split('/').Select(segment => Uri.EscapeDataString(segment)));
+
+            // Escape each path segment individually to preserve slashes, while rejecting path traversal segments ('.' or '..').
```
efevans, line 262: Update here to reflect changes in what the code is doing

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,27 @@ public string TryFormat(IMessage request)
...
+                    if (!_isReserved)
+                    {
+                        throw new ArgumentException($"Invalid value '{segment}' for {JsonFieldPath}");
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"Value for {JsonFieldPath} must not contain segments that are exactly . or ..");
+                    }
+                }
+                segments[i] = Uri.EscapeDataString(segment);
```
efevans, line 278: Correct me if I'm misunderstanding, but we need to remove this since we aren't unescaping anymore and we dont want to double escape

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,27 @@ public string TryFormat(IMessage request)
             {
                 return null;
             }
-            // Escape everything except slashes
-            return string.Join("/", result.Split('/').Select(segment => Uri.EscapeDataString(segment)));
+
+            // Escape each path segment individually to preserve slashes, while rejecting path traversal segments ('.' or '..').
+            string[] segments = result.Split('/');
+            for (int i = 0; i < segments.Length; i++)
```
efevans, line 264: See the other comments for more details, but we shouldn't need to mutate the string anymore so change this to a `foreach`

```suggestion
            foreach (var segment in segments)
```

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -256,8 +258,27 @@ public string TryFormat(IMessage request)
...
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"Value for {JsonFieldPath} must not contain segments that are exactly . or ..");
+                    }
+                }
+                segments[i] = Uri.EscapeDataString(segment);
+            }
+
+            return string.Join("/", segments);
```
efevans, line 281: Again this is predicated on my understanding on not needing to escape the strings anymore, but we should just be able to return `result` here.

efevans (changes requested):
Few comments, cut the review short and let me know if I'm wrong on not needing to escape the segments anymore.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
amanda-tarafa: Un-scaping was introduced by this PR. This is the original scaping that we were doing to begin with.

efevans: Got it, I see it in the older version. Resolving my comments

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
efevans: My misunderstanding here, looks good.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
efevans: My misunderstanding here, looks good.

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
efevans: My misunderstanding here, looks good.

efevans (approved):
LGTM

efevans:
Note to squash before merge

quartzmo:
@efevans What is the condition to remove the `do not merge` label?

efevans:
> @efevans What is the condition to remove the `do not merge` label?

@quartzmo For the .NET repos we usually merge with one commit per piece of work to keep the `main` branch clean. I just get in the habit of putting the label on my own PRs as a reminder and put it here since its your first PR for the .NET side

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -127,6 +127,7 @@ private sealed class FieldSegment : IPathSegment
 
         private readonly Regex _validationRegex;
         private readonly Func<IMessage, string> _propertyAccessor;
+        private readonly bool _isReserved;
```
amanda-tarafa: I think we can simplify this code a lot if instead of storing a bool we store the error message template.
```suggestion
        private readonly string _dotsErrorTemplate;
```

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
```diff
@@ -144,6 +145,7 @@ internal FieldSegment(string fieldText, MessageDescriptor descriptor)
             string fieldPath = bits[0];
             string pattern = bits.Length == 2 ? bits[1] : "*";
             _validationRegex = ConvertPatternForValidation(pattern);
+            _isReserved = pattern.Contains("**");
```
amanda-tarafa: Better to use EndsWith which is more efficient. Plus the change I'm suggesting regarding the message template.
```suggestion
            _isReserved = pattern.EndsWith("**");
```

Google.Api.Gax.Grpc/Rest/HttpRulePathPattern.cs
amanda-tarafa: And then here we can go back to using linq:
```suggestion
            return string.Join("/", result.Split('/').Select(segment =>
            {
                if (segment == "." || segment == "..")
                {
                    throw new ArgumentException(string.Format(_dotsErrorTemplate, JsonFieldPath, segment));
                }
                return Uri.EscapeDataString(segment);
            }));
```

amanda-tarafa (changes requested):
In general looks good, but making the code more idiomatic.

Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
```diff
@@ -66,6 +66,72 @@ public void InvalidPattern(string pattern)
...
+        [InlineData("v1/{x=**}/indexes", "../escape-db")]
+        [InlineData("v1/{x=projects/*/databases/*/documents/**}/indexes", "projects/sys-prod-123/databases/default/documents/doc-1/./child")]
+        // Webhooks (multiple standard wildcards)
+        [InlineData("v3/projects/{x}/webhooks/{nested.a}", "..")]
+        [InlineData("v3/projects/{x}/webhooks/{nested.a}", ".")]
+        public void PathTraversalAndInjection_ThrowsArgumentException(string pattern, string xValue)
+        {
+            var rulePathPattern = ParsePattern(pattern);
+            RuleTestRequest request;
+            if (pattern.Contains("nested.a"))
```
amanda-tarafa: Tests shouldn't be conditional. Follow the pattern that's earlier on this file, deifining TheoryData with the RuleTestRequest and the use the ConvertTheoryData method (or a similar one if that one is not good enough because of the amount of parameters).

Google.Api.Gax.Grpc.Tests/Rest/HttpRulePathPatternTest.cs
```diff
@@ -66,6 +66,72 @@ public void InvalidPattern(string pattern)
...
+
+            string paramName = pattern.Contains("nested.a") ? "nested.a" : "x";
+            if (!isReserved)
+            {
+                string matchedDot = hasDoubleDot ? ".." : (hasSingleDot ? "." : "");
+                Assert.StartsWith($"Invalid value '{matchedDot}' for {paramName}", exception.Message);
+            }
+            else
+            {
+                Assert.StartsWith($"Value for {paramName} must not contain segments that are exactly . or ..", exception.Message);
```
amanda-tarafa: The expected error message should be part of the input received by the test as InlineData or TheoryData.

amanda-tarafa (changes requested):

